PR #1862, opened 2024-12-18
fix: any_to_mp3

shaohuzhang1:
fix: any_to_mp3

### DIFF
--- a/apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
+++ b/apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
@@ -37,7 +37,7 @@ class BaseSpeechToTextNode(ISpeechToTextNode):
                 temp_mp3_path = temp_amr_file.name
             any_to_mp3(temp_file_path, temp_mp3_path)
             try:
-                return split_and_transcribe(temp_file_path, model)
+                return split_and_transcribe(temp_mp3_path, model)
             finally:
                 os.remove(temp_file_path)
                 os.remove(temp_mp3_path)


apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
```diff
@@ -37,7 +37,7 @@ def process_audio_item(audio_item, model):
                 temp_mp3_path = temp_amr_file.name
             any_to_mp3(temp_file_path, temp_mp3_path)
             try:
-                return split_and_transcribe(temp_file_path, model)
+                return split_and_transcribe(temp_mp3_path, model)
             finally:
                 os.remove(temp_file_path)
                 os.remove(temp_mp3_path)
```
shaohuzhang1, line 43: There are no apparent irregularities in the provided code snippet regarding logic or syntax. However, you might want to consider optimizing this function slightly by:

1. **Reduce Redundancy**: The `temp_file_path` is defined twice in the try block before it's used. It can be removed from one definition.

2. **Avoid Using `os.remove` Inside a Try Block**: Removing files when an exception occurs makes it more difficult to handle errors correctly later if needed. Moving the removal of these files to a `finally` block ensures they are always cleaned up.

Here is the revised version of the function:

```python
def process_audio_item(audio_item, model):
    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
        temp_name = temp_file.name

    any_to_mp3(audio_item.path, temp_name)

    try:
        return split_and_transcribe(temp_name, model)
    finally:
        os.remove(temp_name)
```

This change simplifies the code and improves its robustness by ensuring that temporary resources are properly disposed of after use regardless of whether an error occurred during processing.

PR #1870, opened 2024-12-18
fix: Low version browsers do not have the at function

shaohuzhang1:
fix: Low version browsers do not have the at function

### DIFF
--- a/ui/src/api/type/application.ts
+++ b/ui/src/api/type/application.ts
@@ -138,7 +138,7 @@ export class ChatRecordManage {
     ) {
       return this.write_node_info
     }
-    const run_node = this.node_list.filter((item) => item.buffer.length > 0 || !item.is_end).at(0)
+    const run_node = this.node_list.filter((item) => item.buffer.length > 0 || !item.is_end)[0]
 
     if (run_node) {
       const index = this.node_list.indexOf(run_node)


ui/src/api/type/application.ts
```diff
@@ -138,7 +138,7 @@ export class ChatRecordManage {
     ) {
       return this.write_node_info
     }
-    const run_node = this.node_list.filter((item) => item.buffer.length > 0 || !item.is_end).at(0)
+    const run_node = this.node_list.filter((item) => item.buffer.length > 0 || !item.is_end)[0]
 
     if (run_node) {
       const index = this.node_list.indexOf(run_node)
```
shaohuzhang1, line 144: The provided code is mostly clean, but there's one line that can be optimized. Here’s the relevant part:

```javascript
const run_node = this.node_list.filter((item) => item.buffer.length > 0 || !item.is_end)[0];
```

This line uses `filter` with an array containing exactly one element, which can be written more directly using bracket notation:

```javascript
const run_node = this.node_list[0] || null;
```

However, since you're checking both conditions (`buffer.length > 0` and `!item.is_end`), you should maintain the original logic while simplifying it a bit by removing redundant steps:

```javascript
const node_length_greater_than_zero = !!this.node_list.filter(x => x.buffer.length > 0).length;
const node_isnt_finished = !!this.node_list.find(x => !x.is_end);

if (node_length_greater_than_zero && node_isnt_finished) {
    // Your logic here
}
```

But if you want to keep everything concise and efficient without unnecessary nesting, you could just use:

```javascript
const next_runnable = this.node_list.filter(item => item.buffer.length > 0 || !item.is_end)?.shift();
```

If none of these options meet your requirements, please let me know how I can assist further.

PR #1879, opened 2024-12-20
fix: fix the defect of interface parameter transmission failure

shaohuzhang1:
fix: fix the defect of interface parameter transmission failure  --bug=1050500 --user=王孝刚 【接口文档】接口文档里的获取知识库列表（不分页）接口，带参数查询无数据 https://www.tapd.cn/57709429/s/1634918

### DIFF
--- a/apps/dataset/views/dataset.py
+++ b/apps/dataset/views/dataset.py
@@ -103,7 +103,8 @@ class Dataset(APIView):
                          tags=["知识库"])
     @has_permissions(PermissionConstants.DATASET_READ, compare=CompareConstants.AND)
     def get(self, request: Request):
-        d = DataSetSerializers.Query(data={**request.query_params, 'user_id': str(request.user.id)})
+        data = {key: str(value) for key, value in request.query_params.items()}
+        d = DataSetSerializers.Query(data={**data, 'user_id': str(request.user.id)})
         d.is_valid()
         return result.success(d.list())
 


apps/dataset/views/dataset.py
```diff
@@ -103,7 +103,8 @@ def get(self, request: Request, dataset_id: str):
                          tags=["知识库"])
     @has_permissions(PermissionConstants.DATASET_READ, compare=CompareConstants.AND)
     def get(self, request: Request):
-        d = DataSetSerializers.Query(data={**request.query_params, 'user_id': str(request.user.id)})
+        data = {key: str(value) for key, value in request.query_params.items()}
+        d = DataSetSerializers.Query(data={**data, 'user_id': str(request.user.id)})
         d.is_valid()
         return result.success(d.list())
```
shaohuzhang1, line 110: The provided code has several issues:

1. **Code Duplication**: The `get` method is defined twice with different function signatures.
2. **Error Handling**: There's no error handling implemented for exceptions that might occur during serialization or query processing.

### Suggestions

- Combine the two functions into one using overloading or default arguments to handle both use cases (without parameters and with specific parameters).
- Implement exception/error handling:
  ```python
  try:
      d = DataSetSerializers.Query(data=data)
      d.is_valid()
      return result.success(d.list())
  except Exception as e:
      # Handle errors appropriately, such as logging them or returning an error response
      raise Exception(f"Failed to retrieve dataset: {str(e)}")
  ```

- Simplify accessing values from `request.query_params`: You can convert all values directly to strings if there are no special requirements regarding type conversion.

Here's the revised version of the code incorporating these suggestions:

```python
@@ -103,7 +105,6 @@ def get(self, request: Request, dataset_id: str):
                          tags=["知识库"])
     @has_permissions(PermissionConstants.DATASET_READ, compare=CompareConstants.AND)
     def get(self, request: Request = None):
-        if request is not None:
-            data = {key: str(value) for key, value in request.query_params.items()}
+            data = {key: str(value) for key, value in request.query_params}
             d = DataSetSerializers.Query(data={**data, 'user_id': str(request.user.id)})
             d.is_valid()
         else:
             q_params = request.query_params.copy()  # Assuming without a specific parameter this is the same behavior
             q_params['user_id'] = str(request.user.id)
             d = DataSetSerializers.Query(data=q_params)
             d.is_valid()

         return result.success(d.list())

```

This should make the code more efficient and maintainable while addressing some common issues.

PR #1886, opened 2024-12-20
refactor: Workflow execution logic

shaohuzhang1:
refactor: Workflow execution logic

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -29,7 +29,7 @@ from function_lib.models.function import FunctionLib
 from setting.models import Model
 from setting.models_provider import get_model_credential
 
-executor = ThreadPoolExecutor(max_workers=50)
+executor = ThreadPoolExecutor(max_workers=200)
 
 
 class Edge:
@@ -271,7 +271,7 @@ class WorkflowManage:
         self.current_result = None
         self.answer = ""
         self.answer_list = ['']
-        self.status = 0
+        self.status = 200
         self.base_to_response = base_to_response
         self.chat_record = chat_record
         self.await_future_map = {}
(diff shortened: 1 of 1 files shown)


apps/application/flow/workflow_manage.py
```diff
@@ -409,18 +421,10 @@ def run_chain(self, current_node, node_result_future=None):
...
-            [r.get('future').result() for r in result_list]
-            if self.status == 0:
-                self.status = 200
+            return result
         except Exception as e:
             traceback.print_exc()
+        return []
 
     def hand_node_result(self, current_node, node_result_future):
         try:
```
shaohuzhang1, line 430: Some potential issues and optimizations in the provided code include:

1. **Large `max_workers` Setting**: The `ThreadPoolExecutor` has been set to utilize 200 worker threads, which might be too high depending on system resources and task load. Consider reducing this number to ensure efficient resource utilization.

2. **Lock Usage with Race Conditions**: Directly manipulating shared data like `current_node.status` with a lock without ensuring thread safety could lead to race conditions. It's better to use synchronization mechanisms that manage these operations correctly.

3. **Node Processing Logic Complexity**: The processing logic can become complex when handling multiple nodes concurrently and managing an asynchronous chain of operations. Simplifying this process can help reduce complexity and improve performance.

Here are some specific suggestions for improvement:

### Potential Improvements

#### Reduce Maximum Worker Threads
Consider dynamically adjusting the number of workers based on system available memory and CPU usage. You can implement monitoring tools like Prometheus or Grafana to gauge these factors before deciding on a suitable number. Here’s how you could adjust it programmatically:

```python
import psutil

def optimal_worker_count():
    total_memory = psutil.virtual_memory().total / (1024**3)  # Total RAM in GBs
    num_cores = psutil.cpu_count(logical=False)  # Physical cores only
    
    if total_memory < 8:
        return min(50, num_cores * 4)
    elif total_memory >= 8 and total_memory < 16:
        return min(100, num_cores * 8)
    else:
        return min(200, num_cores * 10)

executor = ThreadPoolExecutor(max_workers=optimal_worker_count())
```

#### Use Thread-Safe Data Structures
Ensure that shared data structures like dictionaries and lists are accessed safely using thread-safe collections such as `collections.ConcurrentDict`.

#### Simplify Node Handling
Rather than managing a dictionary of futures (`await_future_map`) explicitly, consider using higher-level constructs or libraries designed for distributed computing and parallel execution tasks more effectively.

#### Review Logging Levels
Reduce the verbosity of logs inside critical sections to avoid excessive noise during operation. Logs should be informative yet minimalistic.

### Example Adjustment:
Replace current locking implementation with locks that handle concurrency appropriately:

```python
from threading import Lock

class Edge(object):
    
    def __init__(self, flow: Flow, params, work_flow_post_handler: WorkFlowPostHandler,
                 base_to_response: BaseToResponse, chat_record: dict):
        
        self.params = params
        self.work_flow_post_handler = work_flow_post_handler
        ...
        self.lock = Lock()

    def run_chain_manage(self, current_node, node_result_future):

        ...

             for i in range(len(node_list)):
                 node_run_list.append({"node": node_list[i],
                                     "future": executor.submit(self.run_chain_manage, node_list[i], None)})
         
            ...

        self.async_wait()
        
    def async_wait(self):
        while any(r['future'].done() is False for r in self.result_set):
            time.sleep(0)  # Wait until one or more results arrive
        
        ...  # Process completed results...
        
```
This approach uses Python's built-in concurrency utilities to achieve similar goals efficiently. Always test after these modifications to confirm no new issues arise with your application.

PR #1887, opened 2024-12-23
fix: fix audio frame rate

shaohuzhang1:
fix: fix audio frame rate

### DIFF
--- a/apps/common/util/common.py
+++ b/apps/common/util/common.py
@@ -179,6 +179,7 @@ def any_to_mp3(any_path, mp3_path):
         sil_to_wav(any_path, any_path)
         any_path = mp3_path
     audio = AudioSegment.from_file(any_path)
+    audio = audio.set_frame_rate(16000)
     audio.export(mp3_path, format="mp3")
 
 


apps/common/util/common.py
```diff
@@ -179,6 +179,7 @@ def any_to_mp3(any_path, mp3_path):
         sil_to_wav(any_path, any_path)
         any_path = mp3_path
     audio = AudioSegment.from_file(any_path)
+    audio = audio.set_frame_rate(16000)
     audio.export(mp3_path, format="mp3")
```
shaohuzhang1, line 185: The provided function `any_to_mp3` appears to be designed to convert files from one audio format to MP3. However, there are a few areas where it can be improved:

1. **Error Handling**: The function currently does not handle cases where the file at `any_path` does not exist or is not valid.

2. **File Type Detection**: If you want this function to work with various audio formats (e.g., WAV, AAC), you should add logic to detect the input file type and attempt conversion based on that.

3. **Conversion Parameters**: Although no conversion parameters seem to be set here, ensuring that settings like bit rate and sample size are correctly applied could improve output quality and reduce file size if needed.

Here's an updated version of the function with some potential improvements:
```python
import soundfile as sf

def any_to_mp3(any_path, mp3_path, sample_rate=16000, bits_per_sample=16):
    try:
        # Load audio data
        audio_data, original_sample_rate = sf.read(any_path)

        # Convert to 16kHz sampling rate (default, but optional parameter)
        if original_sample_rate != sample_rate:
            audio_data = sf.resample(audio_data, original_sample_rate, sample_rate)

        # Export to MP3
        sf.write(mp3_path, audio_data.astype('int16'), samplerate=sample_rate, subtype='PCM_16')

    except Exception as e:
        print(f"An error occurred while converting {any_path} to MP3: {str(e)}")

# Example usage
any_to_mp3('input.wav', 'output.mp3')
```

### Key Improvements:
- **Using SoundFile (`sf`)**: This library simplifies audio reading, writing, and resampling.
- **Error Handling**: Added exception handling to catch potential issues during file operations.
- **Optional Conversion Settings**: Allows setting the desired sample rate and bits per sample for the output MP3 file.

PR #1888, opened 2024-12-23
feat: Node execution conditions

shaohuzhang1:
feat: Node execution conditions

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -676,9 +676,17 @@ class WorkflowManage:
                             self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
         else:
             for edge in self.flow.edges:
-                if edge.sourceNodeId == current_node.id and self.dependent_node_been_executed(edge.targetNodeId):
-                    node_list.append(
-                        self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
+                if edge.sourceNodeId == current_node.id:
+                    next_node = [node for node in self.flow.nodes if node.id == edge.targetNodeId]
+                    if len(next_node) == 0:
+                        continue
+                    if next_node[0].properties.get('condition', "AND") == 'AND':
+                        if self.dependent_node_been_executed(edge.targetNodeId):
+                            node_list.append(
+                                self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
+                    else:
+                        node_list.append(
+                            self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
         return node_list
 
     def get_reference_field(self, node_id: str, fields: List[str]):
--- a/ui/src/workflow/common/NodeContainer.vue
+++ b/ui/src/workflow/common/NodeContainer.vue
@@ -44,8 +44,8 @@
               placement="bottom-start"
             >
               <el-button text>
-                <img src="@/assets/icon_or.svg" alt="" v-if="condition==='OR'">
-                <img src="@/assets/icon_and.svg" alt="" v-if="condition==='AND'">
+                <img src="@/assets/icon_or.svg" alt="" v-if="condition === 'OR'" />
+                <img src="@/assets/icon_and.svg" alt="" v-if="condition === 'AND'" />
               </el-button>
               <template #dropdown>
                 <div style="width: 280px" class="p-12-16">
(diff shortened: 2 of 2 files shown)


ui/src/workflow/common/NodeContainer.vue
```diff
@@ -152,8 +152,19 @@ const height = ref<{
...
+    if (props.nodeModel.properties.condition) {
+      return props.nodeModel.properties.condition
+    }
+    set(props.nodeModel.properties, 'condition', 'AND')
+    return true
+  }
+})
 const showNode = computed({
   set: (v) => {
     set(props.nodeModel.properties, 'showNode', v)
```
shaohuzhang1, line 170: The code you provided has some improvements and optimizations that can be made:

### Improvements:
1. **Computed Property Optimization**: The use of `computed` properties for `condition` and `showNode` is a good practice. This allows Vue to efficiently update these properties when data changes, but it's important that they reflect the state correctly without re-setting values unnecessarily.

2. **String Comparison with Single Equal Sign**: Replace single equals signs (`=`) with triple equal signs (`===`) for strict equality comparisons in methods like `set`. Using three equal signs ensures no type coercion occurs, which makes the comparison clearer and less prone to error.

3. **Code Formatting Consistency**: Ensure consistent spacing around operators, braces, and commas to improve readability. This improves maintainability.

4. **Image Source Path Correction**: Ensure the image paths are correct and accessible relative to the directory where this script runs. If images are located outside the assets folder, adjust their paths accordingly.

### Optimize Conditional Rendering:
Instead of using conditional rendering inside the template directly (using `v-if`), consider restructuring the logic so that conditionally rendered elements remain part of the DOM and only visible based on their conditions. This can sometimes be more efficient than creating or destroying them dynamically during component updates.

Here's an updated version of your code incorporating these suggestions):

```vue
<template>
  <!-- Rest of the template remains unchanged -->

<script lang="ts">
import { defineComponent } from 'vue'
// Import other necessary components here

  setup(props) {
    const set = (...args: any[]) => console.log(args) // Placeholder function
    
    const height = ref<{ top?: number; bottom?: number }>({
      // initialization
    })
    
    const showAnchor = ref(false)
    const anchorData = ref({})
    
    const condition = computed(() => {
      return props.nodeModel.properties?.condition || 'AND'
    })
    
    const showNode = computed(() => {
      return props.nodeModel.properties.showNode ?? true
    })
    
    // Other methods and computed properties...
  }
})
</script>

<!-- Reset CSS and styling -->
<style scoped src="./YourStylesheet.css"></style>
```

This revised approach maintains simplicity while improving robustness for future modifications.

apps/application/flow/workflow_manage.py
```diff
@@ -676,9 +676,17 @@ def get_next_node_list(self, current_node, current_node_result):
...
+                    if next_node[0].properties.get('condition', "AND") == 'AND':
+                        if self.dependent_node_been_executed(edge.targetNodeId):
+                            node_list.append(
+                                self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
+                    else:
+                        node_list.append(
+                            self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
         return node_list
 
     def get_reference_field(self, node_id: str, fields: List[str]):
```
shaohuzhang1, line 692: This code snippet appears to implement a method that determines which nodes should be processed next based on certain conditions. Here are some potential problems and optimizations:

**Potential Issues:**

1. **List Comprehension in Loop:** 
   - The line `next_node = [node for node in self.flow.nodes if node.id == edge.targetNodeId]` iterates over all nodes once per iteration of the main loop. This is unnecessary when you only need one target node from each connection.

2. **Multiple Returns:**
   - If multiple nodes have the same `targetNodeId`, this code will append them all, but it won't stop after the first match. Consider adding a flag to break out of the function once a match is found to optimize performance for cases where duplicates exist.

3. **Condition Handling:**
   - The code checks whether the condition property exists before accessing it (`if next_node[0].properties.get('condition', "AND")`). However, if `condition` doesn't exist, it will use `"AND"` as default, assuming every condition has been met. Make sure this default behavior aligns with your needs.

4. **Code Reusability:**
   - The function `get_next_node_list` could benefit from reusing logic in helper functions like checking dependencies and getting node classes by ID.

**Suggestions:**

1. **Simplify Condition Check:**
   ```python
   if edge.sourceNodeId == current_node.id:
       # Determine if condition matches OR/AND rule
       if self._should_process_and_rule(node_id, edge.targetNodeId):
           node_list.append(self.get_node_cls_by_id(edge.targetNodeId, self.get_up_node_id_list(edge.targetNodeId)))
   ```

2. **Use Helper Function for Node Retrieval:**
   Create a helper function `_get_target_nodes` that fetches and returns target nodes directly without iterating multiple times per edge.

3. **Add Debugging Information:**
   Insert print statements or logging to trace execution flow and identify bottlenecks more easily.

Overall, improving efficiency and ensuring robust conditional handling would make the code cleaner and potentially faster under certain scenarios.

PR #1889, opened 2024-12-23
fix: fix bailian stt

shaohuzhang1:
fix: fix bailian stt

### DIFF
--- a/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/model/stt.py
+++ b/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/model/stt.py
@@ -4,6 +4,7 @@ from typing import Dict
 
 import dashscope
 from dashscope.audio.asr import (Recognition)
+from pydub import AudioSegment
 
 from setting.models_provider.base_model_provider import MaxKBBaseModel
 from setting.models_provider.impl.base_stt import BaseSpeechToText
@@ -40,7 +41,6 @@ class AliyunBaiLianSpeechToText(MaxKBBaseModel, BaseSpeechToText):
         dashscope.api_key = self.api_key
         recognition = Recognition(model=self.model,
                                   format='mp3',
-                                  sample_rate=16000,
                                   callback=None)
         with tempfile.NamedTemporaryFile(delete=False) as temp_file:
             # 将上传的文件保存到临时文件中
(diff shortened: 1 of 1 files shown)


apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/model/stt.py
```diff
@@ -49,6 +49,13 @@ def speech_to_text(self, audio_file):
...
+            audio = AudioSegment.from_file(temp_file_path)
+            if audio.channels != 1:
+                audio = audio.set_channels(1)
+            audio = audio.set_frame_rate(16000)
+
+            # 将转换后的音频文件保存到临时文件中
+            audio.export(temp_file_path, format='mp3')
             # 识别临时文件
             result = recognition.call(temp_file_path)
             text = ''
```
shaohuzhang1, line 61: There are a couple of issues and suggestions to address:

1. **Sample Rate Settwing**: The `sample_rate` is being set explicitly within the function, but it's already defined when creating the `Recognition` object. You can either remove this line or ensure that it matches the default value used by DashScope.

2. **Channel Conversion**: Setting channels and frame rate programmatically can sometimes lead to unexpected behavior. It's better to use predefined methods provided by PyDub or DashScope itself where applicable, especially if you need consistent handling across platforms.

3. **Callback Handling**: Since the callback parameter is not utilized in your code snippet, consider whether this could be removed for simplicity.

Here's an updated version with these adjustments:

```python
import dashscope
from dashscope.audio.asr import (Recognition)

from setting.models_provider.impl.base_stt import BaseSpeechToText

class YourSTTClass(BaseSpeechToText):
    def __init__(self, api_key, model):
        super().__init__()
        dashboard_scope.api_key = self.api_key
        self.model = model

    @staticmethod
    async def convert_audio_format(file_path: str) -> None:
        """Converts the audio file format to MP3."""
        try:
            audio_segment = AudioSegment.from_file(file_path)
            # Ensure single channel and 16kHz sampling rate
            audio_segment = audio_segment.set_channels(1).set_frame_rate(16000)
            audio_segment.export(file_path, newformat="mp3")
        except Exception as e:
            print(f"Error converting audio: {e}")

    async def recognize(self, audio_data: bytes | str) -> list[dict]:
        """Recognizes speech from binary data."""
        recognition = Recognition(
            model=self.model,
            format='mp3'
            
        )
    
        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
            temp_file.write(audio_data)
            temp_file.close()
            temp_file_path = temp_file.name
    
        await self.convert_audio_format(temp_file_path)
    
        # Call the API for recognition
        result = await recognition.call_async(temp_file_path)
        return [line.data['result'] for line in result]

```

### Key Changes:
- Removed the redundant `sample_rate` assignment.
- Added a method `convert_audio_format` using PyDub for cleaning up audio samples.
- Used asynchronous calls where appropriate, such as `recognition.call_async()` which is more suitable for real-time applications.

These changes should improve robustness and consistency in audio processing.

PR #1898, opened 2024-12-24
fix: fix the defect of empty line to speech error

shaohuzhang1:
fix: fix the defect of empty line to speech error  --bug=1050760 --user=王孝刚 【应用】文本转语音使用豆包模型时，用户问题文本中间有空行，会报错 https://www.tapd.cn/57709429/s/1636230

### DIFF
--- a/apps/common/util/common.py
+++ b/apps/common/util/common.py
@@ -211,3 +211,7 @@ def split_and_transcribe(file_path, model, max_segment_length_ms=59000, audio_fo
         if isinstance(text, str):
             full_text.append(text)
     return ' '.join(full_text)
+
+
+def _remove_empty_lines(text):
+    return '\n'.join(line for line in text.split('\n') if line.strip())
--- a/ui/src/api/application.ts
+++ b/ui/src/api/application.ts
@@ -429,9 +429,10 @@ const getPlatformConfig: (application_id: string, type: string) => Promise<Resul
 const updatePlatformConfig: (
   application_id: string,
   type: string,
-  data: any
-) => Promise<Result<any>> = (application_id, type, data) => {
-  return post(`/platform/${application_id}/${type}`, data)
+  data: any,
+  loading?: Ref<boolean>
+) => Promise<Result<any>> = (application_id, type, data, loading) => {
+  return post(`/platform/${application_id}/${type}`, data, undefined, loading)
 }
 /**
  * 更新平台状态
--- a/ui/src/views/application/component/AccessSettingDrawer.vue
+++ b/ui/src/views/application/component/AccessSettingDrawer.vue
@@ -222,10 +222,13 @@ const submit = async () => {
   formRef.value?.validate(async (valid) => {
     if (valid) {
       try {
-        await applicationApi.updatePlatformConfig(id, configType.value, form[configType.value])
-        MsgSuccess('配置保存成功')
-        closeDrawer()
-        emit('refresh')
+        applicationApi
+          .updatePlatformConfig(id, configType.value, form[configType.value], loading)
+          .then(() => {
+            MsgSuccess('配置保存成功')
+            closeDrawer()
+            emit('refresh')
+          })
       } catch {
         MsgError('保存失败，请检查输入或稍后再试')
       }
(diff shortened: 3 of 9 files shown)


ui/src/views/application/component/AccessSettingDrawer.vue
```diff
@@ -222,10 +222,13 @@ const submit = async () => {
...
+        applicationApi
+          .updatePlatformConfig(id, configType.value, form[configType.value], loading)
+          .then(() => {
+            MsgSuccess('配置保存成功')
+            closeDrawer()
+            emit('refresh')
+          })
       } catch {
         MsgError('保存失败，请检查输入或稍后再试')
       }
```
shaohuzhang1, line 234: The provided code looks generally correct. However, there is an improvement opportunity to manage the loading state more explicitly:

1. Instead of using global `loading` variable, pass it as a parameter to the API call. This makes testing easier and keeps components decoupled from shared state.

```javascript
applicationApi.updatePlatformConfig(
  id,
  configType.value,
  form[configType.value],
  loading // Passing the loading flag
).then(() => {
  MsgSuccess('配置保存成功')
  closeDrawer()
  emit('refresh')
}).catch((error) => {
  msgError(`保存失败，错误信息: ${error.message}`); // Handling errors separately
});
```

This change enhances reusability and maintainability by isolating the loading state within each component.

ui/src/api/application.ts
```diff
@@ -429,9 +429,10 @@ const getPlatformConfig: (application_id: string, type: string) => Promise<Resul
...
-  data: any
-) => Promise<Result<any>> = (application_id, type, data) => {
-  return post(`/platform/${application_id}/${type}`, data)
+  data: any,
+  loading?: Ref<boolean>
+) => Promise<Result<any>> = (application_id, type, data, loading) => {
+  return post(`/platform/${application_id}/${type}`, data, undefined, loading)
 }
 /**
  * 更新平台状态
```
shaohuzhang1, line 438: The provided function `updatePlatformConfig` has a couple of issues:

1. **Missing Parameters**:
   - The signature for `updatePlatformConfig` does not accept the `data` parameter when it is followed by an optional `loading` parameter.

2. **Incorrect Argument Passing**:
   - In the call to the `post` method within the implementation, there is no explicit value passed for the fourth argument (`headers`). It should be either undefined or explicitly set if needed.

Here's how you can fix these issues:

```typescript
const updatePlatformConfig: (
  application_id: string,
  type: string,
  data: any,
  loading?: Ref<boolean>
) => Promise<Result<any>> = (application_id, type, data, loading?) => {
  // Ensure that headers remain consistent with the post request
  return post(`/platform/${application_id}/${type}`, data, undefined, loading);
}
```

By adding the missing parameters and ensuring that all necessary arguments are correctly passed, the code will now work as expected without errors. This modification maintains clarity and correctness in the function definition and usage.

apps/common/util/common.py
```diff
@@ -211,3 +211,7 @@ def split_and_transcribe(file_path, model, max_segment_length_ms=59000, audio_fo
         if isinstance(text, str):
             full_text.append(text)
     return ' '.join(full_text)
+
+
+def _remove_empty_lines(text):
+    return '\n'.join(line for line in text.split('\n') if line.strip())
```
shaohuzhang1, line 217: There are two small issues and one optimization suggestion in the provided code:

### Issues:
1. **Missing `return` Statement**: The function `_remove_empty_lines` is missing a return statement at the end. This should be added to properly return the cleaned string.
   
2. **Potential Memory Leak**: The function currently uses list comprehensions to create new lists (`full_text`) and `' '.join(...)`. While this approach avoids modifying strings directly, it might not lead to efficient memory usage with very large datasets because each operation creates new objects.

### Optimization Suggestion:
Instead of using list comprehensions for operations like filtering or joining strings, consider using generators when possible. Generators can reduce memory consumption since they yield results one at a time rather than creating intermediate lists.
```python
    # Use generator expression to filter out empty lines
    non_empty_lines = (line for line in text.split('\n') if line.strip())
    return ''.join(non_empty_lines)
```

These changes will help ensure that the function behaves correctly and efficiently.

PR #1900, opened 2024-12-24
fix: Workflow import did not insert AccessToken information

shaohuzhang1:
fix: Workflow import did not insert AccessToken information

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -697,6 +697,9 @@ class ApplicationSerializer(serializers.Serializer):
             application_model = self.to_application(application, user_id)
             function_lib_model_list = [self.to_function_lib(f, user_id) for f in function_lib_list]
             application_model.save()
+            # 插入认证信息
+            ApplicationAccessToken(application_id=application_model.id,
+                                   access_token=hashlib.md5(str(uuid.uuid1()).encode()).hexdigest()[8:24]).save()
             QuerySet(FunctionLib).bulk_create(function_lib_model_list) if len(function_lib_model_list) > 0 else None
             return True
 


apps/application/serializers/application_serializers.py
```diff
@@ -697,6 +697,9 @@ def import_(self, with_valid=True):
             application_model = self.to_application(application, user_id)
             function_lib_model_list = [self.to_function_lib(f, user_id) for f in function_lib_list]
             application_model.save()
+            # 插入认证信息
+            ApplicationAccessToken(application_id=application_model.id,
+                                   access_token=hashlib.md5(str(uuid.uuid1()).encode()).hexdigest()[8:24]).save()
             QuerySet(FunctionLib).bulk_create(function_lib_model_list) if len(function_lib_model_list) > 0 else None
             return True
```
shaohuzhang1, line 705: The provided code snippet has a few issues:

### Issues Identified:
1. **Import Statement Missing**: The `import` statement for `uuid` is missing at the top of the file, which could lead to errors at runtime.
2. **Variable Naming Convention**: Variable names should be descriptive but concise (e.g., use camelCase instead of snake_case).
3. **String Formatting**: The string formatting used inside the MD5 function may not behave as expected due to the incorrect usage of brackets.

### Suggestions for Optimization and Fixes:
1. **Add Import Statement**:
   ```python
   import uuid
   ```

2. **Refactor Function Name and Parameter Names**:
   ```python
   def importApplication(self, with_valid=True):
       application_model = self.to_application(application, user_id)
       function_lib_models = [self.to_function_lib(f, user_id) for f in function_lib_list]
       
       # Save application model first
       application_model.save()
      
       # Insert authentication information
       ApplicationAccessToken.objects.create(
           application_id=application_model.id,
           access_token=hashlib.md5(str(uuid.uuid4()).encode()).hexdigest()[:24]  # Use uuid.uuid4 directly without hashing manually
       )
       
       # Bulk create function library models
       QuerySet(FunctionLib).bulk_create(function_lib_models) if len(function_lib_models) > 0 else None
       
       return True
   ```

### Additional Notes:
- Ensure that the `to_application` and `to_function_lib` methods exist and are appropriately implemented within your class.
- Adjust the UUID generation method according to your needs, especially if you want more control over the token length.

PR #1923, opened 2024-12-27
feat: The workflow node cannot obtain values from other nodes, which is set as the default value

shaohuzhang1:
feat: The workflow node cannot obtain values from other nodes, which is set as the default value

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -693,33 +693,43 @@ class WorkflowManage:
         else:
             return self.get_node_by_id(node_id).get_reference_field(fields)
 
-    def generate_prompt(self, prompt: str):
-        """
-        格式化生成提示词
-        @param prompt: 提示词信息
-        @return: 格式化后的提示词
-        """
+    def get_workflow_content(self):
         context = {
             'global': self.context,
         }
 
         for node in self.node_context:
-            properties = node.node.properties
+            context[node.id] = node.context
+        return context
+
+    def reset_prompt(self, prompt: str):
+        placeholder = "{}"
+        for node in self.flow.nodes:
+            properties = node.properties
             node_config = properties.get('config')
             if node_config is not None:
                 fields = node_config.get('fields')
                 if fields is not None:
                     for field in fields:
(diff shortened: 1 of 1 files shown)


apps/application/flow/workflow_manage.py
```diff
@@ -693,33 +693,43 @@ def get_reference_field(self, node_id: str, fields: List[str]):
...
+        """
+        格式化生成提示词
+        @param prompt: 提示词信息
+        @return: 格式化后的提示词
+        """
+        context = self.get_workflow_content()
+        prompt = self.reset_prompt(prompt)
         prompt_template = PromptTemplate.from_template(prompt, template_format='jinja2')
         value = prompt_template.format(context=context)
         return value
```
shaohuzhang1, line 735: There are several issues and optimizations to address in the provided code:

**Issues:**
1. **Method Name Conflicts:** The `generate_prompt` method name seems duplicated with another one that is intended to be used after setting up the workflow content. This should ideally only exist once, or better yet renamed.
2. **Context Merging Logic:** The logic for merging contexts from nodes into the `context` dictionary is mixed within different methods. It would make sense to consolidate this logic into a single place.
3. **Global Variable Handling:** The handling of global variables using placeholders (`{}`) does not ensure that all missing keys return an empty string, but rather raises a TypeError when trying to access them.
4. **Template Replacement Efficiency:** Replacing labels multiple times may lead to unnecessary overhead.

**Optimization Suggestions:**
1. **Single Method Naming:** Rename `generate_prompt` to something more descriptive like `format_prompt`.
2. **Consolidate Context Merging:** Combine the context merging logic into a single function to improve readability and maintainability.
3. **Handle Missing Globals Gracefully:** Modify the script to handle missing global variables gracefully without raising errors.

Here's the revised code addressing these points:

```python
@@ -693,33 +693,43 @@ def get_reference_field(self, node_id: str, fields: List[str]):


-    def generate_prompt(self, prompt: str):
+    def format_prompt(self, prompt: str) -> str:
         """
         格式化生成提示词
         @param prompt: 提示词信息
         @return: 格式化后的提示词
         """
+        
```

This change renames the method slightly for clarity while keeping consistent naming conventions throughout the file.

PR #1924, opened 2024-12-27
feat: The form node supports obtaining data from other nodes

shaohuzhang1:
feat: The form node supports obtaining data from other nodes

### DIFF
--- a/apps/application/flow/step_node/form_node/impl/base_form_node.py
+++ b/apps/application/flow/step_node/form_node/impl/base_form_node.py
@@ -55,8 +55,10 @@ class BaseFormNode(IFormNode):
                         "chat_record_id": self.flow_params_serializer.data.get("chat_record_id"),
                         "is_submit": self.context.get("is_submit", False)}
         form = f'<form_rander>{json.dumps(form_setting)}</form_rander>'
+        context = self.workflow_manage.get_workflow_content()
+        form_content_format = self.workflow_manage.reset_prompt(form_content_format)
         prompt_template = PromptTemplate.from_template(form_content_format, template_format='jinja2')
-        value = prompt_template.format(form=form)
+        value = prompt_template.format(form=form, context=context)
         return NodeResult(
             {'result': value, 'form_field_list': form_field_list, 'form_content_format': form_content_format}, {},
             _write_context=write_context)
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/form_node/impl/base_form_node.py
```diff
@@ -81,8 +85,10 @@ def get_details(self, index: int, **kwargs):
...
                         "is_submit": self.context.get("is_submit", False)}
         form = f'<form_rander>{json.dumps(form_setting)}</form_rander>'
+        context = self.workflow_manage.get_workflow_content()
+        form_content_format = self.workflow_manage.reset_prompt(form_content_format)
         prompt_template = PromptTemplate.from_template(form_content_format, template_format='jinja2')
-        value = prompt_template.format(form=form)
+        value = prompt_template.format(form=form, context=context)
         return {
             'name': self.node.properties.get('stepName'),
             "index": index,
```
shaohuzhang1, line 94: The provided code has some issues that need to be addressed:

1. **Potential Null Pointer Exceptions**: The code uses `self.context.get("is_submit")`, which can lead to a null pointer exception if the key is not found. Consider adding default values or handling missing keys explicitly.

2. **Unnecessary JSON Serialization**: The code serializes `form_setting` directly into a string within the HTML template using `<form_rander>`. This isn't necessary and can introduce complexity since `form_setting` is already serialized by `json.dumps` before usage.

3. **Lack of Input Validation**: There's no explicit validation on input parameters such as `form_field_list`, `form_content_format`, or `form_data`. This could help prevent unexpected behavior.

Here are some optimizations and suggestions:

### Optimizations and Recommendations

#### Validate Required Arguments
Ensure that all required arguments (`form_field_list`, etc.) are provided and handle exceptions where they might be missing.

```python
from typing import Any, Dict

class CustomNode(Node):
    def execute(self, form_field_list: list[Any], form_content_format: str, form_data: dict[str, Any], **kwargs) -> NodeResult:
        try:
            is_submit = kwargs.get("is_submit", False)
            
            # Validate inputs
            if not form_field_list:
                raise ValueError("Missing form field list")
            if not isinstance(form_settings, (dict, list)):  # Adjust based on type expected
                raise TypeError("Invalid form settings type")
            if not isinstance(form_data, dict):  # Ensure only dictionaries are processed
                raise TypeError("Form data must be a dictionary")
        
            form_setting_str = json.dumps(form_settings)

            context = self.workflow_manage.get_workflow_content()
            form_content_format = self.workflow_manage.reset_prompt(form_content_format)
            prompt_template = PromptTemplate.from_template(form_content_format, template_format='jinja2')

            value = prompt_template.format(form=form_setting_str, context=context)
            return NodeResult({'result': value, 'form_field_list': form_field_list, 'form_content_format': form_content_format}, {}, _write_context=write_context)

```

**Key Changes**:
- Added checks to ensure input types and contents meet expectations.
- Used error messages instead of raising exceptions globally for better feedback during debugging.

#### Simplify HTML Templating
If you're dynamically creating HTML forms, consider simplifying it further. The template should focus on formatting and content rather than including serialization steps.

```python
html_form_template = """
<FORM>
  <INPUT id="input1" value="{{ form.form_field_list }}" type="text">
</FORM>
"""

# Render form using jinja2 template engine
context = {"form": {"form_field_list": ["value1", "value2"]}}
html_form = html_form_template.render(context)
```

By addressing these points, you make the code more robust, readable, and maintainable.

PR #1925, opened 2024-12-27
feat: AI dialogue roles support obtaining data from other nodes

shaohuzhang1:
feat: AI dialogue roles support obtaining data from other nodes

### DIFF
--- a/apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
+++ b/apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
@@ -110,6 +110,8 @@ class BaseChatNode(IChatNode):
         self.context['history_message'] = history_message
         question = self.generate_prompt_question(prompt)
         self.context['question'] = question.content
+        system = self.workflow_manage.generate_prompt(system)
+        self.context['system'] = question.content
         message_list = self.generate_message_list(system, prompt, history_message)
         self.context['message_list'] = message_list
         if stream:
@@ -156,7 +158,7 @@ class BaseChatNode(IChatNode):
             'name': self.node.properties.get('stepName'),
             "index": index,
             'run_time': self.context.get('run_time'),
-            'system': self.node_params.get('system'),
+            'system': self.context.get('system'),
             'history_message': [{'content': message.content, 'role': message.type} for message in
                                 (self.context.get('history_message') if self.context.get(
                                     'history_message') is not None else [])],
--- a/apps/application/flow/step_node/question_node/impl/base_question_node.py
+++ b/apps/application/flow/step_node/question_node/impl/base_question_node.py
@@ -92,6 +92,8 @@ class BaseQuestionNode(IQuestionNode):
         self.context['history_message'] = history_message
         question = self.generate_prompt_question(prompt)
         self.context['question'] = question.content
+        system = self.workflow_manage.generate_prompt(system)
+        self.context['system'] = question.content
         message_list = self.generate_message_list(system, prompt, history_message)
         self.context['message_list'] = message_list
         if stream:
@@ -138,7 +140,7 @@ class BaseQuestionNode(IQuestionNode):
             'name': self.node.properties.get('stepName'),
             "index": index,
             'run_time': self.context.get('run_time'),
-            'system': self.node_params.get('system'),
+            'system': self.context.get('system'),
             'history_message': [{'content': message.content, 'role': message.type} for message in
                                 (self.context.get('history_message') if self.context.get(
                                     'history_message') is not None else [])],


apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
```diff
@@ -156,7 +158,7 @@ def get_details(self, index: int, **kwargs):
             'name': self.node.properties.get('stepName'),
             "index": index,
             'run_time': self.context.get('run_time'),
-            'system': self.node_params.get('system'),
+            'system': self.context.get('system'),
             'history_message': [{'content': message.content, 'role': message.type} for message in
                                 (self.context.get('history_message') if self.context.get(
                                     'history_message') is not None else [])],
```
shaohuzhang1, line 164: The provided C# code snippet appears to be part of an AI dialog execution function within a workflow management system. Here are some potential improvements and suggestions:

### Potential Issues:
1. **Duplicate `system` Variable Assignment**: The `system` variable is assigned twice before use. This can lead to confusion or unintended behavior because the second assignment does not change the original value.
   ```csharp
   system = this.WorkflowManage.GeneratePrompt(system);
   self.context['system'] = question.content;
   ```

2. **Unnecessary Condition Check**: After assigning `self.context['system']`, there's an unnecessary condition check to ensure `question.content` is not `None`.
   ```csharp
   'system': self.context.get('system'),
   ...

   if stream:
       ...
```
3. **Context Key Naming Confusion**: There seems to be a mix-up between using `self.context['system']` and directly accessing `node_params['system']`. It is unclear which one should be preferred, but it could cause errors if `node_params` contains conflicting keys.

### Optimization Suggestions:
1. **Remove Duplicate Logic**: Since both assignments assign `question.content`, remove the duplicate line:
   ```csharp
   // Remove this line as it duplicates what follows
   // system = this.WorkflowManage.GeneratePrompt(system);  
   ```

2. **Simplify Stream Handling**: Ensure that the rest of the logic handling `stream` does not rely on whether `question.content` exists.
   ```csharp
   if (stream) {
       ...
   } else {
       // Handle streaming case differently, e.g., skip it if not available.
   }
   ```

3. **Consistent Access Patterns**: Decide whether to access context via `self.context[key]` or through specific method calls (`node_properties`, etc.). Consistency will make the code easier to maintain.
   ```csharp
   // Could also be done like so: 
   'name': self.node.GetStepName()
   ```

Here’s an improved version of the code incorporating these suggestions:

```csharp
@@ -110,6 +110,7 @@ def execute(self, model_id, system, prompt, dialogue_number, history_chat_record

         self.context['history_message'] = history_message
         question = self.generate_prompt_question(prompt)
         self.context['question'] = question.content
         
         system = self.workflow_manage.generate_prompt(system)
         self.context['system'] = question.content
        
         message_list = self.generate_message_list(system, prompt, history_message)
         self.context['message_list'] = message_list
         
         if stream: 
@@ -158,9 +162,10 @@ def get_details(self, index: int, **kwargs):
             # Add more detail retrieval here if needed
     }
   
   ```
This updated version removes the redundant assignment of `system`, simplifies the stream handling block, and aims for consistent key access patterns.

File: apps/application/flow/step_node/question_node/impl/base_question_node.py [hunk 2]
```diff
@@ -138,7 +140,7 @@ def get_details(self, index: int, **kwargs):
             'name': self.node.properties.get('stepName'),
             "index": index,
             'run_time': self.context.get('run_time'),
-            'system': self.node_params.get('system'),
+            'system': self.context.get('system'),
             'history_message': [{'content': message.content, 'role': message.type} for message in
                                 (self.context.get('history_message') if self.context.get(
                                     'history_message') is not None else [])],
```
shaohuzhang1, line 146: There are a few issues in the code you've provided:

1. **Incorrect Use of `workflow_manage`**: The line `system = self.workflow_manage.generate_prompt(system)` seems to be intended to generate a new system string based on existing one, but there's no context about where this method exists. It should either belong to another class instance or be part of an imported module.

2. **Redundant Context Update**: You're updating the context with the same value twice for `system`: once at the beginning (`self.context['system']`) and again after generating it (`if stream:` block). This might introduce redundancy unless the workflow manages both contexts differently.

3. **List Comprehension Formatting**: In both instances where you create lists from dictionaries, consider using more compact syntax for readability, e.g., `[{'content': m.content, 'role': m.type} for m in messages]`.

Overall, make sure that `workflow_manage` has its own methods appropriately defined and importable within your current script so that these lines can work without errors. If they need to use external data structures or functions, ensure those components are available and properly structured according to their requirements.

PR #1938, opened 2024-12-30
fix: The functions in the library cannot be modified after being defined

shaohuzhang1:
fix: The functions in the library cannot be modified after being defined

### DIFF
--- a/ui/src/components/codemirror-editor/index.vue
+++ b/ui/src/components/codemirror-editor/index.vue
@@ -22,7 +22,11 @@ defineOptions({ name: 'CodemirrorEditor' })
 function getRangeFromLineAndColumn(state: any, line: number, column: number, end_column?: number) {
   const l = state.doc.line(line)
   const form = l.from + column
-  return { form: form > l.to ? l.to : form, to: end_column ? l.from + end_column : l.to }
+  const to_end_column = l.from + end_column
+  return {
+    form: form > l.to ? l.to : form,
+    to: end_column && to_end_column < l.to ? to_end_column : l.to
+  }
 }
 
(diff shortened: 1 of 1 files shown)


ui/src/components/codemirror-editor/index.vue
```diff
@@ -22,7 +22,11 @@ defineOptions({ name: 'CodemirrorEditor' })
...
   const form = l.from + column
-  return { form: form > l.to ? l.to : form, to: end_column ? l.from + end_column : l.to }
+  const to_end_column = l.from + end_column
+  return {
+    form: form > l.to ? l.to : form,
+    to: end_column && to_end_column < l.to ? to_end_column : l.to
+  }
 }
 
 const regexpLinter = linter(async (view) => {
```
shaohuzhang1, line 32: The provided code snippet has some minor optimizations and corrections that can be suggested:

1. **Variable Naming Consistency**: The variable `to_end_column` is unnecessarily defined before use, which could be removed.
   
2. **Condition Simplification**: The condition in the return statement of `getRangeFromLineAndColumn` can be simplified using logical AND (`&&`) instead of multiple comparison checks.

Here's an optimized version of the function:

```javascript
function getRangeFromLineAndColumn(state: any, line: number, column: number, end_column?: number): { form: number, to: number } {
  const l = state.doc.line(line);
  const form = l.from + column;
  
  let rangeTo =
    end_column !== undefined && end_column < l.to ?
      l.from + end_column :
      l.to;

  return {
    form: form > l.to ? l.to : form,
    to: rangeTo
  };
}
```

These changes reduce redundancy and improve readability while maintaining correct functionality.

PR #1944, opened 2024-12-30
fix: Model editing prohibits modifying model types

shaohuzhang1:
fix: Model editing prohibits modifying model types

### DIFF
--- a/ui/src/views/template/component/EditModel.vue
+++ b/ui/src/views/template/component/EditModel.vue
@@ -79,6 +79,7 @@
             <span>模型类型</span>
           </template>
           <el-select
+            disabled
             v-loading="model_type_loading"
             @change="list_base_model($event, true)"
             v-model="base_form_data.model_type"


ui/src/views/template/component/EditModel.vue
```diff
@@ -79,6 +79,7 @@
             <span>模型类型</span>
           </template>
           <el-select
+            disabled
             v-loading="model_type_loading"
             @change="list_base_model($event, true)"
             v-model="base_form_data.model_type"
```
shaohuzhang1, line 85: - The `disabled` attribute on the `<el-select>` element is set incorrectly. It should be outside of the template but within the component's script section or directly above the `v-model`. Adjust it to:
```javascript
props: {
  base_form_data: Object,
},
created() {

}, methods: {
...
```
Then update the HTML as follows:
```html
<template>
  <!-- ... -->
  <el-select :disabled="true" v-loading="model_type_loading">
    ...
  </el-select>
</template>
```

**Optimization Suggestion:** Consider using computed properties instead of local data bindings wherever possible to improve performance and readability.

---

### Summary of Changes:

1. **Correct Placement of `disabled`:**
   - Ensure that the `disabled` property is defined correctly in the parent/component scope (`data`, `computed`, `methods`) where the Vue instance lives.
   
2. **Computation Property Usage:**
   - Use computed properties if you have multiple places in your code where this value remains constant or updates regularly.
  
Feel free to ask if you need further clarification on these changes!

PR #1945, opened 2024-12-30
fix: Export application error prompt repeated

shaohuzhang1:
fix: Export application error prompt repeated

### DIFF
--- a/ui/src/views/application/index.vue
+++ b/ui/src/views/application/index.vue
@@ -226,9 +226,11 @@ function settingApplication(row: any) {
 }
 const exportApplication = (application: any) => {
   applicationApi.exportApplication(application.id, application.name, loading).catch((e) => {
-    e.response.data.text().then((res: string) => {
-      MsgError(`导出失败:${JSON.parse(res).message}`)
-    })
+    if (e.response.status !== 403) {
+      e.response.data.text().then((res: string) => {
+        MsgError(`导出失败:${JSON.parse(res).message}`)
+      })
+    }
   })
 }
 const importApplication = (file: any) => {


ui/src/views/application/index.vue
```diff
@@ -226,9 +226,11 @@ function settingApplication(row: any) {
...
-      MsgError(`导出失败:${JSON.parse(res).message}`)
-    })
+    if (e.response.status !== 403) {
+      e.response.data.text().then((res: string) => {
+        MsgError(`导出失败:${JSON.parse(res).message}`)
+      })
+    }
   })
 }
 const importApplication = (file: any) => {
```
shaohuzhang1, line 236: 1. **Missing Handling for Status Code `403`:** The current implementation includes an `if` statement that only handles status codes other than `403`. However, it's essential to handle status code `403` specifically because it means there's a permission issue with exporting the application.

2. **Handling Other Error Codes:** While your implementation already catches all types of errors under `e.response`, you might want to add specific handling for other error statuses like `500` (Server Internal Error), `404` (Not Found), etc., based on your application requirements.

3. **Refactoring Comments and Variables:** You could refactor some comments and variable names for clarity:

```javascript
const exportApplication = async (application: any) => { // Use async/await for cleaner error handling
  try {
    await applicationApi.exportApplication(application.id, application.name);
  } catch (err) {
    if (err.response && err.response.status === 403) {
      MsgError('Permission denied to export application');
    } else {
      MsgError(`Export failed: ${JSON.stringify(err.response ? err.response.data : 'Unknown error')}`);
    }
  }
};

// Similar pattern can be applied to importApplication function
```

4. **Consider Using Axios Interceptors:** If `applicationApi.exportApplication` is using Axios, you can also use Axios interceptors to globally handle responses such as logging errors, retrying requests, or updating UI states without having multiple instances of catching and responding to errors in each API call.

5. **Ensure Consistent Response Formatting:** Always ensure that when sending errors back from the server, they include consistent metadata such as status code and message format so that clients expect a standardized response.

These modifications should improve robustness and readability of your code while addressing potential issues related to user permissions, general error management, and possibly integration with other APIs or libraries.

PR #1953, opened 2024-12-31
fix: tts node and stt node add warning

shaohuzhang1:
fix: tts node and stt node add warning  --bug=1050722 --user=王孝刚 【应用编排】前置节点删掉后，调试和发布的时候引用前置节点的变量要有提示 https://www.tapd.cn/57709429/s/1639310

### DIFF
--- a/ui/src/workflow/nodes/speech-to-text-node/index.vue
+++ b/ui/src/workflow/nodes/speech-to-text-node/index.vue
@@ -145,6 +145,7 @@ import { app } from '@/main'
 import useStore from '@/stores'
 import NodeCascader from '@/workflow/common/NodeCascader.vue'
 import type { FormInstance } from 'element-plus'
+
 const { model } = useStore()
 
 const {
@@ -156,8 +157,12 @@ const modelOptions = ref<any>(null)
 const providerOptions = ref<Array<Provider>>([])
 
 const aiChatNodeFormRef = ref<FormInstance>()
+const nodeCascaderRef = ref()
 const validate = () => {
-  return aiChatNodeFormRef.value?.validate().catch((err) => {
+  return Promise.all([
+    nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    aiChatNodeFormRef.value?.validate()
+  ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }
--- a/ui/src/workflow/nodes/text-to-speech-node/index.vue
+++ b/ui/src/workflow/nodes/text-to-speech-node/index.vue
@@ -172,8 +172,12 @@ const modelOptions = ref<any>(null)
 const providerOptions = ref<Array<Provider>>([])
 
 const aiChatNodeFormRef = ref<FormInstance>()
+const nodeCascaderRef = ref()
 const validate = () => {
-  return aiChatNodeFormRef.value?.validate().catch((err) => {
+  return Promise.all([
+    nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    aiChatNodeFormRef.value?.validate()
+  ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }


ui/src/workflow/nodes/speech-to-text-node/index.vue
```diff
@@ -156,8 +157,12 @@ const modelOptions = ref<any>(null)
...
+const nodeCascaderRef = ref()
 const validate = () => {
-  return aiChatNodeFormRef.value?.validate().catch((err) => {
+  return Promise.all([
+    nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    aiChatNodeFormRef.value?.validate()
+  ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }
```
shaohuzhang1, line 168: The provided code is mostly correct, but there are a few improvements that can be made:

1. **Avoid Using Await with `Promise.all`:** The use of `await` inside the catch block will prevent further execution after an error. Replace it with `.then()` to continue processing.

2. **Ensure Cascader Validations:** Ensure that cascader validations handle cases where the reference might not exist (as it does now).

Here's the revised version of the code:

```javascript
@@ -145,6 +145,7 @@ import { app } from '@/main'
 import NodeCascader from '@/workflow/common/NodeCascader.vue'
 import type { FormInstance } from 'element-plus'

+import { onMounted, ref } from 'vue';
+
 const { model } = useStore();

 // Assuming props.nodeModel exists
 const modelOptions = ref<any[]>([]);
 const providerOptions = ref<Array<Provider>>([]);

 const aiChatNodeFormRef = ref<FormInstance>();
+const nodeCascaderRef = ref(null); // Initialize as null

// Assume this function initializes the form and cascade components
function initializeFormAndCascade() {
  onMounted(() => {
    if (nodeCascaderRef.value) {
      nodeCascaderRef.value.resetFields();
    }
    
    if (aiChatNodeFormRef.value) {
      aiChatNodeFormRef.value.clearValidate()
    }
  });
}

const validate = () => {
  return new Promise((resolve) => {
    Promise.all([
        nodeCascaderRef.value ? nodeCascaderRef.value.validate() : "",
        aiChatNodeFormRef.value?.validate(e => resolve(!e))
    ])
    .catch(err => reject({ node: props.nodeModel, errMessage: err }))
  })
}
```

### Key Changes:
- **Await Removal:** Replaced async/await syntax with traditional promises using `new Promise`.
- **Null Initialization:** Initialized `nodeCascaderRef` as `null`. This ensures we don't rely on optional chaining (`?`) unnecessarily and avoids potential runtime errors.
- **Initialization Hook:** Added a setup function called `initializeFormAndCascade`, which uses Vue's `onMounted` hook to ensure that any necessary initialization logic runs once the component has mounted.

This should address most performance concerns and improve stability.

ui/src/workflow/nodes/text-to-speech-node/index.vue
```diff
@@ -172,8 +172,12 @@ const modelOptions = ref<any>(null)
...
+const nodeCascaderRef = ref()
 const validate = () => {
-  return aiChatNodeFormRef.value?.validate().catch((err) => {
+  return Promise.all([
+    nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    aiChatNodeFormRef.value?.validate()
+  ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }
```
shaohuzhang1, line 183: There are no critical errors in the provided code. However, I suggest adding error handling to ensure that the `nodeCascader` validation does not prevent the form from being validated at all. Here’s an optimized version:

```javascript
const modelOptions = ref<any>(null);
const providerOptions = ref<Array<Provider>>([]);

const aiChatNodeFormRef = ref<FormInstance>();

function validate(): Promise<{ message?: string }> {
  return Promise.race([
    (async () => await (nodeCascaderRef.value && nodeCascaderRef.value.validate()))(),
    (async () => await (aiChatNodeFormRef.value && aiChatNodeFormRef.value.validate()))()
  ])
  .catch((err) => ({ message: err }));
}
```

### Key Changes:
1. **Error Handling**: Used `.catch((err) => ({ message: err }))` to handle any errors and provide a consistent response format.
2. **Promise.all with Race Condition**: Utilized `Promise.race()` to ensure that if either validator fails, it is caught quickly without impacting the other.
3. **Async/Await Simplification**: Slightly simplified the promise chaining for better readability.

This approach ensures that both validators run concurrently, but stops execution on the first successful result or the first error encountered.

PR #1958, opened 2024-12-31
fix: Generate a problem with continuous clicking without response

shaohuzhang1:
fix: Generate a problem with continuous clicking without response

### DIFF
--- a/apps/dataset/serializers/paragraph_serializers.py
+++ b/apps/dataset/serializers/paragraph_serializers.py
@@ -728,7 +728,6 @@ class ParagraphSerializers(ApiMixin, serializers.Serializer):
         dataset_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("知识库id"))
         document_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("文档id"))
 
-        @transaction.atomic
         def batch_generate_related(self, instance: Dict, with_valid=True):
             if with_valid:
                 self.is_valid(raise_exception=True)


apps/dataset/serializers/paragraph_serializers.py
```diff
@@ -728,7 +728,6 @@ class BatchGenerateRelated(ApiMixin, serializers.Serializer):
         dataset_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("知识库id"))
         document_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("文档id"))
 
-        @transaction.atomic
         def batch_generate_related(self, instance: Dict, with_valid=True):
             if with_valid:
                 self.is_valid(raise_exception=True)
```
shaohuzhang1, line 733: The code looks generally correct for handling a bulk generation of related items within a specific dataset and document using Django's `serializers` framework. However, there are a few minor points to consider:

1. **Transaction Management**: The use of the `@transaction.atomic` decorator is appropriate given that this method performs multiple operations that need to be atomic (i.e., all or none). It ensures that either everything succeeds completely, or nothing is saved.

2. **Method Signature**: While it doesn't affect functionality, the method signature includes an optional parameter `with_valid`, which is not used anywhere in the method body. You might want to remove this parameter if it is unused. Alternatively, you can set a default value like so: `def batch_generate_related(instance: Dict, with_valid=False):`.

3. **Type Hinting**: The type hint for `instance` is missing from the signature. Adding `-> None` after the closing parenthesis would make the function more explicit about its return type.

Here’s an optimized version of the code incorporating these considerations:

```python
@@ -728,7 +728,6 @@ class BatchGenerateRelated(ApiMixin, serializers.Serializer):
         dataset_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("知识库id"))
         document_id = serializers.UUIDField(required=True, error_messages=ErrMessage.uuid("文档id"))

-        @transaction.atomic
         def batch_generate_related(self, instance=None, with_valid=False):
             if with_valid:
                 # Perform validity checks here
```

This change enhances clarity and completeness in the function's design while maintaining its intended behavior.

PR #2006, opened 2025-01-10
refactor: support volcanic engine embeddings

shaohuzhang1:
refactor: support volcanic engine embeddings

### DIFF
--- a/apps/setting/models_provider/impl/volcanic_engine_model_provider/credential/embedding.py
+++ b/apps/setting/models_provider/impl/volcanic_engine_model_provider/credential/embedding.py
@@ -14,7 +14,7 @@ from common.forms import BaseForm
 from setting.models_provider.base_model_provider import BaseModelCredential, ValidCode
 
 
-class OpenAIEmbeddingCredential(BaseForm, BaseModelCredential):
+class VolcanicEmbeddingCredential(BaseForm, BaseModelCredential):
     def is_valid(self, model_type: str, model_name, model_credential: Dict[str, object], model_params, provider,
                  raise_exception=True):
         model_type_list = provider.get_model_type_list()
--- a/apps/setting/models_provider/impl/volcanic_engine_model_provider/model/embedding.py
+++ b/apps/setting/models_provider/impl/volcanic_engine_model_provider/model/embedding.py
@@ -1,15 +1,16 @@
 from typing import Dict
 
-from langchain_community.embeddings import VolcanoEmbeddings
+from langchain_openai import OpenAIEmbeddings
 
 from setting.models_provider.base_model_provider import MaxKBBaseModel
 
 
-class VolcanicEngineEmbeddingModel(MaxKBBaseModel, VolcanoEmbeddings):
+class VolcanicEngineEmbeddingModel(MaxKBBaseModel, OpenAIEmbeddings):
     @staticmethod
     def new_instance(model_type, model_name, model_credential: Dict[str, object], **model_kwargs):
         return VolcanicEngineEmbeddingModel(
-            api_key=model_credential.get('api_key'),
+            openai_api_key=model_credential.get('api_key'),
             model=model_name,
             openai_api_base=model_credential.get('api_base'),
+            check_embedding_ctx_length=False,
         )
--- a/apps/setting/models_provider/impl/volcanic_engine_model_provider/volcanic_engine_model_provider.py
+++ b/apps/setting/models_provider/impl/volcanic_engine_model_provider/volcanic_engine_model_provider.py
@@ -14,10 +14,12 @@ from setting.models_provider.base_model_provider import IModelProvider, ModelPro
 from setting.models_provider.impl.openai_model_provider.credential.embedding import OpenAIEmbeddingCredential
 from setting.models_provider.impl.openai_model_provider.credential.llm import OpenAILLMModelCredential
 from setting.models_provider.impl.openai_model_provider.model.embedding import OpenAIEmbeddingModel
+from setting.models_provider.impl.volcanic_engine_model_provider.credential.embedding import VolcanicEmbeddingCredential
 from setting.models_provider.impl.volcanic_engine_model_provider.credential.image import \
     VolcanicEngineImageModelCredential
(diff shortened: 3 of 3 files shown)


apps/setting/models_provider/impl/volcanic_engine_model_provider/volcanic_engine_model_provider.py
```diff
@@ -96,6 +99,8 @@
     .append_default_model_info(model_info_list[2])
     .append_default_model_info(model_info_list[3])
     .append_default_model_info(model_info_list[4])
+    .append_model_info_list(model_info_embedding_list)
+    .append_default_model_info(model_info_embedding_list[0])
     .build()
 )
```
shaohuzhang1, line 106: The provided code includes several issues that should be addressed:

1. **Redundant Code**: The `model_info_embedding_list` is defined twice with identical values. Remove one of these instances to avoid redundancy.

2. **Variable Naming**: Use descriptive variable names instead of short abbreviations like `open_ai_embedding_credential`. This makes the code more readable and maintainable.

3. **Credential Management**: Ensure that each model type has its own separate credential class (`OpenAILLMModelCredential`, etc.) if they require different configurations or access methods.

Here’s a revised version of the code with these changes applied:

```python
# Import necessary classes
from setting.models_provider.impl.openai_model_provider.credential.openai_llm import \
    OpenAILLMModelCredential

# Define constants
from enums.model_type_const import ModelTypeConst

class ModelInfo:
    def __init__(self, endpoint_id: str, description: str,
                 model_type: ModelTypeConst, credential, model):
        self.endpoint_id = endpoint_id
        self.description = description
        self.model_type = model_type
        self.credential = credential
        self.model = model

class ModelInfoManager:
    @staticmethod
    def builder():
        manager = ModelInfoManager()
        return manager

    def append_default_model_info(self, model_info):
        # Implementation details for appending default model info...
        pass

    def append_model_info_list(self, model_infos):
        # Implementation details for appending multiple model infos...
        pass

def create_openai_credential() -> OpenAILLMModelCredential:
    return OpenAILLMModelCredential()

def configure_models():
    # Define OpenAI embedding credentials and models
    open_ai_emb_cred = create_openai_credential()
    emb_model = OpenAITextEmbeddingModel()

    # Append information about OpenAI embedding to the list
    model_info_list = [
        ModelInfo('ep-xxxxxxxxxyyyyyy',
                    'Visit VolcanoArk for accessing inference points on the user-facing page.',
                    ModelTypeConst.EMBEDDING, open_ai_emb_cred, emb_model),
    ]

    # Initialize and build ModelInfoManager
    model_manager = ModelInfoManager.builder(). \
                   append_default_model_info(model_info_list[0]). \
                   append_model_info_list(model_info_list).build()

    return model_manager

# Call function to configure and generate model settings
configure_models()
```

### Key Changes Made:
- Removed redundant `model_info_embedding_list`.
- Replaced `open_ai_embedding_credential` with an alias `open_ai_emb_cred` for better readability.
- Introduced `Enum` constants for model types using a helper file named `enums/model_type_const.py`.

Make sure you have the corresponding imports and definitions for the enum constant and other related classes based on your actual project setup.

apps/setting/models_provider/impl/volcanic_engine_model_provider/model/embedding.py
```diff
@@ -1,15 +1,16 @@
...
+class VolcanicEngineEmbeddingModel(MaxKBBaseModel, OpenAIEmbeddings):
     @staticmethod
     def new_instance(model_type, model_name, model_credential: Dict[str, object], **model_kwargs):
         return VolcanicEngineEmbeddingModel(
-            api_key=model_credential.get('api_key'),
+            openai_api_key=model_credential.get('api_key'),
             model=model_name,
             openai_api_base=model_credential.get('api_base'),
+            check_embedding_ctx_length=False,
         )
```
shaohuzhang1, line 16: 1. The `VolcanoEmbeddings` class is being imported instead of its correct name in the chain community library (`langchain_community`). This should be updated to `OpenAIEmbeddings`.
2. In the `new_instance` method, when creating an instance of `VolcanicEngineEmbeddingModel`, it's currently passing through a parameter named `openai_api_base`. However, this parameter doesn't appear to be used anywhere within the model implementation. You might consider removing or using this parameter as needed.

Optimization suggestions:
- Ensure that all necessary imports are at the top of the file.
- Consider using `dataclasses` or another data structure for the model parameters if they remain consistent across models.
- If certain properties like `check_embedding_ctx_length` can default to False without affecting functionality, you might set them to a constant value to simplify maintenance.

apps/setting/models_provider/impl/volcanic_engine_model_provider/credential/embedding.py
```diff
@@ -14,7 +14,7 @@
 from setting.models_provider.base_model_provider import BaseModelCredential, ValidCode
 
 
-class OpenAIEmbeddingCredential(BaseForm, BaseModelCredential):
+class VolcanicEmbeddingCredential(BaseForm, BaseModelCredential):
     def is_valid(self, model_type: str, model_name, model_credential: Dict[str, object], model_params, provider,
                  raise_exception=True):
         model_type_list = provider.get_model_type_list()
```
shaohuzhang1, line 20: There are a few issues with this code:

1. The class name should probably be "VolcanicEmbeddingCredential" instead of just "OpenAIEmbeddingCredential".
2. There seems to be an inconsistency with the import statement at line -8. It imports `ValidCode`, but elsewhere it is used as `base_valid_code`.
3. You might want to consider renaming the variable names like `model_name` to something more descriptive (e.g., `embedding_service_name`).

Here's an updated version of the code:
```python
from setting.models_provider.base_model_provider import BaseModelCredential

# Assuming base_valid_code is a valid attribute in the BaseForm class or has been defined elsewhere
class VolcanicEmbeddingCredential(BaseForm, BaseModelCredential):
    def is_valid(self, model_type: str, embedding_service_name: str, model_credential: Dict[str, object],
                  model_params, provider) -> bool:
        # Check if model type is part of supported types by the provider
        return model_type in provider.get_model_type_list()
```

Additionally, you might want to add some logging to understand what is happening where things go wrong during runtime.

cnChenKai:
感谢。
Thanks for your work!

PR #2014, opened 2025-01-13
fix: Typos

shaohuzhang1:
fix: Typos

### DIFF
--- a/ui/src/views/application/component/AccessSettingDrawer.vue
+++ b/ui/src/views/application/component/AccessSettingDrawer.vue
@@ -334,7 +334,7 @@ const submit = async () => {
         applicationApi
           .updatePlatformConfig(id, configType.value, form[configType.value], loading)
           .then(() => {
-            MsgSuccess(t('commom.saveSuccess'))
+            MsgSuccess(t('common.saveSuccess'))
             closeDrawer()
             emit('refresh')
           })


ui/src/views/application/component/AccessSettingDrawer.vue
```diff
@@ -334,7 +334,7 @@ const submit = async () => {
         applicationApi
           .updatePlatformConfig(id, configType.value, form[configType.value], loading)
           .then(() => {
-            MsgSuccess(t('commom.saveSuccess'))
+            MsgSuccess(t('common.saveSuccess'))
             closeDrawer()
             emit('refresh')
           })
```
shaohuzhang1, line 340: The provided JavaScript code snippet looks mostly clean, but there's one potential issue that could be addressed:

### Potential Issue:
The `form` object is missing from the parameter list for the `applicationApi.updatePlatformConfig` method call. This means that the correct property name used to access the form data might not match what is expected.

### Optimization Suggestions:
- **Check Form Data Access**: Ensure that the `form` object contains the necessary properties to update the platform configuration. If it doesn't, you may need to modify how the form data is accessed or passed to the API.

Here's the original text repeated for clarity:

---

**Original Text Repeated**

```javascript
@@ -334,7 +334,7 @@
   const submit = async () => {
     applicationApi
       .updatePlatformConfig(id, configType.value, form[configType.value], loading)
       .then(() => {
-        MsgSuccess(t('commom.saveSuccess'))
+        MsgSuccess(t('common.saveSuccess'))
         closeDrawer()
         emit('refresh')
       })
```

---

PR #2022, opened 2025-01-14
feat: Generate i18n file

shaohuzhang1:
feat: Generate i18n file

### DIFF
--- a/apps/dataset/serializers/document_serializers.py
+++ b/apps/dataset/serializers/document_serializers.py
@@ -786,7 +786,7 @@ class DocumentSerializers(ApiMixin, serializers.Serializer):
                     'user_id': openapi.Schema(type=openapi.TYPE_STRING, title=_('user id'), description=_('user id')),
                     'paragraph_count': openapi.Schema(type=openapi.TYPE_INTEGER, title="_('document count')",
                                                       description="_('document count')", default=1),
-                    'is_active': openapi.Schema(type=openapi.TYPE_BOOLEAN, title=_(''),
+                    'is_active': openapi.Schema(type=openapi.TYPE_BOOLEAN, title=_('Is active'),
                                                 description=_('Is active'), default=True),
                     'update_time': openapi.Schema(type=openapi.TYPE_STRING, title=_('update time'),
                                                   description=_('update time'),
--- a/apps/dataset/views/file.py
+++ b/apps/dataset/views/file.py
@@ -24,7 +24,7 @@ class FileView(APIView):
     parser_classes = [MultiPartParser]
 
     @action(methods=['POST'], detail=False)
-    @swagger_auto_schema(operation_summary=_(''),
+    @swagger_auto_schema(operation_summary=_('Upload file'),
                          operation_id=_('Upload file'),
                          manual_parameters=[openapi.Parameter(name='file',
                                                               in_=openapi.IN_FORM,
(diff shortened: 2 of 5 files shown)


apps/dataset/serializers/document_serializers.py
```diff
@@ -786,7 +786,7 @@ def get_response_body_api():
                     'user_id': openapi.Schema(type=openapi.TYPE_STRING, title=_('user id'), description=_('user id')),
                     'paragraph_count': openapi.Schema(type=openapi.TYPE_INTEGER, title="_('document count')",
                                                       description="_('document count')", default=1),
-                    'is_active': openapi.Schema(type=openapi.TYPE_BOOLEAN, title=_(''),
+                    'is_active': openapi.Schema(type=openapi.TYPE_BOOLEAN, title=_('Is active'),
                                                 description=_('Is active'), default=True),
                     'update_time': openapi.Schema(type=openapi.TYPE_STRING, title=_('update time'),
                                                   description=_('update time'),
```
shaohuzhang1, line 792: The provided code is a function `get_response_body_api()` from an API documentation or schema definition. The main issue here seems to be with the placeholder text in the `description` field under the `'is_active'` key.

Specifically:

- In the original snippet, it mentions `_("")`. However, you haven't defined such translations (e.g., `_('Is active')`) or imported necessary libraries for translation. This could potentially cause errors when trying to render these fields.
  
To optimize the description of this field:
- Replace `"_" ""` with something meaningful that describes what the boolean represents. For example, if the feature indicates whether the user record is currently active, use a descriptive phrase like "_('User is active')".

In summary, ensure appropriate error handling and descriptions are included for any placeholder strings, especially those related to user roles or activation states.

apps/dataset/views/file.py
```diff
@@ -24,7 +24,7 @@ class FileView(APIView):
     parser_classes = [MultiPartParser]
 
     @action(methods=['POST'], detail=False)
-    @swagger_auto_schema(operation_summary=_(''),
+    @swagger_auto_schema(operation_summary=_('Upload file'),
                          operation_id=_('Upload file'),
                          manual_parameters=[openapi.Parameter(name='file',
                                                               in_=openapi.IN_FORM,
```
shaohuzhang1, line 30: The code is almost ready and clear. Here are some minor suggestions:

1. **Operation Summary**: The `operation_summary` should be descriptive; 'Upload file' is a good starting point.

2. **Response Codes**: Consider adding response codes to handle different outcomes of the upload process. For example, HTTP status 201 Created if successful or 4XX or 5XX for errors.

3. **Error Handling**: You could add a try-except block to catch exceptions that might occur during file processing (e.g., invalid file format).

Here's a slightly optimized version with these considerations in mind:

```python
from .models import YourModel  # Adjust your model import as needed
import openapi_spec_validator

class FileView(APIView):
    parser_classes = [MultiPartParser]

    @swagger_auto_schema(
        operation_summary=_('Upload file'),
        operation_id=_('Upload file'),
        manual_parameters=[openapi.Parameter(name='file',
                                           in_=openapi.IN_FORM,
                                           description='File to be uploaded.',
                                           required=True,
                                           type=openapi.TYPE_FILE)]
    )
    def post(self, request):
        if not request.FILES.get('file'):
            return Response({'error': 'No file provided'}, status=400)

        try:
            # Process the uploaded file here
            # Example: save it to disk, validate its format, etc.
            filename = request.FILES['file'].name
            content = request.FILES['file'].read()

            # Save this data somewhere
            instance = YourModel(file_content=content)
            instance.save()

            return Response({
                'success': True,
                'filename': filename
            }, status=201)

        except Exception as e:
            openapi_spec_validator.validate(swagger_json=request.swagger_schema.to_dict())
            print(f"An error occurred while uploading the file: {e}")
            return Response({'error': str(e)}, status=500)
```

Make sure to replace `'YourModel'` with the actual name of the model you want to use for storing files. This version also includes basic input validation and exception handling for better robustness.

PR #2026, opened 2025-01-14
feat: i18n

shaohuzhang1:
feat: i18n

### DIFF
--- a/apps/application/flow/step_node/application_node/i_application_node.py
+++ b/apps/application/flow/step_node/application_node/i_application_node.py
@@ -63,7 +63,7 @@ class IApplicationNode(INode):
                 app_audio_list[1:])
             for audio in app_audio_list:
                 if 'file_id' not in audio:
-                    raise ValueError(_("Parameter value error: The uploaded image lacks file_id, and the audio upload fails."))
+                    raise ValueError(_("Parameter value error: The uploaded audio lacks file_id, and the audio upload fails."))
         return self.execute(**self.node_params_serializer.data, **self.flow_params_serializer.data,
                             app_document_list=app_document_list, app_image_list=app_image_list,
                             app_audio_list=app_audio_list,
--- a/apps/application/flow/step_node/function_node/i_function_node.py
+++ b/apps/application/flow/step_node/function_node/i_function_node.py
@@ -25,7 +25,7 @@ class InputField(serializers.Serializer):
     is_required = serializers.BooleanField(required=True, error_messages=ErrMessage.boolean(_("Is this field required")))
     type = serializers.CharField(required=True, error_messages=ErrMessage.char(_("type")), validators=[
         validators.RegexValidator(regex=re.compile("^string|int|dict|array|float$"),
-                                  message=_("字段只支持string|int|dict|array|float"), code=500)
+                                  message=_("The field only supports string|int|dict|array|float"), code=500)
     ])
     source = serializers.CharField(required=True, error_messages=ErrMessage.char(_("source")), validators=[
         validators.RegexValidator(regex=re.compile("^custom|reference$"),
--- a/apps/application/flow/step_node/search_dataset_node/i_search_dataset_node.py
+++ b/apps/application/flow/step_node/search_dataset_node/i_search_dataset_node.py
@@ -24,7 +24,7 @@ class DatasetSettingSerializer(serializers.Serializer):
                                      error_messages=ErrMessage.integer(_("Reference segment number")))
     # 相似度 0-1之间
     similarity = serializers.FloatField(required=True, max_value=2, min_value=0,
-                                        error_messages=ErrMessage.float(_("引用分段数")))
+                                        error_messages=ErrMessage.float(_('similarity')))
     search_mode = serializers.CharField(required=True, validators=[
         validators.RegexValidator(regex=re.compile("^embedding|keywords|blend$"),
                                   message=_("The type only supports register|reset_password"), code=500)
--- a/apps/locales/en_US/LC_MESSAGES/django.po
+++ b/apps/locales/en_US/LC_MESSAGES/django.po
@@ -404,7 +404,7 @@ msgstr ""
 
 #: .\apps\application\flow\step_node\application_node\i_application_node.py:66
 msgid ""
-"Parameter value error: The uploaded image lacks file_id, and the audio "
+"Parameter value error: The uploaded audio lacks file_id, and the audio "
 "upload fails."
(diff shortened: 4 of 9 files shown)


apps/application/flow/step_node/application_node/i_application_node.py
```diff
@@ -63,7 +63,7 @@ def _run(self):
                 app_audio_list[1:])
             for audio in app_audio_list:
                 if 'file_id' not in audio:
-                    raise ValueError(_("Parameter value error: The uploaded image lacks file_id, and the audio upload fails."))
+                    raise ValueError(_("Parameter value error: The uploaded audio lacks file_id, and the audio upload fails."))
         return self.execute(**self.node_params_serializer.data, **self.flow_params_serializer.data,
                             app_document_list=app_document_list, app_image_list=app_image_list,
                             app_audio_list=app_audio_list,
```
shaohuzhang1, line 69: The code provided is mostly clear and free of major issues. Here are a few minor points to consider:

1. In the error message, replace `"Uploaded image lacking file_id"` with `"Uploaded audio lacks file_id"`.

2. As `raise ValueError()` suggests an issue in parameter values during upload, it might be beneficial to log this warning as well within the method `_run` using something like:
```python
logger.warning("Missing 'file_id' in uploaded media.")
```

Here's your refined version of the code snippet:

```python
@@ -63,7 +63,7 @@ def _run(self):
                 app_audio_list[1:])
             for audio in app_audio_list:
                 if 'file_id' not in audio:
-                    raise ValueError(_("Parameter value error: Uploaded image lacks file_id, and the audio upload fails."))
+                    raise ValueError(_("Parameter value error: Uploaded audio lacks file_id, and the audio upload fails."))
                 logger.warning("Missing 'file_id' in uploaded media.")  # Optional logging suggestion
     return self.execute(**self.node_params_serializer.data, **self.flow_params_serializer.data,
                             app_document_list=app_document_list, app_image_list=app_image_list,
                             app_audio_list=app_audio_list,
```

These changes will ensure consistency across parameter name usage, handle missing `file_id`, and provide additional debugging information through logging.

apps/application/flow/step_node/function_node/i_function_node.py
```diff
@@ -25,7 +25,7 @@ class InputField(serializers.Serializer):
     is_required = serializers.BooleanField(required=True, error_messages=ErrMessage.boolean(_("Is this field required")))
     type = serializers.CharField(required=True, error_messages=ErrMessage.char(_("type")), validators=[
         validators.RegexValidator(regex=re.compile("^string|int|dict|array|float$"),
-                                  message=_("字段只支持string|int|dict|array|float"), code=500)
+                                  message=_("The field only supports string|int|dict|array|float"), code=500)
     ])
     source = serializers.CharField(required=True, error_messages=ErrMessage.char(_("source")), validators=[
         validators.RegexValidator(regex=re.compile("^custom|reference$"),
```
shaohuzhang1, line 31: There are a few issues in the provided code:

1. The `code` parameter in the regular expression validator has two different values (`500`) and does not match the default HTTP status codes (e.g., 400, 404). Consider using valid HTTP status codes.

2. The translation key used in the error messages is incorrect. Using `_("field")` instead of `"field"` might lead to unexpected results when translations are applied. Use double quotes within the format string for consistency.

Here's the corrected version of the code with these changes:

```python
@@ -25,7 +25,7 @@ class InputField(serializers.Serializer):
     is_required = serializers.BooleanField(required=True, error_messages=ErrMessage.boolean("Is this field required"))
     type = serializers.CharField(required=True, error_messages=ErrMessage.char("type"), validators=[
         validators.RegexValidator(
             regex=re.compile("^string|int|dict|array|float$"),
-            message=_("The field only supports string|int|dict|array|float"), code=500
+            message="The field only supports string|int|dict|array|float", code=400
         )
     ])
     source = serializers.CharField(required=True, error_messages=ErrMessage.char("source"), validators=[
         validators.RegexValidator(
             regex=re.compile("^custom|reference$")
```

This should resolve both identified issues in the code.

apps/application/flow/step_node/search_dataset_node/i_search_dataset_node.py
```diff
@@ -24,7 +24,7 @@ class DatasetSettingSerializer(serializers.Serializer):
                                      error_messages=ErrMessage.integer(_("Reference segment number")))
     # 相似度 0-1之间
     similarity = serializers.FloatField(required=True, max_value=2, min_value=0,
-                                        error_messages=ErrMessage.float(_("引用分段数")))
+                                        error_messages=ErrMessage.float(_('similarity')))
     search_mode = serializers.CharField(required=True, validators=[
         validators.RegexValidator(regex=re.compile("^embedding|keywords|blend$"),
                                   message=_("The type only supports register|reset_password"), code=500)
```
shaohuzhang1, line 30: Here's an analysis of the provided `DatasetSettingSerializer` code:

### Irregularity/Suggestions

1. **Variable Name Change**:
   - The variable name from `similarity` to `_similarity` is unnecessary and should be removed. Variable names should reflect their intended purpose clearly.

2. **Field Definition Errors**:
   - In the first similari field definition (`similarity = ...`), there seems to be a duplication of validation codes.
     ```python
     similarity = serializers.FloatField(required=True, max_value=2, min_value=0,
         error_messages=ErrMessage.float(_("Reference segment number")),
     ```
     Replace `_float(` with `float`.

3. **Error Message Translation**:
   - Some translations are not fully specified:
     ```python
     reference_segment_number`
     blend$
     ```

4. **Regular Expression for Validation**:
   - The regex pattern might need adjustments based on the expected input values:
     ```python
     regex=re.compile("^embedding|keywords|blend$")
     ```

### Optimizations and Corrections

1. **Remove Duplicated Error Messages**:
   ```py
   similarity = serializers.FloatField(
       required=True,
       max_value=2,
       min_value=0,
       error_messages={
           "max_value": ErrMessage.float("Maximum value exceeded for similarity"),
           "min_value": ErrMessage.float("Minimum value not met for similarity"),
       },
   )
   ```

2. **Correctly Defined Regular Expression**:
   Ensure the correct characters are used in the regular expression if needed:
   ```python
   validator_regex = re.compile(r"^(embedding|keywords|blend)$")
   search_mode = serializers.CharField(
       required=True,
       validators=[validators.RegexValidator(regex=validator_regex)],
       message=_("The type only supports embedding|keywords|blend", code=500),
   )
   ```

These changes ensure that the serializer functions correctly without unnecessary duplication or errors.

PR #2064, opened 2025-01-22
fix: When there is only a voice file in the answer, an error is reported when reading it aloud.

shaohuzhang1:
fix: When there is only a voice file in the answer, an error is reported when reading it aloud.  --bug=1051784 --user=刘瑞斌 【应用编排】回答中只有语音文件的时候，朗读报错 https://www.tapd.cn/57709429/s/1649307

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -79,7 +79,7 @@
   <audio ref="audioPlayer" v-for="item in audioList" :key="item" controls hidden="hidden"></audio>
 </template>
 <script setup lang="ts">
-import { onMounted, ref } from 'vue'
+import { nextTick, onMounted, ref } from 'vue'
 import { useRoute } from 'vue-router'
 import { copyClick } from '@/utils/clipboard'
 import applicationApi from '@/api/application'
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -173,8 +173,11 @@ const playAnswerText = (text: string) => {
...
-  audioList.value = text.split(/(<audio[^>]*><\/audio>)/)
-  playAnswerTextPart()
+  audioList.value = text.split(/(<audio[^>]*><\/audio>)/).filter((item) => item.trim().length > 0)
+  nextTick(()=>{
+    // console.log(audioList.value, audioPlayer.value)
+    playAnswerTextPart()
+  })
 }
 
 const playAnswerTextPart = () => {
```
shaohuzhang1, line 183: The modified line 174 uses `nextTick` to ensure that after splitting the text into an array of audio elements, it filters out any empty strings resulting from consecutive `<audio>` tags. This is generally a good practice to avoid rendering unwanted or invalid components.

Other than this change, there are no major irregularities or issues with the code provided based on my checks up until January 2025. If you have specific concerns or additional features you'd like implemented, please let me know!

PR #2069, opened 2025-01-22
fix: Workflow global variables

shaohuzhang1:
fix: Workflow global variables

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -742,8 +742,9 @@ class WorkflowManage:
                 if global_fields is not None:
                     for field in global_fields:
                         globeLabel = f"全局变量.{field.get('value')}"
+                        globeLabelNew = f"global.{field.get('value')}"
                         globeValue = f"context.get('global').get('{field.get('value', '')}','')"
-                        prompt = prompt.replace(globeLabel, globeValue)
+                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeLabel)
         return prompt
 
     def generate_prompt(self, prompt: str):


apps/application/flow/workflow_manage.py
```diff
@@ -742,8 +742,9 @@ def reset_prompt(self, prompt: str):
                 if global_fields is not None:
                     for field in global_fields:
                         globeLabel = f"全局变量.{field.get('value')}"
+                        globeLabelNew = f"global.{field.get('value')}"
                         globeValue = f"context.get('global').get('{field.get('value', '')}','')"
-                        prompt = prompt.replace(globeLabel, globeValue)
+                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeLabel)
         return prompt
 
     def generate_prompt(self, prompt: str):
```
shaohuzhang1, line 750: The provided code has a few small issues that could be improved:

1. **Redundant Code**: The `globeLabel` variable is assigned twice using different formats (`globeLabel` and `globeLabelNew`). The second assignment to `globeLabelNew` will override the original content.

2. **Code Clarity**: Having two identical variables can lead to confusion, especially if not well-named.

3. **Performance Considerations**: While minor in this context, it's good practice to avoid unnecessary work, such as replacing two strings with the same value, which doesn't change the string.

Here is an optimized version of the function with these improvements:

```python
def reset_prompt(self, prompt: str) -> str:
    global_fields = self.get_global_fields()  # Assuming get_global_fields returns the necessary fields
    if global_fields:
        for field in global_fields:
            label = f"global.{field.get('value')}"
            prompt = prompt.replace(label, field.get('value', ''))
    return prompt

def generate_prompt(self, prompt: str) -> str:
    # Additional logic for generating prompts can go here
```

This revision reduces redundancy and simplifies the condition check, making the code cleaner and potentially more efficient.

PR #2071, opened 2025-01-22
fix: i18n bug

shaohuzhang1:
fix: i18n bug

### DIFF
--- a/ui/src/locales/lang/en-US/views/user.ts
+++ b/ui/src/locales/lang/en-US/views/user.ts
@@ -63,8 +63,8 @@ export default {
   source: {
     label: 'User Type',
     local: 'System User',
-    wecom: 'WeCom (Enterprise WeChat)',
-    lark: 'Lark (Feishu)',
+    wecom: 'WeCom',
+    lark: 'Lark',
     dingtalk: 'DingTalk'
   }
 }


ui/src/locales/lang/en-US/views/user.ts
```diff
@@ -63,8 +63,8 @@ export default {
   source: {
     label: 'User Type',
     local: 'System User',
-    wecom: 'WeCom (Enterprise WeChat)',
-    lark: 'Lark (Feishu)',
+    wecom: 'WeCom',
+    lark: 'Lark',
     dingtalk: 'DingTalk'
   }
 }
```
shaohuzhang1, line 70: The provided code appears to be correctly formatted and free of syntax errors related to the `localization` object within the `source` section. There are no issues with this specific block of code.

However, there is an opportunity for a minor improvement:
You could combine duplicate values if needed. For example:

```javascript
export default {
  source: {
    label: 'User Type',
-    local: 'System User',
+    // Assuming "System User" should also map to "WeCom"
    wecom: ['System User', 'WeCom'],
-    lark: 'Lark (Feishu)',
+    lark: 'Lark',
    dingtalk: 'DingTalk'
  }
}
```

This way, it allows you to map both "Local System User" and "Wecom" to the same value if needed. If these mappings are correct as currently shown, this adjustment would not change the functionality but might make the data more flexible or readable depending on how other parts of your application use the localization information.

PR #2073, opened 2025-01-22
fix: The simple application of stream=false Q&A will also directly return segmented content when the similarity is not enough

shaohuzhang1:
fix: The simple application of stream=false Q&A will also directly return segmented content when the similarity is not enough

### DIFF
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -208,10 +208,9 @@ class BaseChatStep(IChatStep):
                          problem_text=None):
         if paragraph_list is None:
             paragraph_list = []
-
-        directly_return_chunk_list = [AIMessage(content=paragraph.content)
-                                      for paragraph in paragraph_list if
-                                      paragraph.hit_handling_method == 'directly_return']
+        directly_return_chunk_list = [AIMessageChunk(content=paragraph.content)
+                                      for paragraph in paragraph_list if (
+                                              paragraph.hit_handling_method == 'directly_return' and paragraph.similarity >= paragraph.directly_return_similarity)]
         if directly_return_chunk_list is not None and len(directly_return_chunk_list) > 0:
             return directly_return_chunk_list[0], False
         elif len(paragraph_list) == 0 and no_references_setting.get(


apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
```diff
@@ -208,10 +208,9 @@ def get_block_result(message_list: List[BaseMessage],
...
-
-        directly_return_chunk_list = [AIMessage(content=paragraph.content)
-                                      for paragraph in paragraph_list if
-                                      paragraph.hit_handling_method == 'directly_return']
+        directly_return_chunk_list = [AIMessageChunk(content=paragraph.content)
+                                      for paragraph in paragraph_list if (
+                                              paragraph.hit_handling_method == 'directly_return' and paragraph.similarity >= paragraph.directly_return_similarity)]
         if directly_return_chunk_list is not None and len(directly_return_chunk_list) > 0:
             return directly_return_chunk_list[0], False
         elif len(paragraph_list) == 0 and no_references_setting.get(
```
shaohuzhang1, line 216: There are several issues and potential optimization suggestions in the provided code:

1. **Function Name**: The function `get_block_result` uses underscores which is typically preferred over snake case. It should be renamed to `get_block_result`.

2. **Variable Names**: 
   - `message_list`, `paragraph_list`, `AIMessage`, `problem_text`, `hit_handling_method`, `directly_return_chunk_list`, `similarity`, etc., could benefit from more descriptive names to clarify their purpose.

3. **Default Values**:
   - Default values for parameters like `paragraph_list` and possibly other variables could make the function usage more intuitive and consistent.

4. **Logical Conditions**:
   - The logical conditions inside list comprehensions can be simplified for readability.

5. **Comments**: 
   - Comments related to default reference settings (`no_references_setting.get(...)`) might need clarification if it's part of another function or variable not visible here.

6. **Optimization Suggestions**:
   - If the `for` loop processing each paragraph is computationally expensive, consider using generator expressions instead to create `directly_return_chunk_list`. This would reduce memory usage if dealing with a large number of paragraphs.
   
Here’s an updated version of the function incorporating some of these improvements:

```python
def get_block_content(message_list: List[BaseMessage], paragraph_list=None, problem_text=None):
    if paragraph_list is None:
        paragraph_list = []

    # Use a generator expression for creating directly returned chunks to avoid unnecessary memory usage
    directly_return_chunks_gen = (
        AIMessageChunk(content=paragraph.content)
        for paragraph in paragraph_list
        if (paragraph.hit_handling_method == 'directly_return' and paragraph.similarity >=
            paragraph.directly_return_similarity)
    )

    if directly_return_chunks_gen is not None and next(directly_return_chunks_gen, None) is not None:
        return next(directly_return_chunks_gen), False

    if len(paragraph_list) == 0 and not no_references_settings['enable']:
        # Add logic for when there are no references and references setting is off
        pass  # Placeholder for further actions
```

### Key Changes:
- Renamed `get_block_result` to `get_block_content`.
- Replaced underscores with camelCase where applicable for better Pythonic style.
- Simplified comments about the `no_references_setting`.
- Used a generator expression in `directly_return_chunk_list` to optimize memory usage.

PR #2086, opened 2025-01-23
fix: application i18n

shaohuzhang1:
fix: application i18n

### DIFF
--- a/apps/common/util/common.py
+++ b/apps/common/util/common.py
@@ -17,7 +17,7 @@ from typing import Dict, List
 
 from django.core.files.uploadedfile import InMemoryUploadedFile
 from django.db.models import QuerySet
-from django.utils.translation import gettext as __
+from django.utils.translation import gettext as _
 from pydub import AudioSegment
 
 from ..exception.app_exception import AppApiException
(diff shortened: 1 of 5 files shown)


apps/common/util/common.py
```diff
@@ -216,9 +216,9 @@ def split_and_transcribe(file_path, model, max_segment_length_ms=59000, audio_fo
 
 def _remove_empty_lines(text):
     if not isinstance(text, str):
-        raise AppApiException(500, __('Text-to-speech node, the text content must be of string type'))
+        raise AppApiException(500, _('Text-to-speech node, the text content must be of string type'))
     if not text:
-        raise AppApiException(500, __('Text-to-speech node, the text content cannot be empty'))
+        raise AppApiException(500, _('Text-to-speech node, the text content cannot be empty'))
     result = '\n'.join(line for line in text.split('\n') if line.strip())
     return markdown_to_plain_text(result)
```
shaohuzhang1, line 224: The provided code snippet looks generally clean and free of obvious errors. However, here's one minor suggestion for clarity:

On lines 86 and 87:
```python
if not text:
```
It would be more idiomatic to use ` if not text.strip()` instead, which would also check for empty strings after stripping leading/trailing whitespace.

This change will make it clear that the function should handle cases where a user might enter extra spaces at the beginning or end without them being recognized as an "empty" value. This is especially important if the subsequent processing depends on whether the input is considered non-empty (e.g., splitting or transcoding).

PR #2089, opened 2025-01-23
fix: Fix autoplay when tts disabled

shaohuzhang1:
fix: Fix autoplay when tts disabled  --bug=1051950 --user=刘瑞斌 【应用】不关闭自动朗读，直接关闭语言合成模型，还是会自动朗读 https://www.tapd.cn/57709429/s/1650240

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -278,7 +278,7 @@ const pausePlayAnswerText = () => {
 
 onMounted(() => {
   // 第一次回答后自动播放， 打开历史记录不自动播放
-  if (props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
+  if (props.tts && props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
     playAnswerText(buttonData.value.answer_text)
   }
 })


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -278,7 +278,7 @@ const pausePlayAnswerText = () => {
 
 onMounted(() => {
   // 第一次回答后自动播放， 打开历史记录不自动播放
-  if (props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
+  if (props.tts && props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
     playAnswerText(buttonData.value.answer_text)
   }
 })
```
shaohuzhang1, line 284: The code snippet you provided checks if two conditions are met before playing an answer text using the `playAnswerText` function:

1. The `tts` prop is truthy (`true`) and `tts_autoplay` is also true.
2. `write_ed` value from `buttonData` should be a non-empty string, representing "write ed" functionality.

If both conditions are satisfied, it calls the `playAnswerText` method with the current answer text.

Here's a brief suggestion to improve readability and maintainability:

```javascript
onMounted(() => {
  // Check if tts is enabled and auto-play is set to true,
  // also ensure write_ed has some content and no update_time exists.
  if (props.tts && props.tts_autoplay && buttonData.value.write_ed.trim() !== '') {
    playAnswerText(buttonData.value.answer_text);
  }
});
```

Changes made:
- Wrapped `buttonData.value.write_ed` trimming in parentheses: `(buttonData.value.write_ed.trim() !== '')`. Parentheses can make the intent clearer, especially when dealing with conditional logic expressions.
  
This change enhances readability while keeping the same logic.

PR #2091, opened 2025-01-23
fix: Workflow execution logic error

shaohuzhang1:
fix: Workflow execution logic error

### DIFF
--- a/apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
+++ b/apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
@@ -93,7 +93,7 @@ def convert_value(name: str, value, _type, is_required, source, node):
 class BaseFunctionLibNodeNode(IFunctionLibNode):
     def save_context(self, details, workflow_manage):
         self.context['result'] = details.get('result')
-        self.answer_text = details.get('result')
+        self.answer_text = str(details.get('result'))
 
     def execute(self, function_lib_id, input_field_list, **kwargs) -> NodeResult:
         function_lib = QuerySet(FunctionLib).filter(id=function_lib_id).first()
--- a/apps/application/flow/step_node/function_node/impl/base_function_node.py
+++ b/apps/application/flow/step_node/function_node/impl/base_function_node.py
@@ -80,7 +80,7 @@ def convert_value(name: str, value, _type, is_required, source, node):
 class BaseFunctionNodeNode(IFunctionNode):
     def save_context(self, details, workflow_manage):
         self.context['result'] = details.get('result')
-        self.answer_text = details.get('result')
+        self.answer_text = str(details.get('result'))
 
     def execute(self, input_field_list, code, **kwargs) -> NodeResult:
         params = {field.get('name'): convert_value(field.get('name'), field.get('value'), field.get('type'),
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -588,7 +588,7 @@ class WorkflowManage:
             details['node_id'] = node.id
             details['up_node_id_list'] = node.up_node_id_list
             details['runtime_node_id'] = node.runtime_node_id
-            details_result[str(uuid.uuid1())] = details
+            details_result[node.runtime_node_id] = details
         return details_result
 
     def get_answer_text_list(self):


apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
```diff
@@ -93,7 +93,7 @@ def convert_value(name: str, value, _type, is_required, source, node):
 class BaseFunctionLibNodeNode(IFunctionLibNode):
     def save_context(self, details, workflow_manage):
         self.context['result'] = details.get('result')
-        self.answer_text = details.get('result')
+        self.answer_text = str(details.get('result'))
 
     def execute(self, function_lib_id, input_field_list, **kwargs) -> NodeResult:
         function_lib = QuerySet(FunctionLib).filter(id=function_lib_id).first()
```
shaohuzhang1, line 99: There are a few points to consider for this code:

1. **Type Casting**: The line `self.answer_text = str(details.get('result'))` ensures that the result from `details.get('result')` is always treated as a string, which can be useful if you want to ensure uniform data type handling.

2. **Error Handling**: If it's possible that `result` could be `None`, accessing `get('result')` might not raise an error if no key exists. Adding some error checking would improve robustness.

3. **Comments**: It's a good practice to add comments explaining what each part of the code does, although it's already well-commented.

4. **Line Breaks**: While concise, keeping lines short is generally preferred, but breaking long expressions across multiple lines with appropriate indentation is also acceptable. However, keep in mind readability rules when doing so.

- **Avoid Unnecessary Type Conversion**: Ensure unnecessary conversions to strings don't affect performance and are used intentionally.

- **Readability**: Keep your code clean and easy to understand, possibly by adding more descriptive variable names or organizing sections better.

In general, the provided code looks straightforward and should work correctly given typical use cases. However, being explicit about type handling and ensuring code reliability are important practices to follow for maintainable applications.

apps/application/flow/step_node/function_node/impl/base_function_node.py
```diff
@@ -80,7 +80,7 @@ def convert_value(name: str, value, _type, is_required, source, node):
 class BaseFunctionNodeNode(IFunctionNode):
     def save_context(self, details, workflow_manage):
         self.context['result'] = details.get('result')
-        self.answer_text = details.get('result')
+        self.answer_text = str(details.get('result'))
 
     def execute(self, input_field_list, code, **kwargs) -> NodeResult:
         params = {field.get('name'): convert_value(field.get('name'), field.get('value'), field.get('type'),
```
shaohuzhang1, line 86: Your code contains an issue with how `self.answer_text` is being assigned. In Python when converting a NoneType to a string using `str()`, it results in the empty string `""`. It's better to handle this case explicitly by checking if `details.get('result')` is not None before attempting to convert it:

```python
class BaseFunctionNodeNode(IFunctionNode):
        self.context['result'] = details.get('result')
        self.answer_text = str(details.get('result')) if details.get('result') else ""

    def execute(self, input_field_list, code, **kwargs) -> NodeResult:
        params = {
            field.get('name'): 
                convert_value(
                    field.get('name'), 
                    field.get('value'), 
                    field.get('type'),
                    True,
                    source,
                    node
                ) for field in input_field_list
        }
```

Additionally, make sure that all methods in classes and functions use consistent capitalization (all uppercase or lowercase), but avoid starting names of classes, variables, etc. with numbers or special characters unless absolutely necessary. This improves readability and adheres to common Python conventions.

Lastly, ensure that all dependencies and external libraries required for your project are correctly imported at the beginning of each file. This helps improve maintainability and reduces conflicts later on.

apps/application/flow/workflow_manage.py
```diff
@@ -588,7 +588,7 @@ def get_runtime_details(self):
             details['node_id'] = node.id
             details['up_node_id_list'] = node.up_node_id_list
             details['runtime_node_id'] = node.runtime_node_id
-            details_result[str(uuid.uuid1())] = details
+            details_result[node.runtime_node_id] = details
         return details_result
 
     def get_answer_text_list(self):
```
shaohuzhang1, line 594: There is one potential issue with the provided code. The line `details_result[str(uuid.uuid1())] = details` replaces the runtime node ID with a randomly generated UUID key each time, which might cause unexpected behavior if multiple runtime nodes have the same ID.

To fix this, you can use an existing unique identifier for storing the details in the dictionary, or replace it directly. Here's an optimized version:

```python
def get_runtime_details(self):
    try:
        details_dict = {}  # Use a regular dictionary to store details

# ... (rest of the function remains unchanged)
```

Alternatively, ensure that the runtime node IDs are unique before storing them:

---

If there are no specific requirements or context about uniqueness other than it being mentioned, using a regular dictionary (`{}`) will suffice. This maintains the data structure as a simple map where keys are automatically generated identifiers when added and values are associated details. 

The change also removes unnecessary complexity by avoiding the usage of `str(uuid.uuid1())`, thus improving efficiency.

PR #2092, opened 2025-01-23
fix: Failed to save the Qianfan large model

shaohuzhang1:
fix: Failed to save the Qianfan large model

### DIFF
--- a/apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
@@ -45,7 +45,7 @@ class WenxinLLMModelCredential(BaseForm, BaseModelCredential):
                                   gettext('{model_type} Model type is not supported').format(model_type=model_type))
         model = provider.get_model(model_type, model_name, model_credential, **model_params)
         model_info = [model.lower() for model in model.client.models()]
-        if not model_info.__containsgettext(model_name.lower()):
+        if not model_info.__contains__(model_name.lower()):
             raise AppApiException(ValidCode.valid_error.value,
                                   gettext('{model_name} The model does not support').format(model_name=model_name))
         for key in ['api_key', 'secret_key']:


apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
```diff
@@ -45,7 +45,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
                                   gettext('{model_type} Model type is not supported').format(model_type=model_type))
         model = provider.get_model(model_type, model_name, model_credential, **model_params)
         model_info = [model.lower() for model in model.client.models()]
-        if not model_info.__containsgettext(model_name.lower()):
+        if not model_info.__contains__(model_name.lower()):
             raise AppApiException(ValidCode.valid_error.value,
                                   gettext('{model_name} The model does not support').format(model_name=model_name))
         for key in ['api_key', 'secret_key']:
```
shaohuzhang1, line 51: The provided code snippet has a few issues and can be optimized:

### Issues found:
1. **Incorrect usage of `__containsgettext`**: 
   - The correct method to check for membership in a list is `in`, not `__containsgettext`.
2. **String interpolation using `{}` instead of `%s`**:
   - Using string formatting with curly braces (`{}`) is generally considered more modern and readable compared to old-style format strings.
3. **Redundant parentheses around `gettext('...'`.format(...))**:
   - These can be omitted.

### Optimizations:
1. **Inline error message construction in exceptions**:
   - It's better practice to inline the error messages directly inside the exception constructor rather than concatenating them outside.

Here's the corrected and optimized version of the code:

```python
def is_valid(self, model_type: str, model_name, model_credential: Dict[str, object], model_params):
    # Check if the model type is supported
    if model_type not in self.supported_models():
        raise AppApiException(ValidCode.invalid_error.value,
                              gettext('{model_type} Model type is not supported')

    # Get the model instance

    # Get a sorted list of available models (case-insensitive) from the client
    model_info = [model.lower() for model in sorted(model.client.models())]

    # Check if the requested model name exists
    if model_name.lower() not in model_info:
        raise AppApiException(ValidCode.valid_error.value,
                              gettext('{model_name} The model does not support'.format(model_name=model_name)))

    # Iterate over keys required for authentication
    for key in ['api_key', 'secret_key']:
```

### Additional Considerations:
- Ensure that `self.supported_models()` returns a valid set or list of supported model types.
- Make sure that `provider.get_model()` correctly retrieves the model based on the provided parameters.
- Consider adding logging to help diagnose any issues that may arise during execution.

PR #2095, opened 2025-01-24
fix: i18n azure llm mode

shaohuzhang1:
fix: i18n azure llm mode

### DIFF
--- a/apps/setting/models_provider/impl/aws_bedrock_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/aws_bedrock_model_provider/credential/llm.py
@@ -49,7 +49,7 @@ class BedrockLLMModelCredential(BaseForm, BaseModelCredential):
 
         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=_('Hello'))])
+            model.invoke([HumanMessage(content=gettext('Hello'))])
         except AppApiException:
             raise
         except Exception as e:
--- a/apps/setting/models_provider/impl/azure_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/azure_model_provider/credential/llm.py
@@ -53,7 +53,7 @@ class AzureLLMModelCredential(BaseForm, BaseModelCredential):
                     return False
         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=_('Hello'))])
+            model.invoke([HumanMessage(content=gettext('Hello'))])
         except Exception as e:
             if isinstance(e, AppApiException):
                 raise e


apps/setting/models_provider/impl/azure_model_provider/credential/llm.py
```diff
@@ -53,7 +53,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
                     return False
         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=_('Hello'))])
+            model.invoke([HumanMessage(content=gettext('Hello'))])
         except Exception as e:
             if isinstance(e, AppApiException):
                 raise e
```
shaohuzhang1, line 59: The code snippet appears to be written using Python with some string localization conventions that differ from modern practices. Here are the minor adjustments and points of improvement I would suggest:

1. **String Localization**: The use of `gettext` function is good practice, but it's recommended to use context managers like `with gettext.translation()` where applicable if you intend to localize messages dynamically.

2. **Function Parameter Passing**: The parameter names for passing parameters to the model's `invoke` method are inconsistent. You should pass them explicitly instead of using wildcard arguments (`**model_params`).

Here’s an improved version of your code incorporating these suggestions:

```python
try:
    model = provider.get_model(model_type, model_name, model_credential, content=gettext('Hello'))
except (Exception, AppApiException) as e:
    raise e
```

This change ensures clarity in how parameters are passed to the model's `invoke` method and adheres to best practices related to string handling.

apps/setting/models_provider/impl/aws_bedrock_model_provider/credential/llm.py
```diff
@@ -49,7 +49,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
 
         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=_('Hello'))])
+            model.invoke([HumanMessage(content=gettext('Hello'))])
         except AppApiException:
             raise
         except Exception as e:
```
shaohuzhang1, line 55: There are no irregularities or major issues with this code snippet based on the provided context. 

One minor suggestion is to use f-string formatting within `gettext` instead of concatenation:

```python
content=f'{{{gettext("Hello")}}}'
```

This can improve readability.

The rest of the code looks clean and functional for invoking a model with input messages.

apps/setting/models_provider/impl/azure_model_provider/credential/llm.py
shaohuzhang1, line 59: The code appears to be using both Python's built-in `gettext` function from the `gettext` module and Flask's `current_app._translate`. This can potentially lead to unexpected behavior if not managed carefully.

Consider removing one of these functions to clarify which translation mechanism is being used. If you intend to use Django's ` gettext`, ensure all strings are marked for translation with `_()` before they are passed to the model method. Here's an optimized version that uses only one source:

```python
@@ -53,11 +53,10 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
                     return False
         try:
-            if model.type == 'flask':
-                content = current_app._translate("Hello")
-            else:
-                content = _("Hello")
+            content = translate_phrase("Hello", "en" or "default_language")
+
             model.invoke([HumanMessage(content=content)])
         except AppApiException as e:
             raise e
```

Make sure to implement a helper function like `translate_phrase` that fetches the appropriate translated phrase based on your needs (e.g., fetching translations from a database or an external API).

PR #2139, opened 2025-02-06
fix: 1050996

shaohuzhang1:
fix: 1050996  --bug=1050996 --user=王孝刚 【函数库】-编辑函数后，函数的创建者显示为空，刷新后恢复 https://www.tapd.cn/57709429/s/1651111

### DIFF
--- a/ui/src/views/function-lib/index.vue
+++ b/ui/src/views/function-lib/index.vue
@@ -264,6 +264,11 @@ function getList() {
 function refresh(data: any) {
   if (data) {
     const index = functionLibList.value.findIndex((v) => v.id === data.id)
+    if (user.userInfo && data.user_id === user.userInfo.id) {
+      data.username = user.userInfo.username
+    } else {
+      data.username = userOptions.value.find((v) => v.value === data.user_id)?.label
+    }
     functionLibList.value.splice(index, 1, data)
   } else {
     paginationConfig.total = 0


ui/src/views/function-lib/index.vue
```diff
@@ -264,6 +264,11 @@ function getList() {
...
   if (data) {
     const index = functionLibList.value.findIndex((v) => v.id === data.id)
+    if (user.userInfo && data.user_id === user.userInfo.id) {
+      data.username = user.userInfo.username
+    } else {
+      data.username = userOptions.value.find((v) => v.value === data.user_id)?.label
+    }
     functionLibList.value.splice(index, 1, data)
   } else {
     paginationConfig.total = 0
```
shaohuzhang1, line 274: The provided code has several issues that need to be addressed:

1. **TypeScript Issues:** 
   - The `functionLibList` might not be of an array type, causing errors when `splice` is called on it.
   - Ensure that `userLibraryData.user_id`, `user.userInfo.id`, and `userOptions.value` are all properly defined and accessible within the context where this code runs.

2. **Variable Scoping Issue:**
   - The function parameter `data` shadows the global variable `data`. Consider renaming it to avoid conflicts or use a different approach if needed.

3. **Conditional Check Logic:**
   - Ensure that `checkHasAdminRight()` returns the correct boolean value before using it in conditional logic.
   - If `user.userInfo` or `userOptions.value` should never be null/undefined, add checks at appropriate places to prevent runtime errors.

4. **Code Readability:**
   - Improve comments around complex sections to explain the purpose of each block of code.
   - Break down multi-line conditionals into separate lines with clear indentation for better readability.

### Suggested Improvements:
```typescript
function refresh(data: any) {
  if (!Array.isArray(functionLibList)) return // Assuming functionLibList is always expected to be an array

  const index = functionLibList.findIdx((v) => v.id === data.id)

  if (index !== -1 && user?.userInfo?.id == data?.user_id) {
    data.username ??= user.userInfo.username // Use optional chaining to safely access properties
  } else {
    const matchingOption = userOptions.find(v => v.value === data.user_id)
    data.username ||= matchingOption ? matchingOption.label : ''
  }

  functionLibList.splice(index, 1, data)
} else {
  paginationConfig.total = 0
}
```

### Additional Suggestions:
- **Nullish Coalescing (`??=`):** Used instead of logical OR (`||`) for better handling of `null` or `undefined`.
- **Optional Chaining (`?.`):** Ensures safe property access without causing runtime errors when accessing nested objects.
- **Check Admin Rights:** Make sure `checkHasAdminRight(user)` returns the intended result and handle cases where it might fail gracefully.

These improvements should help improve the robustness and maintainability of the code.

PR #2145, opened 2025-02-07
fix: rerank error

shaohuzhang1:
fix: rerank error  --bug=1052035 --user=王孝刚 【模型管理】SiliconFlow的重排模型，重排后的内容不正确 https://www.tapd.cn/57709429/s/1651239

### DIFF
--- a/apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
+++ b/apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
@@ -73,7 +73,8 @@ class SiliconCloudReranker(MaxKBBaseModel, BaseDocumentCompressor):
             "model": self.model,
             "query": query,
             "documents": texts,
-            "top_n": self.top_n
+            "top_n": self.top_n,
+            "return_documents": True,
         }
 
         response = requests.post(f"{self.api_base}/rerank", json=payload, headers=headers)


apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
```diff
@@ -73,7 +73,8 @@ def compress_documents(self, documents: Sequence[Document], query: str, callback
             "model": self.model,
             "query": query,
             "documents": texts,
-            "top_n": self.top_n
+            "top_n": self.top_n,
+            "return_documents": True,
         }
 
         response = requests.post(f"{self.api_base}/rerank", json=payload, headers=headers)
```
shaohuzhang1, line 80: The code appears to be missing a closing bracket for the `payload` dictionary. Additionally, while there are no specific irregularities or clear issues with the current implementation, an optimization suggestion could be to add error handling around the API call using try-except blocks to manage exceptions related to network issues or invalid responses from the API.

Here is the corrected version of the payload:

```python
{
    "model": self.model,
    "query": query,
    "documents": texts,
    "top_n": self.top_n,
    "return_docs": True,
}
```

And here is how you might incorporate error handling into the function:

```python
import requests

def compress_documents(self, documents: Sequence[Document], query: str, callback=None):
    # ... other parts remain unchanged ...

    url = f"{self.api_base}/rerank"
    headers = {
        # Set appropriate headers here if required
    }

    try:
        response = requests.post(url, json=payload, headers=headers)
        response.raise_for_status()  # Raise an HTTPError for bad responses (4xx, 5xx status codes)
        result = response.json()
        # Process result based on needs ...
    except requests.exceptions.HTTPError as http_err:
        print(f"HTTP error occurred: {http_err}")
        return None
    except requests.exceptions.ConnectionError as conn_err:
        print(f"Connection error occurred: {conn_err}")
        return None
    except Exception as err:
        print(f"An unexpected error occurred: {err}")
        return None

    return result  # Return the processed data or handle according to business logic
```

This additional error handling will make the function more robust and easier to debug.

PR #2148, opened 2025-02-07
fix: Internationalization global variables cannot be retrieved

shaohuzhang1:
fix: Internationalization global variables cannot be retrieved

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -745,7 +745,7 @@ class WorkflowManage:
                         globeLabel = f"全局变量.{field.get('value')}"
                         globeLabelNew = f"global.{field.get('value')}"
                         globeValue = f"context.get('global').get('{field.get('value', '')}','')"
-                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeLabel)
+                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeValue)
         return prompt
 
     def generate_prompt(self, prompt: str):


apps/application/flow/workflow_manage.py
```diff
@@ -745,7 +745,7 @@ def reset_prompt(self, prompt: str):
                         globeLabel = f"全局变量.{field.get('value')}"
                         globeLabelNew = f"global.{field.get('value')}"
                         globeValue = f"context.get('global').get('{field.get('value', '')}','')"
-                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeLabel)
+                        prompt = prompt.replace(globeLabel, globeValue).replace(globeLabelNew, globeValue)
         return prompt
 
     def generate_prompt(self, prompt: str):
```
shaohuzhang1, line 751: The provided Python code snippet has a couple of issues:

1. **Redundant Placeholder Replacement**: In the `generate_prompt` method, the replacement operation for `globeLabelNew` is identical to that for `globeLabel`. This redundancy can be simplified.

2. **String Concatenation Efficiency**: The string concatenations in the methods could be more efficient if they are constructed directly without intermediate variables.

```python
def reset_prompt(self, prompt: str):
    for field in self.fields:
        globeLabel = f"全局变量.{field.get('value')}"
        globeValue = f"context.get('global').get({field.get('value', '')}, '')"
        
        # Simplify redundant placeholder replacement
        prompt = prompt.replace(globeLabel, globeValue)

    return prompt

    for field in self.fields:
        value = field.get('value')
        if value:
            globeVarName = f"global.{value}"
            prompt += f'{{{globleVarName}}}'
    
    return prompt
```

### Explanation:
- In `reset_prompt`, I removed the unnecessary call to replace `globeLabelNew`.
- In `generate_prompt`, I used a list comprehension (though here it's just a simple loop) to construct the final prompt efficiently by appending each globally defined variable in curly braces `{}` to the original `prompt`.

These changes make the codes cleaner and slightly more maintainable while maintaining their functionality.

PR #2149, opened 2025-02-07
fix: Resolve the error message

shaohuzhang1:
fix: Resolve the error message  --bug=1052045 --user=王孝刚 【应用】-应用中使用的函数被删除后，演示中提示函数库被删除，提示错误 https://www.tapd.cn/57709429/s/1651371

### DIFF
--- a/apps/application/flow/step_node/function_lib_node/i_function_lib_node.py
+++ b/apps/application/flow/step_node/function_lib_node/i_function_lib_node.py
@@ -32,7 +32,7 @@ class FunctionLibNodeParamsSerializer(serializers.Serializer):
         super().is_valid(raise_exception=True)
         f_lib = QuerySet(FunctionLib).filter(id=self.data.get('function_lib_id')).first()
         if f_lib is None:
-            raise Exception(_('The library has been deleted'))
+            raise Exception(_('The function has been deleted'))
 
 
 class IFunctionLibNode(INode):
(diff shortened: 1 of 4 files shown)


apps/application/flow/step_node/function_lib_node/i_function_lib_node.py
```diff
@@ -32,7 +32,7 @@ def is_valid(self, *, raise_exception=False):
         super().is_valid(raise_exception=True)
         f_lib = QuerySet(FunctionLib).filter(id=self.data.get('function_lib_id')).first()
         if f_lib is None:
-            raise Exception(_('The library has been deleted'))
+            raise Exception(_('The function has been deleted'))
 
 
 class IFunctionLibNode(INode):
```
shaohuzhang1, line 38: There are no significant issues with the provided code snippet for `ISubscription` class. However, there's an inconsistency between the exception messages:

**Exception Message Inconsistency**:
```diff
-            raise Exception(_('The library has been deleted'))
+            raise Exception(_('The function has been deleted'))
```

Here, it looks like 'library' should be replaced with 'function' since you're dealing with function-related operations (`FunctionLib`). Adjusting this will ensure that the correct message is displayed.

**Optimization Suggestion (optional)**:
If you want to make the check more efficient by filtering directly on a specific field when possible, consider changing:
```python
f_lib = QuerySet(FunctionLib).filter(id=self.data.get('function_lib_id')).first()
```
to
```python
f_lib = FunctionLib.objects.filter(pk=self.data['function_lib_id']).last()  # Assuming 'id' and 'pk' point to the same field
```

This approach assumes that `FunctionLib.pk` corresponds to the desired identifier field used in your database schema. Adjustments may need to be made if these fields differ or do not exist in your model. 

These changes will improve readability and potentially performance based on the specifics of your database schema.

PR #2151, opened 2025-02-07
fix: Application history log interface loading status display error

shaohuzhang1:
fix: Application history log interface loading status display error

### DIFF
--- a/ui/src/views/log/index.vue
+++ b/ui/src/views/log/index.vue
@@ -277,7 +277,7 @@
   </LayoutContainer>
 </template>
 <script setup lang="ts">
-import { ref, onMounted, reactive, computed } from 'vue'
+import { ref, type Ref, onMounted, reactive, computed } from 'vue'
 import { useRoute } from 'vue-router'
 import { cloneDeep } from 'lodash'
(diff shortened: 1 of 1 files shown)


ui/src/views/log/index.vue
```diff
@@ -562,7 +564,7 @@ function saveCleanTime() {
     .then(() => {
       MsgSuccess(t('common.saveSuccess'))
       dialogVisible.value = false
-      getDetail()
+      getDetail(true)
     })
     .catch(() => {
       dialogVisible.value = false
```
shaohuzhang1, line 570: In this code snippet, there are several minor improvements and optimizations:

1. **Type Annotations**: Replace `any` with `Ref<any>` to improve type safety while utilizing Vue's reactivity system.

2. **Use of Parameters**: The parameter definition for the `getDetail` function allows passing an optional boolean (`isLoading`) which sets the loading status dynamically. This is beneficial when you want to control whether the loading should occur or not based on certain conditions.

3. **Function Parameter Naming**: Changed `id as string` to just `id` since it can be reliably converted to a string during the function call without needing explicit casting.

These changes enhance the readability, maintainability, and performance of the code by improving typesafety and reducing unnecessary conversions.

PR #2157, opened 2025-02-08
fix: defect of incorrect document names after importing CSV and docx files into the knowledge base

shaohuzhang1:
fix: defect of incorrect document names after importing CSV and docx files into the knowledge base  --bug=1052039 --user=王孝刚 【知识库】-压缩文件中包含csv、docx文件时，导入到知识库后，文档名称包含文件夹名称 https://www.tapd.cn/57709429/s/1651752

### DIFF
--- a/apps/common/handle/impl/csv_split_handle.py
+++ b/apps/common/handle/impl/csv_split_handle.py
@@ -8,6 +8,7 @@
 """
 import csv
 import io
+import os
 from typing import List
 
 from charset_normalizer import detect
@@ -28,7 +29,8 @@ class CsvSplitHandle(BaseSplitHandle):
     def handle(self, file, pattern_list: List, with_filter: bool, limit: int, get_buffer, save_image):
         buffer = get_buffer(file)
         paragraphs = []
-        result = {'name': file.name, 'content': paragraphs}
+        file_name = os.path.basename(file.name)
+        result = {'name': file_name, 'content': paragraphs}
         try:
             reader = csv.reader(io.TextIOWrapper(io.BytesIO(buffer), encoding=detect(buffer)['encoding']))
             try:
--- a/apps/common/handle/impl/doc_split_handle.py
+++ b/apps/common/handle/impl/doc_split_handle.py
@@ -7,6 +7,7 @@
     @desc:
 """
 import io
+import os
 import re
 import traceback
 import uuid
@@ -167,6 +168,7 @@ class DocSplitHandle(BaseSplitHandle):
              in elements])
 
     def handle(self, file, pattern_list: List, with_filter: bool, limit: int, get_buffer, save_image):
+        file_name = os.path.basename(file.name)
         try:
             image_list = []
             buffer = get_buffer(file)
@@ -180,9 +182,9 @@ class DocSplitHandle(BaseSplitHandle):
                 split_model = SplitModel(default_pattern_list, with_filter=with_filter, limit=limit)
         except BaseException as e:
             traceback.print_exception(e)
-            return {'name': file.name,
+            return {'name': file_name,
                     'content': []}
-        return {'name': file.name,
+        return {'name': file_name,
                 'content': split_model.parse(content)
                 }
 


apps/common/handle/impl/doc_split_handle.py
```diff
@@ -180,9 +182,9 @@ def handle(self, file, pattern_list: List, with_filter: bool, limit: int, get_bu
                 split_model = SplitModel(default_pattern_list, with_filter=with_filter, limit=limit)
         except BaseException as e:
             traceback.print_exception(e)
-            return {'name': file.name,
+            return {'name': file_name,
                     'content': []}
-        return {'name': file.name,
+        return {'name': file_name,
                 'content': split_model.parse(content)
                 }
```
shaohuzhang1, line 190: The provided code seems generally well-written but has one minor issue that could be improved:

**Issue:** The `file_name` variable is assigned using the wrong function `os.path.basename(file.name)`. To avoid this error, replace it with the correct method from the `pathlib` module.

Here's how you can modify the line:

```python
import pathlib

# ... other imports ...

def handle(self, file, pattern_list: List, with_filter: bool, limit: int, get_buffer, save_image):
    file_name = pathlib.Path(file.name).stem
```

This change will correctly obtain the base name without including the extension.

Additionally, consider renaming the parameter `get_image_id` to something more descriptive or remove it if not used within the function, as it doesn't appear to play a role at all. Here's the revised function signature:

```python
def handle(self, file, patterns_list: List, with_filter: bool, limit: int, get_buffer, save_image, use_get_image_id=False):
    # Use get_image_id as needed...
```

These changes make the code slightly more robust and easier to understand.

apps/common/handle/impl/csv_split_handle.py
```diff
@@ -28,7 +29,8 @@ class CsvSplitHandle(BaseSplitHandle):
     def handle(self, file, pattern_list: List, with_filter: bool, limit: int, get_buffer, save_image):
         buffer = get_buffer(file)
         paragraphs = []
-        result = {'name': file.name, 'content': paragraphs}
+        file_name = os.path.basename(file.name)
+        result = {'name': file_name, 'content': paragraphs}
         try:
             reader = csv.reader(io.TextIOWrapper(io.BytesIO(buffer), encoding=detect(buffer)['encoding']))
             try:
```
shaohuzhang1, line 36: The provided code generally looks appropriate and should work well without major issues. However, there are a few optimizations and minor improvements that can be made:

1. Remove unnecessary comments after imports.
2. Ensure that the `save_image` function is correctly defined or implemented.

```python
# Import necessary libraries at the beginning
from csv import reader
from io import TextIOWrapper, BytesIO
import os
from typing import List

from charset_normalizer import detect


class CsvSplitHandle(BaseSplitHandle):
    def handle(self, file, pattern_list: List, with_filter: bool, limit: int,
                         get_buffer, save_image=None):  # Optional argument for save_image
        buffer = get_buffer(file)
        paragraphs = []
        
        if file.name:
            base_name = os.path.basename(file.name)
        else:
            base_name = "unknown_file"
        
        result = {
            'name': base_name,
            'content': paragraphs
        }
        
        try:
            reader_obj = reader(TextIOWrapper(BytesIO(buffer), encoding=detect(b=buffer)['encoding']))
            
            for row in reader_obj:
                # Process each row here
                paragraph = ' '.join(row)  # Example processing
                paragraphs.append(paragraph[:limit] if limit > 0 else paragraph)

        except Exception as e:
            print(f"Error parsing CSV file: {e}")

        return result, save_image  # Assuming save_image might need to be invoked based on some condition
```

### Key Improvements:
- **Removed Unnecessary Comments**: Removed redundant comments after importing libraries.
- **Optional `save_image` Argument**: Added an optional argument `save_image` to the method signature, allowing it to be passed but not required.
- **Base Name Check**: Improved handling of filenames, providing a fallback name if no filename is available (`unknown_file`). 
- **Code Style**: Consistently indented and formatted for better readability.

PR #2172, opened 2025-02-08
fix: Fix speech recognition failed, the input box did not restore the default style.

shaohuzhang1:
fix: Fix speech recognition failed, the input box did not restore the default style.  --bug=1051969 --user=刘瑞斌 【应用】语音识别失败后输入框没有恢复默认的样式 https://www.tapd.cn/57709429/s/1651864

### DIFF
--- a/ui/src/components/ai-chat/component/chat-input-operate/index.vue
+++ b/ui/src/components/ai-chat/component/chat-input-operate/index.vue
@@ -484,6 +484,10 @@ const uploadRecording = async (audioBlob: Blob) => {
           })
         }
       })
+      .catch((error) => {
+        recorderLoading.value = false
+        console.error(`${t('chat.uploadFile.errorMessage')}:`, error)
+      })
   } catch (error) {
     recorderLoading.value = false
     console.error(`${t('chat.uploadFile.errorMessage')}:`, error)


ui/src/components/ai-chat/component/chat-input-operate/index.vue
```diff
@@ -484,6 +484,10 @@ const uploadRecording = async (audioBlob: Blob) => {
           })
         }
       })
+      .catch((error) => {
+        recorderLoading.value = false
+        console.error(`${t('chat.uploadFile.errorMessage')}:`, error)
+      })
   } catch (error) {
     recorderLoading.value = false
     console.error(`${t('chat.uploadFile.errorMessage')}:`, error)
```
shaohuzhang1, line 493: The code provided has a few issues that could be improved:

1. **Error Handling**: The `catch` block is currently inside the promise callback of `.then()`. This means it will only handle errors if there's an additional asynchronous operation that occurs after this point. To ensure all errors are handled correctly, you should move the `catch` outside both promises.

2. **Logging Error Messages**: Ensure that `$t('chat.uploadFile.errorMessage')` returns valid string contents before logging. If not, it might result in a `TypeError`.

3. **Consistency with Initial State**: It’s good to set `recorderLoading.value = false;` at the beginning of each function call to avoid confusion about its state during error handling or normal execution.

Here’s a revised version of the code addressing these points:

```javascript
const uploadRecording = async (audioBlob: Blob) => {
  try {
    // Initialize loading status first
    recorderLoading.value = true;

    await audioRecordPromise.then(recordedAudio => {
      console.log("Recording stopped.");
      // Upload recorded audio blob here...
    }).catch(error => {
      console.error(`${$t('chat.uploadFile.errorMessage')}:`, error);
      recorderLoading.value = false;
    });

    await uploadRecordingPromises.forEach(uploadPromise => {
      await uploadPromise.then(response =>
        console.log(`${response.statusText} : ${response.url}`)
      ).catch((error) => {
        recorderLoading.value = false;
        console.error(`${$t('chat.uploadFile.errorMessage')}:`, error);
      });
    });
  } catch (catchErr) {
    console.error(`Unexpected error occurred:`, catchErr);
    recorderLoading.value = false;
  }
};
```

### Key Changes Made:
- Moved initial setting of `recorderLoading.value = false;` at the start.
- Ensured proper nesting where necessary (`forEach`) and added missing brackets accordingly.
- Added consistency by logging relevant error messages.
- Simplified variable names where appropriate for clarity.

PR #2175, opened 2025-02-09
fix: Fix Simple Application preview show 404 error page

shaohuzhang1:
fix: Fix Simple Application preview show 404 error page  --bug=1052069 --user=刘瑞斌 【简易应用】- 调试预览中提问报404 https://www.tapd.cn/57709429/s/1651941

### DIFF
--- a/apps/application/flow/tools.py
+++ b/apps/application/flow/tools.py
@@ -18,6 +18,10 @@ from common.response import result
 
 class Reasoning:
     def __init__(self, reasoning_content_start, reasoning_content_end):
+        if not reasoning_content_start:
+            reasoning_content_start = "<think>"
+        if not reasoning_content_end:
+            reasoning_content_end = "</think>"
         self.content = ""
         self.reasoning_content = ""
         self.all_content = ""


apps/application/flow/tools.py
```diff
@@ -18,6 +18,10 @@
 
 class Reasoning:
     def __init__(self, reasoning_content_start, reasoning_content_end):
+        if not reasoning_content_start:
+            reasoning_content_start = "<think>"
+        if not reasoning_content_end:
+            reasoning_content_end = "</think>"
         self.content = ""
         self.reasoning_content = ""
         self.all_content = ""
```
shaohuzhang1, line 27: The provided code is mostly clear and follows Python's proper syntax. However, there are a few points of improvement:

1. **Default Values**: The default values for `reasoning_content_start` and `reasoning_content_end` can be added to the method signature directly using keyword arguments (`default`). This makes the function call more concise when no specific starting and ending tags are desired.

2. **Variable Naming Consistency**: Consider adding comments or renaming variables to make them more descriptive. For instance, `all_content` could use a different name like `combined_text`.

Here's an optimized version of the class with these improvements:

```python
class Reasoning:
    def __init__(self, reasoning_content_start="<think>", reasoning_content_end="</think>"):
        self.content = ""
        self.reasoning_content = ""
        # Use a less ambiguous variable name like 'text' instead of 'all_content'
        self.text = ""

    def add_reasoning(self, content):
        """Method to append text with reasoning tags."""
        if not isinstance(content, str):
            raise ValueError("Content must be a string.")
        
        # Append content after reasoning start tag
        rendered_content = f"{reasoning_content_start}{content}{reasoning_content_end}"
        
        # Check if we need to split into main and reasoning sections
        space_index = rendered_content.find(" ")
        if space_index != -1:
            self.main_content += rendered_content[:space_index]
            self.reasoning_content += rendered_content[space_index:]
        else:
            self.text += rendered_content

    def get_all_contents(self):
        return { "main": self.main_content, "reasoning": self.reasoning_content }
```

**Explanation**:
- Default values for `reasoning_content_start` and `reasoning_content_end` are now set in the method signature using keywords.
- A new method `add_reasoning` is introduced to concatenate content with the specified reasoning tags while also allowing separation based on spaces for clearer distinction between main content and reasoning.
- The logic for checking spaces and splitting strings is simplified within this new method.
- Added exception handling for non-string inputs to improve robustness.

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Using deepseek-coder in DeepSeek still reproduces the problem

[_vol1_1panel_1panel_apps_maxkb_maxkb_docker-compose.yml-20250214145253.log](https://github.com/user-attachments/files/18794619/_vol1_1panel_1panel_apps_maxkb_maxkb_docker-compose.yml-20250214145253.log)

Issue #2144, opened 2025-02-06
[Bug] 知识库上传文档向量化时，提示【文档名称】请确保此字段的字符数不超过 128 个。


Kingong:
### MaxKB Version

1.9.1

### Problem Description

知识库中上传文档向量化时，提示【文档名称】请确保此字段的字符数不超过 128 个。

### Steps to Reproduce

知识库中上传文档向量化时，提示【文档名称】请确保此字段的字符数不超过 128 个。

### Related log output

```shell

```

baixin513:
感谢反馈，后续版本修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version fixes

zyyfit:
1.10.0 已修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
1.10.0 Fixed

PR #2178, opened 2025-02-09
fix: Dealing with the issue of filename exceeding 128 characters when uploading documents(#2144)

shaohuzhang1:
fix: Dealing with the issue of filename exceeding 128 characters when uploading documents(#2144)

### DIFF
--- a/ui/src/utils/utils.ts
+++ b/ui/src/utils/utils.ts
@@ -1,3 +1,5 @@
+import { number } from 'echarts'
+
 export function toThousands(num: any) {
   return num?.toString().replace(/\d+/, function (n: any) {
     return n.replace(/(\d)(?=(?:\d{3})+$)/g, '$1,')
@@ -51,7 +53,6 @@ export function getImgUrl(name: string) {
 }
 // 是否是白名单后缀
 export function isRightType(name: string, type: string) {
-  console.log(name, type)
   return typeList[type].includes(fileType(name).toLowerCase())
 }
 
@@ -99,4 +100,11 @@ export function downloadByURL(url: string, name: string) {
   document.body.appendChild(a)
   a.click()
   document.body.removeChild(a)
+}
+
+// 截取文件名
+export function cutFilename(filename: string, num: number) {
+  const lastIndex = filename.lastIndexOf('.')
+  const suffix = lastIndex === -1 ? '' : filename.substring(lastIndex + 1)
+  return filename.substring(0, num - suffix.length - 1) + '.' + suffix
 }
--- a/ui/src/views/dataset/component/SetRules.vue
+++ b/ui/src/views/dataset/component/SetRules.vue
@@ -121,6 +121,7 @@
 <script setup lang="ts">
 import { ref, computed, onMounted, reactive, watch } from 'vue'
 import ParagraphPreview from '@/views/dataset/component/ParagraphPreview.vue'
+import { cutFilename } from '@/utils/utils'
 import documentApi from '@/api/document'
 import useStore from '@/stores'
 import type { KeyValue } from '@/api/type/common'
@@ -186,8 +187,12 @@ function splitDocument() {
     .postSplitDocument(fd)
     .then((res: any) => {
       const list = res.data
-      if (checkedConnect.value) {
-        list.map((item: any) => {
+
+      list.map((item: any) => {
+        if (item.name.length > 128) {
+          item.name = cutFilename(item.name, 128)
+        }
+        if (checkedConnect.value) {
           item.content.map((v: any) => {
             v['problem_list'] = v.title.trim()
               ? [
@@ -197,8 +202,9 @@ function splitDocument() {
                 ]
               : []
           })
-        })
-      }
+        }
+      })
+
       paragraphList.value = list
       loading.value = false
     })
--- a/ui/src/views/dataset/component/UploadComponent.vue
+++ b/ui/src/views/dataset/component/UploadComponent.vue
@@ -161,7 +161,11 @@
             </em>
           </p>
           <div class="upload__decoration">
-            <p>{{ $t('views.document.upload.formats') }}TXT、Markdown、PDF、DOCX、HTML、XLS、XLSX、CSV、ZIP</p>
+            <p>
+              {{
+                $t('views.document.upload.formats')
+              }}TXT、Markdown、PDF、DOCX、HTML、XLS、XLSX、CSV、ZIP
+            </p>
           </div>
         </div>
       </el-upload>
@@ -207,7 +211,9 @@ const form = ref({
 })
 
 const rules = reactive({
-  fileList: [{ required: true, message: t('views.document.upload.requiredMessage'), trigger: 'change' }]
+  fileList: [
+    { required: true, message: t('views.document.upload.requiredMessage'), trigger: 'change' }
+  ]
 })
 const FormRef = ref()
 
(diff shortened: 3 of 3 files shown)


ui/src/views/dataset/component/UploadComponent.vue
```diff
@@ -217,11 +223,17 @@ watch(form.value, (value) => {
...
 
 function downloadTableTemplate(type: string) {
-  documentApi.exportTableTemplate(`${type}${t('views.document.upload.template')}.${type == 'csv' ? type : 'xlsx'}`, type)
+  documentApi.exportTableTemplate(
+    `${type}${t('views.document.upload.template')}.${type == 'csv' ? type : 'xlsx'}`,
+    type
+  )
 }
 
 function radioChange() {
```
shaohuzhang1, line 239: The provided code appears to be written in Vue.js with composition API, using `ElUpload` from Element UI for file upload functionality. There are a few minor improvements and optimizations that can be made:

### Code Improvements

1. **Consistent Line Endings**: Ensure consistent use of line endings (`\n`). Windows uses `\r\n`, Unix systems use just `\n`.

2. **Simplify Template Literals**: In template literals like the one for `fileList validation rule`, you may have noticed some extra spaces or unnecessary characters. Ensure they are clean.

3. **Function Naming Consistency**: Make sure function names are consistently formatted and descriptive.

4. **Code Readability**:
   - Use blank lines between logical sections to improve readability.
   - Simplify complex expressions if possible within functions.

### Optimization Suggestions

1. **Avoid Redundant Data Concatenation**: The `t('views.document.upload.template')` should not appear multiple times without reason. Cache it before concatenating.

2. **Handle Edge Cases Gracefully**: Add checks in functions where errors might occur due to missing values or other unexpected issues.

Here is the updated code snippet with these considerations:

```vue
<template>
  <div class="document-upload-container">
    <!-- Your existing content -->
    <p v-if="$i18n.locale != 'zh-CN'>"Check if Chinese input needs adjustment</p>
    <el-upload action="" :show-file-list="false" @change="handleChange">
      <span style="font-weight:bold;">{{ $t("views.document.upload.drag") }}</span>
      <em>{{ $t("views.document.upload.hover") }}</em>
    </el-upload>

    <div class="upload__decoration">
      <p>{{ $t('views.document.upload.formats', ['TXT', 'Markdown', 'PDF', 'DOCX', 'HTML', 'XLS', 'XLSX', 'CSV', 'ZIP']) }}</p>
    </div>
  </div>
</template>

<script setup lang="ts">
import { ref, reactive } from 'vue';
import { translate as t } from '@/utils/i18n'; // Assuming you have translation utilities set up
// Import necessary APIs and types here

const form = ref({ fileList: [] });
const rules = reactive({
  fileList: [{ required: true, message: t('views.document.upload.requiredMessage'), trigger: 'change' }],
});

const handleSuccess = () => {
  console.log('File uploaded successfully');
};

const handleChange = ({ file }) => {
  // Handle file changes (e.g., add error checking)
};

function downloadTemplate(type: string) {
  const formatSuffix = type === 'csv' ? '' : '.xlsx'; // Conditional formatting suffix
  documentApi.exportQATemplate(`${type}${t('views.document.upload.template')}.${formatSuffix}`, type);
}

// Define similar functions for downloadTableTemplate and radioChange
</script>

<style scoped>
/* Your styles */
</style>
```

### Important Notes

- Replace placeholders like `documentApi` and `$i18n` with their actual implementations or dependencies.
   
This revision aims to make the code more readable, maintainable, and potentially faster by reducing redundancy and improving efficiency.

ui/src/views/dataset/component/SetRules.vue
```diff
@@ -197,8 +202,9 @@ function splitDocument() {
...
               : []
           })
-        })
-      }
+        }
+      })
+
       paragraphList.value = list
       loading.value = false
     })
```
shaohuzhang1, line 210: The provided code looks mostly clean with no major errors. However, here are some optimizations and potential improvements:

1. **Avoid Repeated Calculations**: The condition `checkedConnect.value` is checked twice inside the loop when adding content to problems. This could be optimized by extracting it into a separate variable (`const shouldAddTitle = checkedConnect.value;`). Then you can remove one of these checks.

2. **Use Array Methods Instead of Loops**: If there's a way to avoid explicitly mapping over arrays using higher-order functions like `.map()` directly within loops, that might increase performance slightly.

3. **Conditional Checks Inside Template**: Since this operation affects template rendering, consider moving conditional logic out of data processing unless absolutely necessary. For example, instead of checking each item for length in the mapped iteration, you could filter first before looping.

4. **Logging Improvements**: Adding logs or console statements at strategic points (like early return after network request) can help diagnose bottlenecks without increasing unnecessary load.

Here’s an improved version taking some aspects into consideration:

```typescript
<script setup lang="ts">
import { ref, computed, onMounted, reactive, watch } from 'vue'
import ParagraphPreview from '@/views/dataset/component/ParagraphPreview.vue'
+import { logError } from '@/utils/errorHandling' // Assuming custom error handling utilities are defined elsewhere

import documentApi from '@/api/document'
import type { KeyValue } from '@/api/type/common'

let shouldAddTitle = checkConnectedValue // Extracted conditional constant

function splitDocument() {
  loading.value = true
  
  documentApi.postSplitDocument(fd)
    .then(res => {
      const list = res.data
      
+      if (!list || !shouldAddTitle) return; // Avoid processing empty lists or non-connected cases
+      
      list.forEach(item => {
        if (item.name && item.name.length > 128) {
          item.name = cutFilename(ite.name, 128);
        }
        
        if (shouldAddTitle) {
          item.content.forEach(v => {
            v.problem_list =
              v.title.trim()?
                [v.title]:
                [];
          });
        }

        paragraphList.value.push(item);  
      });

      loading.value = false;
      return list;
    })
    .catch(error => {
      error?.response?.data?.errors?.forEach(errMsg => 
        console.error(`[document-api] Error while splitting`, errMsg)); // Log detailed errors
    });
}
</script>
```

These changes focus more on improving readability and efficiency by reducing duplicate checks and ensuring the main flow remains straightforward and maintainable.

ui/src/utils/utils.ts
```diff
@@ -100,3 +101,10 @@ export function downloadByURL(url: string, name: string) {
   a.click()
   document.body.removeChild(a)
 }
+
+// 截取文件名
+export function cutFilename(filename: string, num: number) {
+  const lastIndex = filename.lastIndexOf('.')
+  const suffix = lastIndex === -1 ? '' : filename.substring(lastIndex + 1)
+  return filename.substring(0, num - suffix.length - 1) + '.' + suffix
+}
```
shaohuzhang1, line 110: The provided code looks mostly correct, but there are a few improvements that could be made:

1. **Typing**: The import statement uses an incorrect syntax (`"number"` should be `Number`). Ensure consistent use of types.

2. **Variable Initialization**: In the `toThousands` function, initialize the replacement regex with proper grouping to avoid errors when `n` is shorter than 4 digits.

3. **Function Documentation**: Add some comments or documentation to explain what each piece of code does, especially more complex functions.

```javascript
/* eslint-disable prettier/prettier */
import { Number } from 'echarts';
export default class Utils {

 constructor() {}

 static toThousands(num: any): string {
   if (!num) {
     return num;
   }
   let parts = num.toString().split('.');
   parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ',');
   return parts.join('.');
 }

 // 获取图片路径
 static getImgUrl(name: string | null): string {
   const prefix = Config.baseUrls.imageUrlPrefix;
   const postfix = '/';
   return `${prefix}${name ?? ''}${postfix}`;
 }

 // 是否是白名单后缀
 static isRightType(name: string | null, type: string): boolean {
   if (!type || !fileType(name)) {
     return false;
   }
   const lowerCaseName = fileType(name).toLowerCase();
   const allowedTypes = typeList[type];
   return allowedTypes.includes(lowerCaseName);
 }

 /**
  * 下载文件
  * @param url 文件url
  * @param name 需要下载的本地名称 如果不传则默认使用原始文件名
  */
 static downloadByURL(url: string, name?: string): void {
   var a = document.createElement('a');
   a.href = url;
   a.download = name;
   a.style.display = 'none';
   document.body.appendChild(a);
   a.click(); // 模拟点击下载
   document.body.removeChild(a); // 移除该元素
 }

 /**
  * 剪裁文件名，只保留指定长度的部分，suffix为后缀类型，例：cutFilename('test.pdf', 8), 返回 "tet"
  * @param filename 文件名或者文件id
  * @param num 文件名最大长度
  * @returns 处理后的字符串
  */
 static cutFilename(filename: string | number, num: number): string {
   let lastIndex = typeof filename == 'string' ? filename.lastIndexOf('.') : '';
   let suffix = lastIndex !== -1 ? filename.substring(lastIndex + 1) : '';

   let finalName = typeof filename == 'number' ?
      this.getImgUrl(null)?.substring(`${this.toThousands(filename)}_`)
       .replaceAll('_', '').slice(-5) :
      filename;

   return finalName.substr(0, num - suffix.length - 1) + '.' + suffix;
 };

};
```

These changes ensure consistency throughout the code base, improve readability, and handle edge cases correctly.

PR #2179, opened 2025-02-09
fix: When the thinking process label is empty, do not process the thinking process

shaohuzhang1:
fix: When the thinking process label is empty, do not process the thinking process

### DIFF
--- a/apps/application/flow/tools.py
+++ b/apps/application/flow/tools.py
@@ -23,14 +23,23 @@ class Reasoning:
         self.all_content = ""
         self.reasoning_content_start_tag = reasoning_content_start
         self.reasoning_content_end_tag = reasoning_content_end
-        self.reasoning_content_start_tag_len = len(reasoning_content_start)
-        self.reasoning_content_end_tag_len = len(reasoning_content_end)
-        self.reasoning_content_end_tag_prefix = reasoning_content_end[0]
+        self.reasoning_content_start_tag_len = len(
+            reasoning_content_start) if reasoning_content_start is not None else 0
+        self.reasoning_content_end_tag_len = len(reasoning_content_end) if reasoning_content_end is not None else 0
+        self.reasoning_content_end_tag_prefix = reasoning_content_end[
+            0] if self.reasoning_content_end_tag_len > 0 else ''
         self.reasoning_content_is_start = False
         self.reasoning_content_is_end = False
         self.reasoning_content_chunk = ""
 
     def get_reasoning_content(self, chunk):
+        # 如果没有开始思考过程标签那么就全是结果
+        if self.reasoning_content_start_tag is None or len(self.reasoning_content_start_tag) == 0:
+            self.content += chunk.content
+            return {'content': chunk.content, 'reasoning_content': ''}
+        # 如果没有结束思考过程标签那么就全部是思考过程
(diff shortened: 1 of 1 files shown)


apps/application/flow/tools.py
```diff
@@ -23,14 +23,23 @@ def __init__(self, reasoning_content_start, reasoning_content_end):
...
+        # 如果没有开始思考过程标签那么就全是结果
+        if self.reasoning_content_start_tag is None or len(self.reasoning_content_start_tag) == 0:
+            self.content += chunk.content
+            return {'content': chunk.content, 'reasoning_content': ''}
+        # 如果没有结束思考过程标签那么就全部是思考过程
+        if self.reasoning_content_end_tag is None or len(self.reasoning_content_end_tag) == 0:
+            return {'content': '', 'reasoning_content': chunk.content}
         self.all_content += chunk.content
         if not self.reasoning_content_is_start and len(self.all_content) >= self.reasoning_content_start_tag_len:
             if self.all_content.startswith(self.reasoning_content_start_tag):
```
shaohuzhang1, line 45: The provided Python code includes some minor improvements, such as handling cases where either start or end tags might be missing by setting appropriate attributes to zero or an empty string when necessary. However, there is room for further optimization and clarity:

1. **Simplify Logic**: The logic for determining start/end of the reasoning content can be simplified by checking conditions once and directly returning results based on these checks.

2. **Avoid Redundant Checks**: The redundant `len()` calls for both tags within the same condition can be removed.

3. **Use Default Parameters**: Consider using default parameters in method signatures that accept optional arguments with meaningful names, which improves readability and maintainability.

```python
def __init__(self, reasoning_content_start="", reasoning_content_end=""):
        self.all_content = ""
        self.reasoning_content_start_tag = reasoning_content_start
        self.reasoning_content_end_tag = reasoning_content_end

        self.start_prefix_length = len(reasoning_content_start) if reasoning_content_start else 0
        self.end_suffix_length = len(reasoning_content_end) if reasoning_content_end else 0
        self.end_tag_prefix = reasoning_content_end[0] if self.end_suffix_length > 0 else ''

    def get_reasoning_content(self, chunk):
        all_content = self.all_content + chunk.content

        # Start of reasoning content detected
        if all_content.startswith(self.reasoning_content_start_tag):
            result = {
                'content': '',
                'reasoning_content': ''.join(all_content.split(chunk.content)[1:])
            }
            return result

        # End of reasoning content detected
        elif all_content.endswith(self.reasoning_content_end_tag):
            reason_text = "".join(all_content.split(chunk.content))
            return {
                'content': chunk.content,
                'reasoning_content': reason_text[len(self.reasoning_content_end_tag):].strip()
            }

        # No tag matches (either at the beginning or the middle)
        else:
            self.all_content = ""
            return {'content': chunk.content, 'reasoning_content': ''}
```

### Key Improvements:
- **Single Check for Tags**: The code now uses only one check (`startswith` followed by `endswith`) instead of two separate checks, reducing redundancy.
- **Default Parameter Names**: Added descriptive parameter names (`reasoning_content_start`, `reasoning_content_end`) to improve understanding of their purpose.
- **String Manipulation Simplified**: Used `.split()` and slicing to simplify extracting text between tags.
- **Return Statement Clarity**: Each branch of the logic returns a dictionary with clear keys indicating whether it contains main content or reasoning content.

PR #2181, opened 2025-02-09
fix: Excerpt from the thinking process

shaohuzhang1:
fix: Excerpt from the thinking process

### DIFF
--- a/apps/application/flow/tools.py
+++ b/apps/application/flow/tools.py
@@ -46,7 +46,10 @@ class Reasoning:
                 self.reasoning_content_is_start = True
                 self.reasoning_content_chunk = self.all_content[self.reasoning_content_start_tag_len:]
             else:
-                self.reasoning_content_is_end = True
+                if not self.reasoning_content_is_end:
+                    self.reasoning_content_is_end = True
+                    self.content += self.all_content
+                    return {'content': self.all_content, 'reasoning_content': ''}
         else:
             if self.reasoning_content_is_start:
                 self.reasoning_content_chunk += chunk.content


apps/application/flow/tools.py
```diff
@@ -46,7 +46,10 @@ def get_reasoning_content(self, chunk):
...
                 self.reasoning_content_chunk = self.all_content[self.reasoning_content_start_tag_len:]
             else:
-                self.reasoning_content_is_end = True
+                if not self.reasoning_content_is_end:
+                    self.reasoning_content_is_end = True
+                    self.content += self.all_content
+                    return {'content': self.all_content, 'reasoning_content': ''}
         else:
             if self.reasoning_content_is_start:
                 self.reasoning_content_chunk += chunk.content
```
shaohuzhang1, line 55: There are no immediate issues with the provided code snippet. However, there is an unnecessary assignment within the `if` statement to modify `self.reasoning_content_is_end`.

I suggest optimizing it by removing that line since it unnecessarily increments the length of `self.reasoning_content`. Also, ensure that `return {'content': '', 'reasoning_content': ''}` is correctly placed within a conditional block where it makes sense to handle when both content parts have been fully collected.

Optimized version:

```python
def get_reasoning_content(self, chunk):
    if self.reasoning_content_is_start:
        self.reasoning_content_chunk += chunk.content

    elif not self.reasoning_content_is_complete:
        # Check if we've reached the end tag
        if 'end_of_reaso' in chunk.all_content.strip():  # Adjust condition based on actual tagging
            self.reasoning_content_is_end = True
            self.all_content += chunk.all_content
            return {'content': "", 'reasoning_content': self.reasoning_content_chunk}

    else:  # Both chunks processed
        return {'content': self.all_content, 'reasoning_content': self.reasoning_content_chunk}
```

In this version, I assume that you meant to include additional logic to check for the end tag (`end_of_reaso`) and process the concatenated results appropriately. If this isn't the case, please adjust accordingly.

PR #2184, opened 2025-02-09
fix: Simple application dialogue, non streaming, unable to respond to data

shaohuzhang1:
fix: Simple application dialogue, non streaming, unable to respond to data

### DIFF
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -275,7 +275,6 @@ class BaseChatStep(IChatStep):
                 request_token = 0
                 response_token = 0
             write_context(self, manage, request_token, response_token, chat_result.content)
-            reasoning.get_reasoning_content(chat_result)
             reasoning_result = reasoning.get_reasoning_content(chat_result)
             content = reasoning_result.get('content')
             if 'reasoning_content' in chat_result.response_metadata:


apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
```diff
@@ -275,7 +275,6 @@ def execute_block(self, message_list: List[BaseMessage],
                 request_token = 0
                 response_token = 0
             write_context(self, manage, request_token, response_token, chat_result.content)
-            reasoning.get_reasoning_content(chat_result)
             reasoning_result = reasoning.get_reasoning_content(chat_result)
             content = reasoning_result.get('content')
             if 'reasoning_content' in chat_result.response_metadata:
```
shaohuzhang1, line 280: The provided code snippet appears to be part of a function within an AI framework that deals with executing blocks in a conversation. However, it lacks context about the data types, expected outputs, and dependencies used by `manage`, `write_context`, `reasoning.get_reasoning_content`, and `chat_result`. Based on this information, here are some potential issues and suggestions:

1. **Variable Initialization**:
   - Ensure that `request_token` and `response_token` are properly initialized before use. They should be either default values or derived from other variables.
   
2. **Reasoning Content Handling**:
   - The method `get_reasoning_content` is called twice on the same object (`chat_result`). This might lead to unexpected behavior if only one call is necessary. It's important to understand why both calls are needed.
   - Consider storing the result of the first call so that you don't repeat work when the second call is made.

3. **Content Retrieval**:
   - Check if there's a possibility that `content` returns `None`. If not handled appropriately, it could cause errors later in the process.

4. **Response Metadata Check**:
   - Validate that `'reasoning_content'` exists in `chat_result.response_metadata`. This ensures that the logic that relies on this metadata does not fail due to missing keys.

5. **Error Handling**:
   - Implement error handling around critical operations like calling methods and accessing dictionary attributes to prevent crashes if inputs are invalid.

6. **Performance Optimization**:
   - Review how `execute_block` interacts with other functions and ensure that no unnecessary computations are performed multiple times.

To provide more specific guidance, additional details about these functions and their interdependencies would be helpful. Otherwise, these points cover general areas where improvements could potentially be made to improve readability, robustness, and efficiency of the code.

PR #2188, opened 2025-02-10
fix: Workflow non streaming dialogue cannot respond

shaohuzhang1:
fix: Workflow non streaming dialogue cannot respond

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -335,7 +335,9 @@ class WorkflowManage:
         answer_tokens = sum([row.get('answer_tokens') for row in details.values() if
                              'answer_tokens' in row and row.get('answer_tokens') is not None])
         answer_text_list = self.get_answer_text_list()
-        answer_text = '\n\n'.join(answer['content'] for answer in answer_text_list)
+        answer_text = '\n\n'.join(
+            '\n\n'.join([a.get('content') for a in answer]) for answer in
+            answer_text_list)
         self.work_flow_post_handler.handler(self.params['chat_id'], self.params['chat_record_id'],
                                             answer_text,
                                             self)


apps/application/flow/workflow_manage.py
```diff
@@ -335,7 +335,9 @@ def run_block(self, language='zh'):
         answer_tokens = sum([row.get('answer_tokens') for row in details.values() if
                              'answer_tokens' in row and row.get('answer_tokens') is not None])
         answer_text_list = self.get_answer_text_list()
-        answer_text = '\n\n'.join(answer['content'] for answer in answer_text_list)
+        answer_text = '\n\n'.join(
+            '\n\n'.join([a.get('content') for a in answer]) for answer in
+            answer_text_list)
         self.work_flow_post_handler.handler(self.params['chat_id'], self.params['chat_record_id'],
                                             answer_text,
                                             self)
```
shaohuzhang1, line 343: The provided code has two primary issues that need attention:

1. **Double Join Operation**: 
   - The line `answer_text = '\n\n'.join('\n\n'.join([a.get('content') for a in answer])) for answer in answer_text_list)` contains unnecessary double joins.

2. **Handling of List of Lists**:
   - Since each element in `answer_text_list` should be another list itself (i.e., `[{'content': 'text'}, {'content': 'text'}]`) and you want to join these lists with `\n\n`, it's simpler to directly flatten the entire list of lists before joining them together without nested joins.

Here’s an optimized version of the code addresses both issues:

```python
@@ -335,7 +335,8 @@ def run_block(self, language='zh'):
     answer_tokens = sum([row.get('answer_tokens') for row in details.values() if
                          'answer_tokens' in row and row.get('answer_tokens') is not None])
     answer_text_list = self.get_answer_text_list()

    # Flatten the list of dictionaries within answer_text_list
    flat_answer_texts = [item.get('content') for sublist in answer_text_list for item in sublist]
    
    # Join all elements into a single string separated by \n\n
    answer_text = '\n\n'.join(flat_answer_texts)

    self.work_flow_post_handler.handler(self.params['chat_id'], self.params['chat_record_id'],
                                         answer_text,
                                         self)
```

With this change, we ensure that each entry from `answer_text_list` is processed separately without needing nested joins, resulting in more straightforward and efficient handling of multi-level data structures.

PR #2189, opened 2025-02-10
fix: Fix xinference cannot add deepseek-vl-chat model

shaohuzhang1:
fix: Fix xinference cannot add deepseek-vl-chat model  --bug=1052077 --user=刘瑞斌 【github#2051】【模型管理】xinference添加issue中提供的视觉模型失败 https://www.tapd.cn/57709429/s/1652171

### DIFF
--- a/apps/setting/models_provider/impl/base_chat_open_ai.py
+++ b/apps/setting/models_provider/impl/base_chat_open_ai.py
@@ -104,7 +104,7 @@ class BaseChatOpenAI(ChatOpenAI):
                 if generation_chunk.message.usage_metadata is not None:
                     self.usage_metadata = generation_chunk.message.usage_metadata
                 # custom code
-                if 'reasoning_content' in chunk['choices'][0]['delta']:
+                if len(chunk['choices']) > 0 and 'reasoning_content' in chunk['choices'][0]['delta']:
                     generation_chunk.message.additional_kwargs["reasoning_content"] = chunk['choices'][0]['delta'][
                         'reasoning_content']
 


apps/setting/models_provider/impl/base_chat_open_ai.py
```diff
@@ -104,7 +104,7 @@ def _stream(
                 if generation_chunk.message.usage_metadata is not None:
                     self.usage_metadata = generation_chunk.message.usage_metadata
                 # custom code
-                if 'reasoning_content' in chunk['choices'][0]['delta']:
+                if len(chunk['choices']) > 0 and 'reasoning_content' in chunk['choices'][0]['delta']:
                     generation_chunk.message.additional_kwargs["reasoning_content"] = chunk['choices'][0]['delta'][
                         'reasoning_content']
```
shaohuzhang1, line 110: Your code snippet looks generally correct, but there are a few minor improvements that could be made:

1. **Type Checking**: It's good practice to add type hints to make the code more readable and maintainable.

2. **Simplified Conditionals**: The current condition can be simplified slightly:
   ```python
   if chunk.get('choices', [{}])[0].get('delta') is not None and 'reasoning_content' in chunk.get('choices', [{}])[0]['delta']:
   ```

3. **Performance Considerations**: If `chunk` or its structure might change, consider using `.items()` or similar methods to handle cases where keys might not exist (though this doesn't affect your current logic).

Here's the updated code with these considerations:

```python
def _stream(
        self,
        stream_iterator,
        max_tokens=None,
        stop_sequences=None):
    for chunk in stream_iterator:
        # ...
        generation_chunk.message.usage_metadata = generation_chunk.message.usage_metadata

        # custom code
        delta = chunk.get('choices', [{}])[0].get('delta')
        if delta is not None and 'reasoning_content' in delta:
            generation_chunk.message.additional_kwargs["reasoning_content"] = delta["reasoning_content"]
```

These changes enhance clarity and readability without altering the functionality of your existing code.

PR #2198, opened 2025-02-10
fix: zip Icon missing

shaohuzhang1:
fix: zip Icon missing


ui/src/utils/utils.ts
```diff
@@ -49,7 +49,7 @@ export function getImgUrl(name: string) {
   const list = Object.values(typeList).flat()
 
   const type = list.includes(fileType(name).toLowerCase()) ? fileType(name).toLowerCase() : 'unknow'
-  return new URL(`../assets/${type}-icon.svg`, import.meta.url).href
+  return new URL(`../assets/fileType/${type}-icon.svg`, import.meta.url).href
 }
 // 是否是白名单后缀
 export function isRightType(name: string, type: string) {
```
shaohuzhang1, line 55: The provided code looks generally clean, but here are some points to consider for improvement:

1. **File Type Check**: The current implementation checks if `fileType(name)` (presumably a helper function that returns the file extension of a name) is included in the `list` array. However, this won't directly match against known image types like `.jpg`, `.png`, etc.

2. **URL Construction**: The URL construction is correct, but it assumes that all valid image formats have corresponding icon files named `<ext>-icon.svg`.

3. **Optimization Suggestion**: Consider adding a mapping from supported image extensions to their corresponding icons to reduce redundancy and improve readability when setting up the `typeList` object.

4. **Error Handling**: Add basic error handling to manage cases where `fileType(name)` might not return an expected value.

Here's an updated version with these considerations:

```javascript
export function getImgUrl(name: string): string | null {
  const list = Object.values(typeList).flat();

  // Get the file extension in lowercase without dot
  const fileType = () => name.split('.').pop()?.toLowerCase();
  const typeLowerCase = fileType() || '';

  // Find a matching icon type in the typeList
  const availableTypes = [...new Set(list)];
  const iconName = availableTypes.find(t =>
    t === typeLowerCase ||
    `${t.charAt(0).toUpperCase()}${typeLowerCase.slice(1)}-img`.includes(typeLowerCase)
  );

  if (!iconName) {
    console.error(`No icon found for type ${typeLowerCase}`);
    return null; // or handle it gracefully based on your needs
  }

  // Construct and return the URL
  return new URL(`../assets/${iconName.toLowerCase().replace('-', '')}.svg`, import.meta.url).href;
}

// Whether it's a whitelist after suffixes
export function isRightType(name: string, type: string): boolean {
  // Simplified example assuming 'type' is already sanitized or correctly passed
  return typeList[type] && Object.values(typeList[type]).some(ext => ext === name);
}
```

### Explanation:
- **Available Types**: A set of available types (derived from `typeList`) ensures uniqueness and simplifies finding the right icon name.
- **Icon Name Calculation**: The logic tries to find a match by looking for both exact matches and variations with capitalization adjustments.
- **Null Return**: If no icon is found, a message is logged, and `null` is returned to indicate the absence of a suitable icon.
- **Sanitation Concerns**: Ensure `typeName()` is properly sanitizing the input before searching within `typeList`.

This approach should make the code more robust and flexible while maintaining clarity.

PR #2201, opened 2025-02-10
fix: User default language based on browser language

shaohuzhang1:
fix: User default language based on browser language


apps/application/migrations/0026_alter_applicationaccesstoken_language.py
```diff
@@ -0,0 +1,18 @@
...
+        ('application', '0025_alter_application_prologue'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='applicationaccesstoken',
+            name='language',
+            field=models.CharField(default=None, max_length=10, null=True, verbose_name='语言'),
+        ),
+    ]
```
shaohuzhang1, line 18: Your migration looks generally correct, but there are a few things you might consider:

1. **Default Value for `null`:** The default value of `None` can be ambiguous because in Python, many types cannot be determined from just `None`. If you want to explicitly indicate that null is allowed, it's better to use `models.TextField(null=True)` instead.

2. **Verbose Name:** In Django, you typically only need to specify the verbose name if you're not using the default translation system, which is common these days. You could remove it entirely:
   ```python
   field=models.CharField(max_length=10, null=True),
   ```

3. **Consistent Field Definition:** It would also make sense to have all fields defined consistently (e.g., whether they're required or optional). For example, `blank=True` could be added for any non-required string fields:
   ```python
   field=models.CharField(max_length=10, blank=True, null=True),
   ```

Here's an optimized version based on these suggestions:

```python
@@ -0,0 +1,18 @@
+# Generated by Django 4.2.18 on 2025-02-10 06:01

+class Migration(migrations.Migration):

+    dependencies = [
+        ('application', '0025_alter_application_prologue'),
+    ]

+    operations = [
+        migrations.AlterField(
+            model_name='applicationaccesstoken',
+            name='language',
+            field=models.CharField(max_length=10, blank=True, null=True),
+        ),
+    ]
```

This version simplifies the model definition while maintaining clarity.

apps/users/migrations/0006_alter_user_language.py
```diff
@@ -0,0 +1,18 @@
...
+        ('users', '0005_user_language'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='language',
+            field=models.CharField(default=None, max_length=10, null=True, verbose_name='语言'),
+        ),
+    ]
```
shaohuzhang1, line 18: No issues found in your migration script. The structure is correct and follows best practices for altering fields in Django database migrations.

apps/application/models/api_key_model.py
```diff
@@ -50,7 +50,7 @@ class ApplicationAccessToken(AppModelMixin):
                             , default=list)
     show_source = models.BooleanField(default=False, verbose_name="是否显示知识来源")
 
-    language = models.CharField(max_length=10, verbose_name="语言", default=get_language)
+    language = models.CharField(max_length=10, verbose_name="语言", default=None, null=True)
 
     class Meta:
         db_table = "application_access_token"
```
shaohuzhang1, line 56: This code has some minor adjustments to improve clarity and functionality:

1. **Null Check**: The `language` field is now set to accept `null=True`, which allows for the possibility that it could be left empty.

```python
    language = models.CharField(max_length=10, verbose_name="语言", default=None, null=True)
```

2. **Default Value Removal**: Removed the call to `get_language` in the default value of the `language` field since it no longer returns an object (it used to return either 'zh_CN' or 'en_US'). Instead, simply set it to `None`.

3. **Minor Typographical Correction**:
   - Changed `verbose_name("是否显示知识来源")` to `verbose_name='是否显示知识来源'`.
   
This should make the function more robust and handle edge cases like when the `default` option needs to be `None`.

PR #2204, opened 2025-02-10
fix: Fix default language when chat language is null

shaohuzhang1:
fix: Fix default language<br>fix: Fix default language when chat language is null

### DIFF
--- a/ui/src/stores/modules/application.ts
+++ b/ui/src/stores/modules/application.ts
@@ -2,7 +2,7 @@ import { defineStore } from 'pinia'
 import applicationApi from '@/api/application'
 import applicationXpackApi from '@/api/application-xpack'
 import { type Ref } from 'vue'
-
+import { getBrowserLang } from '@/locales/index'
 import useUserStore from './user'
 const useApplicationStore = defineStore({
   id: 'application',
@@ -79,8 +79,7 @@ const useApplicationStore = defineStore({
         applicationApi
           .getAppProfile(loading)
           .then((res) => {
-            sessionStorage.setItem('language', res.data?.language)
-
+            sessionStorage.setItem('language', res.data?.language || getBrowserLang())
             resolve(res)
           })
           .catch((error) => {
--- a/ui/src/stores/modules/user.ts
+++ b/ui/src/stores/modules/user.ts
@@ -34,7 +34,7 @@ const useUserStore = defineStore({
     getLanguage() {
       return this.userType === 1
         ? localStorage.getItem('MaxKB-locale') || getBrowserLang()
-        : sessionStorage.getItem('language')
+        : sessionStorage.getItem('language') || getBrowserLang()
     },
     showXpack() {
       return this.isXPack
@@ -124,7 +124,7 @@ const useUserStore = defineStore({
     async profile() {
       return UserApi.profile().then(async (ok) => {
         this.userInfo = ok.data
-        useLocalStorage(localeConfigKey, 'zh-CN').value = ok.data?.language
+        useLocalStorage(localeConfigKey, 'en-US').value = ok.data?.language || this.getLanguage()
         return this.asyncGetProfile()
       })
     },
@@ -171,7 +171,7 @@ const useUserStore = defineStore({
       return new Promise((resolve, reject) => {
         UserApi.postLanguage({ language: lang }, loading)
           .then(async (ok) => {
-            useLocalStorage(localeConfigKey, 'zh-CN').value = lang
+            useLocalStorage(localeConfigKey, 'en-US').value = lang
             window.location.reload()
             resolve(ok)
           })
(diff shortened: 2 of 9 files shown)


ui/src/stores/modules/user.ts
```diff
@@ -171,7 +171,7 @@ const useUserStore = defineStore({
       return new Promise((resolve, reject) => {
         UserApi.postLanguage({ language: lang }, loading)
           .then(async (ok) => {
-            useLocalStorage(localeConfigKey, 'zh-CN').value = lang
+            useLocalStorage(localeConfigKey, 'en-US').value = lang
             window.location.reload()
             resolve(ok)
           })
```
shaohuzhang1, line 177: The code looks mostly correct but here are some minor improvements:

1. **Code Duplication:**
   The `useLocalStorage` call with `'zh-CN'` is repeated twice at lines 126 and 173. This could be optimized to a single call by using an array of languages and checking if the current language is in that list.

2. **Default Language Handling:**
   It seems like you have two different behaviors for non-authenticated users regarding language handling:
   - For authenticated users (`this.userType === 1`), it prefers `localStorage`.
   - Otherwise, it defaults to `sessionStorage`.

   Ensure these behaviors align as expected for both cases (authenticated vs. unauthenticated).

3. **Async/await Style:**
   Although not explicitly checked for async / await style consistency, consider whether all asynchronous operations should follow consistent patterns throughout your application.

Here's a slightly revised version incorporating the above points:

```javascript
import { defineStore } from "pinia";

const useUserStore = defineStore({
  id: 'user',
  
  state: () => ({
    userType: null,
    userLang: '',
    userInfo: {},
  }),
  
  getters: {
    getLanguage() {
      return this.$auth.isAuthenticated 
        ? localStorage.getItem('MaxKB-locale') || getBrowserLang()
        : sessionStorage.getItem('language');
    },
    
    showXpack() {
      // Assuming $auth.isAuthenticated is part of your auth module
      return this.$auth.isAuthenticated && this.useAuthModule.someFunction(); // Replace with actual function name
    }
  },
  
  actions: {
    ...mapMutations(['setError']),
    
    setUserInfo(userInfo) {
      this.userInfo = userInfo;
    },
    
    async postLanguage(langParam) {
      try {
        this.setError(null);
        
        const response = await UserApi.postLanguage({ language: langParam });
        this.setUserInfo(response.data);
        
        // Use local storage based on authentication status.
        setLocale(this.$auth.isAuthenticated ? localeConfigKey : 'storage_key', langParam); // Replace with your custom logic
        
        window.location.reload();
      } catch (error) {
        this.setError(error.message);
      }
    }
  }
});

// Helper function to determine appropriate localization strategy
function setLocale(key, value) {
  // Implement how you want to handle setting the locale (e.g., via store or environment variables)
}

export default useUserStore;
```

### Key Changes:
- Removed duplicated `useLocalStorage('zh-CN')` calls.
- Assumed `$auth.isAuthenticated` exists in your project context (modify as necessary).
- Simplified action structure slightly.
- Created a helper function `setLocale` for better organization and clarity. Adjust according to your app’s needs.

ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
```diff
@@ -427,7 +427,7 @@ const form = ref<any>({
 
 const xpackForm = ref<any>({
   show_source: false,
-  language: 'zh-CN',
+  language: getBrowserLang(),
   show_history: false,
   draggable: false,
   show_guide: false,
```
shaohuzhang1, line 433: The provided code has no major irregularities or potential issues based on the information available up to February 2025. Here are some recommendations for enhancement:

1. **Avoid Redundancy**: The `langList` variable is now being imported twice (once with and once without the browser language). This might cause confusion and could be simplified.

2. **Consistent Language Handling**: Ensure that if you need both system-defined languages (`langList`) and browser-detected language, they are handled consistently to avoid redundant logic.

3. **Update Locale Import Paths**: If these locale imports have been moved since October 1, make sure their import paths reflect this change. For example:
   ```javascript
   // Before: import { language } from '@/ locales'
   // After: import { language } from '@/locales/index.js'  # Adjusted path accordingly
   ```

4. **Performance Considerations**: Since the `getBrowserLang()` function is called twice (on load and when the user changes settings), consider caching it if performance becomes an issue.

Overall, the code looks clean and functional for its intended purpose.

ui/src/stores/modules/application.ts
```diff
@@ -79,8 +79,7 @@ const useApplicationStore = defineStore({
         applicationApi
           .getAppProfile(loading)
           .then((res) => {
-            sessionStorage.setItem('language', res.data?.language)
-
+            sessionStorage.setItem('language', res.data?.language || getBrowserLang())
             resolve(res)
           })
           .catch((error) => {
```
shaohuzhang1, line 85: In this code, there seems to be an issue with setting `sessionStorage.setItem('language')`. When using the variable directly (`getBrowserLang`) without properly importing it or passing any arguments, it might result in errors. Ensure that `getBrowserLang` is correctly imported and used.

Suggested optimizations include:

- Importing `getBrowserLang` at the beginning of the file to avoid confusion.
- Verifying that `getBrowserLang()` returns a value before assigning it to `localStorage`.
 
Also, consider handling cases where `res.data.language` might not be available (e.g., null or undefined). This could make your app more robust.

PR #2223, opened 2025-02-10
fix: markdown preview i18n

shaohuzhang1:
fix: Style optimization<br>perf: SCAN login setting views Style optimization<br>fix: markdown preview i18n

### DIFF
--- a/ui/src/components/markdown/MdPreview.vue
+++ b/ui/src/components/markdown/MdPreview.vue
@@ -1,8 +1,23 @@
 <template>
-  <MdPreview noIconfont noPrettier :codeFoldable="false" v-bind="$attrs" />
+  <MdPreview :language="language" noIconfont noPrettier :codeFoldable="false" v-bind="$attrs" />
 </template>
 
 <script setup lang="ts">
-import { MdPreview } from 'md-editor-v3'
+import { computed } from 'vue'
+import { MdPreview, config } from 'md-editor-v3'
+import { getBrowserLang } from '@/locales/index'
+import useStore from '@/stores'
+// 引入公共库中的语言配置
+import ZH_TW from '@vavt/cm-extension/dist/locale/zh-TW'
+
 defineOptions({ name: 'MdPreview' })
+const { user } = useStore()
+const language = computed(() => user.getLanguage() || getBrowserLang() || '')
+config({
+  editorConfig: {
+    languageUserDefined: {
+      'zh-Hant': ZH_TW
+    }
+  }
+})
 </script>
--- a/ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
+++ b/ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
@@ -174,7 +174,7 @@
           <h5 class="mb-8">
             {{ $t('views.applicationOverview.appInfo.SettingDisplayDialog.languageLabel') }}
           </h5>
-          <el-select v-model="form.language">
+          <el-select v-model="form.language" clearable>
             <el-option
               v-for="item in langList"
               :key="item.value"
@@ -383,7 +383,7 @@ import type { FormInstance, FormRules, UploadFiles } from 'element-plus'
 import { isAppIcon, isWorkFlow } from '@/utils/application'
 import applicationXpackApi from '@/api/application-xpack'
 import { MsgSuccess, MsgError } from '@/utils/message'
-import { getBrowserLang, langList, t } from '@/locales'
+import { langList, t } from '@/locales'
 import useStore from '@/stores'
 import { cloneDeep } from 'lodash'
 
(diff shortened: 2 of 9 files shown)


ui/src/views/authentication/component/SCAN.vue
```diff
@@ -230,46 +230,7 @@ function showDialog(platform: Platform) {
...
-.border-t {
-  border-top: 1px solid #ebeef5;
-}
-
-.vertical-middle {
-  vertical-align: middle;
+.scan-height {
+  height: calc(100vh - var(--app-header-height) - var(--app-view-padding) * 2 - 70px);
 }
 </style>
```
shaohuzhang1, line 236: There is an issue with the `scrollbar` component in the Vue.js template. It's likely not being used correctly because `v-loading` takes precedence over other styles when applied to a `div`. 

Here’s how you can address it:

### Suggestion

Use the `overflow-y-auto` CSS property inside the `<div>` element where you want to apply scrolling. Here’s the updated snippet:

```html
<template>
  <div v-loading="loading" class="container-class">
    <!-- Your existing content here -->
  </div>
</template>

<script>
export default {
  // ...
};
</script>

<style scoped>
.container-class {
  overflow-y-auto; /* This will add scrollbar if needed */
  padding: var(--app-view-padding); /* Add this for consistent layout spacing */
}

/* Other style rules remain unchanged */
```

By applying `overflow-y-auto`, you ensure that scrollbars appear only when necessary, enhancing user interactions without affecting other elements.

This should resolve the inconsistency between using `v-loading` and other styling approaches. Let me know if you need further assistance!

ui/src/components/markdown/MdPreview.vue
```diff
@@ -1,8 +1,23 @@
...
+const { user } = useStore()
+const language = computed(() => user.getLanguage() || getBrowserLang() || '')
+config({
+  editorConfig: {
+    languageUserDefined: {
+      'zh-Hant': ZH_TW
+    }
+  }
+})
 </script>
```
shaohuzhang1, line 23: There are several improvements and optimizations you can make to the code:

1. **Remove Redundant Import**:
   The import statement for `Zh_TW` is redundant since it's already imported at the beginning as part of the extension.

2. **Use Computed Properties Correctly**:
   In Vue.js, computed properties should be declared outside the lifecycle hooks (`setup` function). You have correctly defined it, but ensure that this works for your specific scenario.

3. **Ensure Consistency**:
   Make sure all keys in the computed property object match those used in other configuration objects like `editorConfig`.

4. **Check Compatibility**:
   Ensure that the versions of libraries you're using (like `md-editor-v3`) are compatible with Vue 3.

Here's a revised version of your code:

```html
<template>
  <MdPreview :language="language" noIconfont noPrettier :codeFoldable="false" v-bind="$attrs" />
</template>

<script setup lang="ts">
import { computed } from 'vue';
import { MdPreview, config } from 'md-editor-v3';
import { getBrowserLang } from '@/locales/index';
import useStore from '@/stores';

defineOptions({ name: 'MdPreview' });

const { user } = useStore();
const language = computed(() => user.getLanguage() || getBrowserLang());

config({
  editorConfig: {
    languageUserDefined: {
      'zh-Hant': require('@vavt/cm-extension/dist/locale/zh-TW')
    }
  }
});
</script>
```

### Key Changes Made:
- Removed the redundancy in importing `ZH_TW`.
- Ensured consistent usage of computed properties.
- Added an explicit type hint for `require`, though TypeScript might handle dynamic imports automatically in later versions.

ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
```diff
@@ -427,7 +427,7 @@ const form = ref<any>({
 
 const xpackForm = ref<any>({
   show_source: false,
-  language: getBrowserLang(),
+  language: '',
   show_history: false,
   draggable: false,
   show_guide: false,
```
shaohuzhang1, line 433: There are a few areas where you can optimize and address potential issues in the provided code:

1. **Language Selection Clearance**:
   - In the `<el-select>` component for selecting the language, consider adding `clearable` to make it easier for users to remove their selection if needed.

2. **Initial Language Setting**:
   - When setting the initial value of the `language` field in the forms (`defaultSetting`, `form`, and `xpackForm`), ensure that there's a meaningful default value (e.g., an empty string) rather than using `getBrowserLang()`. This avoids relying on environment settings and makes testing more straightforward.

3. **Code Duplication**: 
   - There appears to be some duplication between lines 51 and 67. Ensure these lines match or combine logically when necessary.

Here are specific changes:

### Revised Code Snippets

- **Line 50:
```javascript
                <div>
                  <!-- Existing content -->
                  <el-select v-model="form.language">
                    ...
```
- **Lines: 174 & 383**:
```javascript
            <h5 class="mb-8">{{ $t('views.applicationOverview.appInfo.SettingDisplayDialog.languageLabel') }}</h5>

...

const form = ref<any>({
...
  language: '', // Change here
});
```

If further modification or adjustments are required based on additional context not shown here, please let me know!

PR #2237, opened 2025-02-11
fix: Filter blank sheet

shaohuzhang1:
fix: Filter blank sheet  --bug=1052097 --user=刘瑞斌 【github#2196】【应用编排】应用对话的时候上传带空白sheet的表格会报错 https://www.tapd.cn/57709429/s/1653414

### DIFF
--- a/apps/common/handle/impl/table/xls_parse_table_handle.py
+++ b/apps/common/handle/impl/table/xls_parse_table_handle.py
@@ -68,6 +68,9 @@ class XlsSplitHandle(BaseParseTableHandle):
             sheets = workbook.sheets()
             md_tables = ''
             for sheet in sheets:
+                # 过滤空白的sheet
+                if sheet.nrows == 0 or sheet.ncols == 0:
+                    continue
 
                 # 获取表头和内容
                 headers = sheet.row_values(0)


apps/common/handle/impl/table/xls_parse_table_handle.py
```diff
@@ -68,6 +68,9 @@ def get_content(self, file, save_image):
             sheets = workbook.sheets()
             md_tables = ''
             for sheet in sheets:
+                # 过滤空白的sheet
+                if sheet.nrows == 0 or sheet.ncols == 0:
+                    continue
 
                 # 获取表头和内容
                 headers = sheet.row_values(0)
```
shaohuzhang1, line 76: Your code looks well-written and should work correctly. However, there is no need to iterate over blank sheets. A better approach would be to skip such sheets directly during the loop instead of after it.

Here's an optimized version:

```python
for sheet in sheets:
    # Get table headers and content only for non-blank sheets
    if sheet.nrows > 0 and sheet.ncols > 0:
        headers = sheet.row_values(0)
```

This way, you avoid iterating over and processing empty worksheets at all.

PR #2256, opened 2025-02-12
fix: Lost content during the answering process

shaohuzhang1:
fix: Lost content during the answering process

### DIFF
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -76,10 +76,12 @@ def event_content(response,
     all_text = ''
     reasoning_content = ''
     try:
+        response_reasoning_content = False
         for chunk in response:
             reasoning_chunk = reasoning.get_reasoning_content(chunk)
             content_chunk = reasoning_chunk.get('content')
             if 'reasoning_content' in chunk.additional_kwargs:
+                response_reasoning_content = True
                 reasoning_content_chunk = chunk.additional_kwargs.get('reasoning_content', '')
             else:
                 reasoning_content_chunk = reasoning_chunk.get('reasoning_content')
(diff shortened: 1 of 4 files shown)


apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
```diff
@@ -276,11 +293,13 @@ def execute_block(self, message_list: List[BaseMessage],
...
+            content = reasoning_result.get('content') + reasoning_result_end.get('content')
             if 'reasoning_content' in chat_result.response_metadata:
                 reasoning_content = chat_result.response_metadata.get('reasoning_content', '')
             else:
-                reasoning_content = reasoning_result.get('reasoning_content')
+                reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get(
+                    'reasoning_content')
             post_response_handler.handler(chat_id, chat_record_id, paragraph_list, problem_text,
                                           chat_result.content, manage, self, padding_problem_text, client_id,
                                           reasoning_content=reasoning_content if reasoning_content_enable else '')
```
shaohuzhang1, line 305: The code contains several issues:

1. `response_reasoning_content` variable is only initialized once per response loop. If multiple chunks have `reasoning_content`, it won't handle them correctly.

2. There might be inconsistencies between how AI-generated reasoning and end-stage reasoning are being handled. The logic to decide which one goes into the final output stream could be improved or removed if you want consistent behavior.

3. In the `execute_block` method, there's a potential mismatch at lines where end of session reasoning is appended to the main content string, leading to repeated text appearing twice.

4. Ensure that each node has unique metadata, especially for view types like "many_view". Having duplicates can confuse downstream systems trying to understand node relationships or views.

Optimization Suggestions:

1. Consider refactoring the logic related to determining and appending the end-of-session reasoning separately from normal inference processing to clean up the overall flow.

2. Double-check and possibly remove any redundant checks on whether a chunk has reasoning_content. This may simplify your conditional statements.

3. For consistency, either include all reasoning in every stream_chunk_response call, or only include what should appear after an end-of-sesion event depending on how your application processes user interactions.

Remember these changes will affect both runtime performance and readability; thoroughly test them afterward with comprehensive test cases.

apps/application/flow/tools.py
```diff
@@ -60,8 +71,7 @@ def get_reasoning_content(self, chunk):
             return {'content': chunk.content, 'reasoning_content': ''}
         # 是否包含结束
         if reasoning_content_end_tag_prefix_index > -1:
-            if len(
-                    self.reasoning_content_chunk) - reasoning_content_end_tag_prefix_index > self.reasoning_content_end_tag_len:
+            if len(self.reasoning_content_chunk) - reasoning_content_end_tag_prefix_index >= self.reasoning_content_end_tag_len:
                 reasoning_content_end_tag_index = self.reasoning_content_chunk.find(self.reasoning_content_end_tag)
                 if reasoning_content_end_tag_index > -1:
                     reasoning_content_chunk = self.reasoning_content_chunk[0:reasoning_content_end_tag_index]
```
shaohuzhang1, line 77: 1. The `__init__` method initializes several instance variables that do not appear to be used later in the class. Consider whether they can be removed.

2. The methods are named inconsistently; one uses snake_case (`get_end_reasoning_content`, `get_reasoning_content`) while others use camelCase or have mixed-case names (e.g., `all_content`). Consistency is important for readability and maintainability.

3. In both `__init__` and `get_reasoning_content`, you initialize `self.reasoning_content_chunk` directly within these methods without resetting it after operation completion, which might lead to unexpected behavior when other parts of the code modify this variable.

4. The logic inside `calculate_and_store_result`, especially regarding handling empty strings and finding indices in `reasoning_content_chunk`, seems complicated and could simplify significantly. 

5. While the code is clean overall, using meaningful comments and adding type hints would improve understanding and make the refactoring process smoother if further changes were needed.

PR #2258, opened 2025-02-12
fix: Display setting bugs

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/application-overview/component/DisplaySettingDialog.vue
+++ b/ui/src/views/application-overview/component/DisplaySettingDialog.vue
@@ -59,7 +59,8 @@ const emit = defineEmits(['refresh'])
 
 const displayFormRef = ref()
 const form = ref<any>({
-  show_source: false
+  show_source: false,
+  language: '',
 })
 
 const detail = ref<any>(null)
@@ -70,18 +71,15 @@ const loading = ref(false)
 watch(dialogVisible, (bool) => {
   if (!bool) {
     form.value = {
-      show_source: false
+      show_source: false,
+      language: '',
     }
   }
 })
 const open = (data: any, content: any) => {
   detail.value = content
   form.value.show_source = data.show_source
-  form.value.language = data.language
-  if (!form.value.language) {
-    form.value.language = getBrowserLang()
-  }
-
+  form.value.language = data.language || getBrowserLang()
   dialogVisible.value = true
 }
 
--- a/ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
+++ b/ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
@@ -146,16 +146,16 @@
         </div>
       </div>
 
-      <el-form ref="displayFormRef" :model="form">
+      <el-form ref="displayFormRef" :model="xpackForm">
         <el-row class="w-full mb-8">
           <el-col :span="12">
             <h5 class="mb-8">
               {{ $t('views.applicationOverview.appInfo.SettingDisplayDialog.customThemeColor') }}
             </h5>
             <div>
-              <el-color-picker v-model="form.custom_theme.theme_color" />
+              <el-color-picker v-model="xpackForm.custom_theme.theme_color" />
               {{
-                !form.custom_theme.theme_color
+                !xpackForm.custom_theme.theme_color
                   ? $t('views.applicationOverview.appInfo.SettingDisplayDialog.default')
                   : ''
               }}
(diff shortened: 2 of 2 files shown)


ui/src/views/application-overview/component/XPackDisplaySettingDialog.vue
```diff
@@ -520,7 +515,6 @@ const open = (data: any, content: any) => {
     header_font_color: data.custom_theme?.header_font_color || '#1f2329'
   }
   xpackForm.value.float_location = data.float_location
-  form.value = xpackForm.value
   dialogVisible.value = true
 }
```
shaohuzhang1, line 520: The code looks generally correct but has some areas that could be optimized or improved. Here are a few suggestions:

### Code Areas with Issues

1. **Duplicate Data Handling**: The `form` and `xpackForm` objects use the same structure (`show_source`, etc.), but they should be separated to avoid confusion and potential bugs. Consider removing one of them.

2. **Missing Return Statements in Conditional Rendering**: In the template section, there's an unhandled case where the disclaimer might not be set, which can cause undefined behavior.

3. **Unnecessary Cloning**: While cloning is necessary for initial setup, it might be redundant within the lifecycle hooks (`open` method). Avoid unnecessary clones unless absolutely necessary.

4. **Comments Quality**: Some comments, such as those inside tooltips or input fields, could be more detailed for better clarity when reading the code later.

### Optimizations/Suggestions

1. **Separate Configuration Objects**: Refactor the `xpackForm` object to separate different configurations like themes, localization, position controls, and visibility settings.

2. **Improve Exception Handling**: For cases where `disclaimer_value` might not be set, consider adding error handling or default values to prevent runtime errors.

3. **Cleaner Tooltip Content Logic**: Simplify how tooltip content is generated using computed properties or computed ternary expressions for cleaner code.

Here’s an updated version of the relevant parts based on these suggestions:

```vue
<template>
  <!-- ... existing elements ... -->
  
  <section>
    <h5>{{ $t('views.applicationOverview.appInfo.SettingDisplayDialog.languageLabel') }}</h5>
    <el-select v-model="lang" clearable @change="handleLanguageChange">
      <el-option
        v-for="item in langList"
        :key="item.value"
        :label="item.label"
        :value="item.value"
      >
      </el-option>
    </el-select>

    
     <!-- Existing sections with logic -->
   
</template>

<script setup lang="ts">
import { ref } from 'vue';
// Import other dependencies here...

interface ThemeConfig {
  theme_color: string;
  header_font_color: string;
}

interface DisplaySettings {
  show_execution_detail: boolean;
  show_version_description: boolean;
  show_source: boolean;
  show_history: boolean;
  show_guide: boolean;
  disclaimer: boolean;
  disclaimer_value?: string;
  float_location?: {
    x: LocationConfig;
    y: LocationConfig;
  };
}

interface LocationConfig {
  type: string;
  value: number | string;
}

const displaySettings = ref<DisplaySettings>({
  show_execution_detail: true,
  show_version_description: true,
  show source: false,
  show_history: true,
  show_guide: true,
  disclaimer: false,
});

const themeConfig = ref<ThemeConfig>({
  theme_color: '#FFFFFF',
  header_font_color: '#000000',
});

function handleLanguageChange(value) {
  // Handle language change logic...
}
</script>
```

This approach separates configuration into manageable components, makes sure exceptions are handled appropriately, and simplifies conditional rendering through Vue's reactive system.

ui/src/views/application-overview/component/DisplaySettingDialog.vue
```diff
@@ -70,18 +71,15 @@ const loading = ref(false)
...
   detail.value = content
   form.value.show_source = data.show_source
-  form.value.language = data.language
-  if (!form.value.language) {
-    form.value.language = getBrowserLang()
-  }
-
+  form.value.language = data.language || getBrowserLang()
   dialogVisible.value = true
 }
```
shaohuzhang1, line 85: There appear to be no apparent technical issues with the provided code. However, there are a few areas where improvements can make the logic more robust and user-friendly:

1. **Language Defaulting**: The default case for `language` being set to `getBrowserLang()` is unnecessary when you have an initial value defined (`'': ''`). It's better to remove this redundant line since it defaults `language` immediately.

2. **Code Formatting**: Although not required, formatting the code consistently could improve readability. Use consistent spacing and indentation patterns.

3. **Variable Initialization**: Initializing `dialogVisible` to `false`, even though its setter function resets it to `true`. This seems superfluous unless there's specific behavior expected that would change depending on previous values.

4. **Comments Clarity**: While current comments are clear, they might benefit from minor clarification or removal if the intent is straightforward.

Here's a revised version of the code incorporating these suggestions:

```javascript
const emit = defineEmits(['refresh'])

const displayFormRef = ref()
const form = ref({
  show_source: false,
  language: '', // Initial/default value
})

const detail = ref(null)
const loading = ref(false)

watch(dialogVisible, (bool) => {
  if (!bool) {
    resetFormValues();
  }
});

function resetFormValues() {
  form.value.show_source = false;
  form.value.language = '';
}

const open = (data: any, content: any) => {
   detail.value = content;
   form.value.show_source = data.show_source;

   // Default language if none given
   form.value.language ||= getBrowserLang();

   dialogVisible.value = true;
}
```

### Key Changes:
- Removed redundant setting of `language`.
- Refactored variable initialization inside the watcher callback.
- Added `resetFormValues` helper function to reduce redundancy in handling form reset.
- Used logical OR assignment operator `(+=)` or short-circuit evaluation `||` for default values.

PR #2259, opened 2025-02-12
fix: Mac upload folder error

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/dataset/component/UploadComponent.vue
+++ b/ui/src/views/dataset/component/UploadComponent.vue
@@ -253,8 +253,11 @@ const fileHandleChange = (file: any, fileList: UploadFiles) => {
     fileList.splice(-1, 1) //移除当前超出大小的文件
     return false
   }
+
   if (!isRightType(file?.name, form.value.fileType)) {
-    MsgError(t('views.document.upload.errorMessage2'))
+    if (file?.name !== '.DS_Store') {
+      MsgError(t('views.document.upload.errorMessage2'))
+    }
     fileList.splice(-1, 1)
     return false
   }


ui/src/views/dataset/component/UploadComponent.vue
```diff
@@ -253,8 +253,11 @@ const fileHandleChange = (file: any, fileList: UploadFiles) => {
...
   }
+
   if (!isRightType(file?.name, form.value.fileType)) {
-    MsgError(t('views.document.upload.errorMessage2'))
+    if (file?.name !== '.DS_Store') {
+      MsgError(t('views.document.upload.errorMessage2'))
+    }
     fileList.splice(-1, 1)
     return false
   }
```
shaohuzhang1, line 263: Here are some suggested improvements to the code:

1. **Remove Redundant Line**: The line `fileList.splice(-1, 1)` can be removed since it is only executed when an error occurs and will not change the final state of the `fileList`.

2. **Use Conditional Statements Consistently**: Ensure that all conditions follow a consistent structure.

3. **Move `MsgError` Call Outside Loop**: If there might be multiple errors per upload attempt, consider moving the `MsgError` call outside the loop to avoid unnecessary processing for valid files.

4. **Check for `.DS_Store` File Explicitly**: The condition `if !file?.name !== '.DS_Store'` looks redundant because you already check if the file extension is correct. Consider removing this part if appropriate.

Here's refactored version of the code with these improvements applied:

```javascript
const fileHandleChange = (file: any, fileList: UploadFiles) => {
  // Check file type consistency
  console.log(form.value.fileType); // Debug statement to ensure form value is correct
  if (!isRightType(file?.name, form.value.fileType)) {
    MsgError(t('views.document.upload.errorMessage2'));
    return false;
  }

  // Remove current file if its size exceeds allowed limit
  if (file.size > MAX_SIZE) {
    fileList.splice(-1, 1);
  }

  return true; // Return true to indicate successful handling of the file
}
```

These changes streamline the function logic and improve readability while addressing potential issues.

PR #2260, opened 2025-02-12
fix: No error message prompted after QR code failure

shaohuzhang1:
fix: No error message prompted after QR code failure  --bug=1052150 --user=王孝刚 【系统设置】钉钉二维码失效后，点击刷新，使用非当前组织的用户扫码登录，没有弹错误提示 https://www.tapd.cn/57709429/s/1653901

### DIFF
--- a/ui/src/views/login/components/dingtalkQrCode.vue
+++ b/ui/src/views/login/components/dingtalkQrCode.vue
@@ -80,6 +80,7 @@ const errorShown = ref(false)
 const initActive = async () => {
   try {
     await load(true)
+    errorShown.value = false
     if (!isConfigReady.value) {
       return
     }
@@ -116,7 +117,7 @@ const initActive = async () => {
       (errorMsg: string) => {
         if (!errorShown.value) {
           MsgError(errorMsg)
-          errorShown.value = true // 设置标志位为 true，表示错误已经显示过
+          errorShown.value = true
         }
       }
     )


ui/src/views/login/components/dingtalkQrCode.vue
```diff
@@ -116,7 +117,7 @@ const initActive = async () => {
       (errorMsg: string) => {
         if (!errorShown.value) {
           MsgError(errorMsg)
-          errorShown.value = true // 设置标志位为 true，表示错误已经显示过
+          errorShown.value = true
         }
       }
     )
```
shaohuzhang1, line 123: The provided code is generally well-written, but there are a few minor improvements and clarifications you can make:

1. **Typo**: There seems to be an issue with the message being passed to `MsgError`. It's likely that it should not include the `{ errorMsg }` syntax within single quotes.

2. **Redundancy**: The line where `errorShown.value = true;` is followed by `(errorMsg: string) => { ... };`, which does not serve any purpose without more context about what this function does.

3. **Whitespace**: Ensure consistent indentation throughout the codebase for better readability.

Here is the revised version of the code:

```javascript
@@ -80,6 +80,7 @@ const errorShown = ref(false)
 const initActive = async () => {
   try {
     await load(true);
+    errorShown.value = false;
     if (!isConfigReady.value) {
       return;
     }
@@ -114,7 +115,7 @@ const initActive = async () => {
       (errorMsg: string) => {
         if (!errorShown.value) {
           MsgError(errorMsg); // Pass the error message directly
+          errorShown.value = true; // Set flag to indicate error has been shown
         }
       },
     );
```

### Additional Comments:
- **Error Handling Flag**: The `errorShown.value = true;` statement inside the callback ensures that each error message is only displayed once during its execution lifetime even if multiple error messages occur.
  
These changes improve clarity and maintainability of the code while ensuring it performs correctly.

PR #2263, opened 2025-02-13
fix: application set language error

shaohuzhang1:
fix: application set language error

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -374,6 +374,8 @@ class ApplicationSerializer(serializers.Serializer):
                 application_access_token.show_source = instance.get('show_source')
             if 'language' in instance and instance.get('language') is not None:
                 application_access_token.language = instance.get('language')
+            if 'language' not in instance or instance.get('language') is None:
+                application_access_token.language = None
             application_access_token.save()
             application_setting_model = DBModelManage.get_model('application_setting')
             xpack_cache = DBModelManage.get_model('xpack_cache')
--- a/ui/src/views/application-overview/component/DisplaySettingDialog.vue
+++ b/ui/src/views/application-overview/component/DisplaySettingDialog.vue
@@ -60,7 +60,7 @@ const emit = defineEmits(['refresh'])
 const displayFormRef = ref()
 const form = ref<any>({
   show_source: false,
-  language: '',
+  language: ''
 })
 
 const detail = ref<any>(null)
@@ -72,14 +72,14 @@ watch(dialogVisible, (bool) => {
   if (!bool) {
     form.value = {
       show_source: false,
-      language: '',
+      language: ''
     }
   }
 })
 const open = (data: any, content: any) => {
   detail.value = content
   form.value.show_source = data.show_source
-  form.value.language = data.language || getBrowserLang()
+  form.value.language = data.language
   dialogVisible.value = true
 }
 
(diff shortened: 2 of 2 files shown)


ui/src/views/application-overview/component/DisplaySettingDialog.vue
```diff
@@ -72,26 +72,22 @@ watch(dialogVisible, (bool) => {
...
     if (valid) {
-      const obj = {
-        show_source: form.value.show_source,
-        language: form.value.language
-      }
-      applicationApi.putAccessToken(id as string, obj, loading).then((res) => {
+      applicationApi.putAccessToken(id as string, form.value, loading).then((res) => {
         emit('refresh')
         // @ts-ignore
         MsgSuccess(t('common.settingSuccess'))
```
shaohuzhang1, line 93: The code provided is mostly correct, but there are a few areas that can be optimized or improved:

1. **Empty Language Check**:
   ```javascript
-      form.value.language || getBrowserLang()
+      form.value.language
```
   The `||` operator seems unnecessary here since it might lead to unexpected behavior if the language is indeed an empty string. It's generally better to directly use the value.

2. **Avoiding Repeated Function Calls**:
   - In the `open` function, you retrieve `id` twice (`data.id` and `content?.id`). You could store this once and reuse it.
   - Ensure `applicationApi.putAccessToken` is only called if `formEl` is not null.

Here’s the revised version of the relevant parts:

```javascript
// Avoid redundant 'data' retrieval
const { id } = data;

const open = async (info: any, content: any) => {
  let currentLanguage;
  
  if (info.language) {
    currentLanguage = info.language;
  } else {
    currentLanguage = getBrowserLang();
  }

  detail.value = content;
  form.value.show_source = info.show_source;
  form.value.language = currentLanguage; 

  dialogVisible.value = true;
}

  if (!formEl) return;
  await formEl.validate((valid, fields) => {
    if (valid && formEl.instance != null) {
      const obj = {
        show_source: form.value.show_source,
        language: form.value.language
      };

      try {
        await applicationApi.putAccessToken(id as string, obj, loading);
        emit('refresh');
          MSG_Success(t('common.settingSuccess'));
      } catch (error) {
        console.error("Failed to update access token:", error.message);
        alert("An error occurred while updating your settings.");
      }
    }
  });
};
```

### Key Changes:
- Removed the conditional check for `language` in `open`.
- Stored the determined language to avoid repeated calls to `get BrowserLang()`.
- Provided feedback on failed API calls by logging errors to the console and showing a user-friendly message.

apps/application/serializers/application_serializers.py
```diff
@@ -374,6 +374,8 @@ def edit(self, instance: Dict, with_valid=True):
                 application_access_token.show_source = instance.get('show_source')
             if 'language' in instance and instance.get('language') is not None:
                 application_access_token.language = instance.get('language')
+            if 'language' not in instance or instance.get('language') is None:
+                application_access_token.language = None
             application_access_token.save()
             application_setting_model = DBModelManage.get_model('application_setting')
             xpack_cache = DBModelManage.get_model('xpack_cache')
```
shaohuzhang1, line 381: The code looks mostly fine, but there's an unnecessary `if` condition. You can remove it since setting the language to `None` when the key isn't present in the input dictionary is already handled in the original `if` statement.

Here’s the refined version:

```python
@@ -373,8 +373,6 @@ def edit(self, instance: Dict, with_valid=True):
         application_access_token.show_source = instance.get('show_source')
     if 'language' in instance and instance.get('language') is not None:
         application_access_token.language = instance.get('language')
 
     # Removed this redundant check
     # if 'language' not in instance or instance.get('language') is None:
     #     application_access_token.language = None  # This line was duplicated
 
     application_access_token.save()
     application_setting_model = DBModelManage.get_model('application_setting')
     xpack_cache = DBModelManage.get_model('xpack_cache')
```

This eliminates redundancy while maintaining the intended logic of handling the case where "language" might be missing from the input dictionary.

PR #2264, opened 2025-02-13
perf: Add custom token_ids method

shaohuzhang1:
perf: Add custom token_ids method

### DIFF
--- a/apps/setting/models_provider/impl/base_chat_open_ai.py
+++ b/apps/setting/models_provider/impl/base_chat_open_ai.py
@@ -15,8 +15,14 @@ from langchain_openai.chat_models.base import _convert_chunk_to_generation_chunk
 from common.config.tokenizer_manage_config import TokenizerManage
 
 
+def custom_get_token_ids(text: str):
+    tokenizer = TokenizerManage.get_tokenizer()
+    return tokenizer.encode(text)
+
+
 class BaseChatOpenAI(ChatOpenAI):
     usage_metadata: dict = {}
+    custom_get_token_ids = custom_get_token_ids
 
     def get_last_generation_info(self) -> Optional[Dict[str, Any]]:
         return self.usage_metadata


apps/setting/models_provider/impl/base_chat_open_ai.py
```diff
@@ -15,8 +15,14 @@
...
+    tokenizer = TokenizerManage.get_tokenizer()
+    return tokenizer.encode(text)
+
+
 class BaseChatOpenAI(ChatOpenAI):
     usage_metadata: dict = {}
+    custom_get_token_ids = custom_get_token_ids
 
     def get_last_generation_info(self) -> Optional[Dict[str, Any]]:
         return self.usage_metadata
```
shaohuzhang1, line 28: There are several points to address:

1. **Function Overloading**: In Python, function overloading is not directly supported like some other languages. However, you can achieve similar functionality using default arguments or decorators.

2. **Custom Function Placement**: The placement of `custom_get_token_ids` inside the class body makes it difficult to understand its logic and potential side effects. Consider moving the entire implementation of this method into a separate file if needed.

3. **Class Method Access**: Direct access to instance methods from the base class (`TokenizerManage`) might violate encapsulation principles unless specifically allowed by design. If possible, consider making these methods accessible via instance attributes.

4. **Docstring Consistency**: Ensure consistent use of docstrings across the codebase. It's important for others (or future you) to understand what each method does.

5. **Usage Metadata Handling**: The presence of `usage_metadata` suggests that the class might be used in contexts where request tracking is required. Verify there aren't any unhandled cases where requests don’t update or clear this metadata, which could lead to incorrect metrics or inconsistencies.

6. **Deprecation warnings**: Using deprecated functions or modules should be discouraged since they may not receive updates or support in the future. Ensure all dependencies are up-to-date and compatible with your project's needs.

7. **Testing Strategy**: Implement comprehensive tests, particularly unit tests, to cover different scenarios and edge cases, including how `custom_get_token_ids` interacts with dependent methods.

Based on these considerations, here’s an improved version of your code:

```python
from typing import Optional, Dict, Any

class BaseChatOpenAI:
    usage_metadata: dict = {}
    
    def __init__(self, custom_get_token_ids=None):
        self.custom_get_token_ids = custom_get_token_ids
    
    @classmethod
    def set_custom_get_token_ids(cls, func):
        cls.custom_get_token_ids = staticmethod(func)

    def get_last_generation_info(self) -> Optional[Dict[str, Any]]:
        return self.usage_metadata


# Example usage to register a custom function
@BaseChatOpenAI.set_custom_get_token_ids
def custom_get_token_ids(text: str):
    # Implementation of token encoding
    pass

# Instantiate the class with the custom get_token_ids function
chat_openai_instance = BaseChatOpenAI(custom_get_token_ids)
```

This refactored code ensures proper separation of concerns and adheres to Python naming conventions, while also providing guidance on best practices for maintaining robust and modular software.

Issue #2126, opened 2025-02-05
[Feature] 建议模型设置中的搜索功能支持忽略大小写


xusongtao:
### MaxKB Version

v1.9.1

### Please describe your needs or suggestions for improvements

模型设置中的搜索功能支持忽略大小写

<img width="896" alt="Image" src="https://github.com/user-attachments/assets/ee39b5c8-07f5-4485-9677-72bbbab90181" />

<img width="905" alt="Image" src="https://github.com/user-attachments/assets/69c86953-9369-4cdc-ad0e-d9df1836167b" />

### Please describe the solution you suggest

模型设置中的搜索功能支持忽略大小写

baixin513:
感谢反馈，后续版本迭代

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version iterations

baixin513:
v1.10.2已发布。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.2 has been released.

PR #2279, opened 2025-02-14
refactor: Search ignores capitalization

shaohuzhang1:
refactor: Search ignores capitalization  --story=1017866 --user=王孝刚 【Bug转需求】【搜索】-函数库、模型、团队成员界面搜索不支持忽略大小写 #2126 https://www.tapd.cn/57709429/s/1654282

### DIFF
--- a/apps/function_lib/serializers/function_lib_serializer.py
+++ b/apps/function_lib/serializers/function_lib_serializer.py
@@ -105,7 +105,7 @@ class FunctionLibSerializer(serializers.Serializer):
             query_set = QuerySet(FunctionLib).filter(
                 (Q(user_id=self.data.get('user_id')) | Q(permission_type='PUBLIC')))
             if self.data.get('name') is not None:
-                query_set = query_set.filter(name__contains=self.data.get('name'))
+                query_set = query_set.filter(name__icontains=self.data.get('name'))
             if self.data.get('desc') is not None:
                 query_set = query_set.filter(desc__contains=self.data.get('desc'))
             if self.data.get('is_active') is not None:
--- a/apps/setting/serializers/provider_serializers.py
+++ b/apps/setting/serializers/provider_serializers.py
@@ -106,7 +106,7 @@ class ModelSerializer(serializers.Serializer):
                 model_query_set = QuerySet(Model).filter((Q(user_id=user_id) | Q(permission_type='PUBLIC')))
             query_params = {}
             if name is not None:
-                query_params['name__contains'] = name
+                query_params['name__icontains'] = name
             if self.data.get('model_type') is not None:
                 query_params['model_type'] = self.data.get('model_type')
             if self.data.get('model_name') is not None:
--- a/ui/src/views/team/component/PermissionSetting.vue
+++ b/ui/src/views/team/component/PermissionSetting.vue
@@ -57,7 +57,12 @@
           />
         </template>
       </el-table-column>
-      <el-table-column :label="$t('views.team.setting.check')"  align="center" width="100" fixed="right">
+      <el-table-column
+        :label="$t('views.team.setting.check')"
+        align="center"
+        width="100"
+        fixed="right"
+      >
         <template #header>
           <el-checkbox
             :disabled="props.manage"
@@ -135,7 +140,9 @@ const allChecked: any = ref({
 
 const filterText = ref('')
 
-const filterData = computed(() => props.data.filter((v: any) => v.name.includes(filterText.value)))
+const filterData = computed(() =>
+  props.data.filter((v: any) => v.name.toLowerCase().includes(filterText.value.toLowerCase()))
+)
 
 const allIndeterminate: any = ref({
   [TeamEnum.MANAGE]: computed(() => {


ui/src/views/team/component/PermissionSetting.vue
```diff
@@ -135,7 +140,9 @@ const allChecked: any = ref({
 
 const filterText = ref('')
 
-const filterData = computed(() => props.data.filter((v: any) => v.name.includes(filterText.value)))
+const filterData = computed(() =>
+  props.data.filter((v: any) => v.name.toLowerCase().includes(filterText.value.toLowerCase()))
+)
 
 const allIndeterminate: any = ref({
   [TeamEnum.MANAGE]: computed(() => {
```
shaohuzhang1, line 148: The provided code looks mostly correct and follows best practices. However, there are a couple of suggestions to optimize or improve it:

1. **Case Insensitive Search**: The current filter function uses `includes`, which is case-sensitive. If you want the search to be case-insensitive, use `toLowerCase()` on both `filterText` and each element's `name`. This ensures that searches are not affected by capitalization.

2. **Use of Computed Values**: The `allIndeterminate` object keys should be properly formatted (i.e., starting with an uppercase letter). In JavaScript/TypeScript, variable names must start with a letter or underscore. It's also generally good practice to make `allChecked` into a reactive object instead of just using a plain reference.

Here's the revised code with these considerations addressed:

```typescript
<template>
  <!-- Existing template -->
</template>

<script setup lang="ts">
import { TeamEnum } from './teamConstants'; // Assuming this file exists

const props = defineProps({
  data: Array<any>,
  manage: Boolean,
  selectAllAction: Function,
});

const allChecked: Record<string, boolean> = reactronRef({ [TeamEnum.MANAGE]: true }, false);

const filterText = ref('');

// Case insensitive search
const filterData = computed(() =>
  props.data.filter(item => item.name.toLowerCase().includes(filterText.value.toLowerCase()))
);

const allIndeterminate: Record<string, boolean> = {
  [TeamEnum.MANAGE]: computed(() => Object.values(allChecked).some(val => val)),
};

// Emit action when selecting/deselecting rows
function handleSelect(row: any, checked: boolean) {
  if (!row.manage) return;
  selectedValues[row.id] = checked;
  allChecked[TeamEnum.MANAGE] = Object.values(selectedValues).every(value => value);
}

onMounted(async () => {
  // Load initial state
});
</script>

<style scoped>
/* Existing styles */
</style>
```

**Explanation of Changes**:
- Added `reactronRef` usage for `allChecked` to create a reactive object.
- Updated the logic inside the `handleSelect` function to correctly update nested properties in `selectedValues`.
- Made `allIndeterminate` a more typical way to track indeterminate states in Vue.
- Corrected the formatting of property keys in `allIndeterminate`.

These changes ensure the code remains efficient, maintainable, and adheres to standard JavaScript/TypeScript conventions.

apps/function_lib/serializers/function_lib_serializer.py
```diff
@@ -105,7 +105,7 @@ def get_query_set(self):
             query_set = QuerySet(FunctionLib).filter(
                 (Q(user_id=self.data.get('user_id')) | Q(permission_type='PUBLIC')))
             if self.data.get('name') is not None:
-                query_set = query_set.filter(name__contains=self.data.get('name'))
+                query_set = query_set.filter(name__icontains=self.data.get('name'))
             if self.data.get('desc') is not None:
                 query_set = query_set.filter(desc__contains=self.data.get('desc'))
             if self.data.get('is_active') is not None:
```
shaohuzhang1, line 111: There are no significant issues with this code snippet. The `name__icontains` filter should be used instead of `name__contains` to ensure that the search matches names which contain the provided substring, regardless of case sensitivity. Here's the corrected line:

```python
query_set = query_set.filter(name__icontains=self.data.get('name'))
```

Additionally, it would be beneficial to add error handling around the data extraction from `self.data`. For example, you can check if the key exists before accessing its value:

```python
if 'user_id' in self.data:
    user_id = self.data['user_id']
else:
    raise ValueError("User ID is required.")

if 'permission_type' in self.data:
    permission_type = self.data['permission_type']
else:
    permission_type = 'PUBLIC'
```

These changes will help prevent potential errors when attempting to access non-existent keys in `self.data`.

apps/setting/serializers/provider_serializers.py
```diff
@@ -106,7 +106,7 @@ def list(self, with_valid):
                 model_query_set = QuerySet(Model).filter((Q(user_id=user_id) | Q(permission_type='PUBLIC')))
             query_params = {}
             if name is not None:
-                query_params['name__contains'] = name
+                query_params['name__icontains'] = name
             if self.data.get('model_type') is not None:
                 query_params['model_type'] = self.data.get('model_type')
             if self.data.get('model_name') is not None:
```
shaohuzhang1, line 112: The provided Python function has the following improvements:

1. Use `self.data.get('model_type')` instead of directly accessing it without a check. This prevents errors if 'model_type' is missing and avoids using an undefined variable.

2. Update the field lookup to use `_icontains` which means partial case-insensitive matching in Django's ORM, while `__contains` would perform exact string matching.

3. Remove unnecessary brackets around conditions (`Q(user_id=user_id) | Q(permission_type='PUBLIC'))`, since this can be simplified as shown above.

This changes will make the query more robust by handling optional parameters safely and improving performance when dealing with potentially large dataset queries.

PR #2300, opened 2025-02-17
refactor: Support image_list, document_list while using swagger api

shaohuzhang1:
refactor: Support image_list, document_list while using swagger api  --story=1017867 --user=刘瑞斌 对话API支持上传文件、语音、图像和视频，用于实现maxkb的文件解析及多模态对话#2228 https://www.tapd.cn/57709429/s/1654842

### DIFF
--- a/apps/dataset/serializers/file_serializers.py
+++ b/apps/dataset/serializers/file_serializers.py
@@ -62,7 +62,7 @@ class FileSerializer(serializers.Serializer):
     def upload(self, with_valid=True):
         if with_valid:
             self.is_valid(raise_exception=True)
-        meta = self.data.get('meta')
+        meta = self.data.get('meta', {'debug': True})
         file_id = meta.get('file_id', uuid.uuid1())
         file = File(id=file_id, file_name=self.data.get('file').name, meta=meta)
         file.save(self.data.get('file').read())
--- a/apps/application/serializers/chat_message_serializers.py
+++ b/apps/application/serializers/chat_message_serializers.py
@@ -222,13 +222,19 @@ class OpenAIChatSerializer(serializers.Serializer):
         client_type = self.data.get('client_type')
         chat_id = self.generate_chat(chat_id, application_id, message, client_id)
         return ChatMessageSerializer(
-            data={'chat_id': chat_id, 'message': message,
-                  're_chat': re_chat,
-                  'stream': stream,
-                  'application_id': application_id,
-                  'client_id': client_id,
-                  'client_type': client_type, 'form_data': instance.get('form_data', {})}).chat(
-            base_to_response=OpenaiToResponse())
+            data={
+                'chat_id': chat_id, 'message': message,
+                're_chat': re_chat,
+                'stream': stream,
+                'application_id': application_id,
+                'client_id': client_id,
+                'client_type': client_type,
+                'form_data': instance.get('form_data', {}),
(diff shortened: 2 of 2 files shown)


apps/dataset/serializers/file_serializers.py
```diff
@@ -62,7 +62,7 @@ class FileSerializer(serializers.Serializer):
     def upload(self, with_valid=True):
         if with_valid:
             self.is_valid(raise_exception=True)
-        meta = self.data.get('meta')
+        meta = self.data.get('meta', {})
         file_id = meta.get('file_id', uuid.uuid1())
         file = File(id=file_id, file_name=self.data.get('file').name, meta=meta)
         file.save(self.data.get('file').read())
```
shaohuzhang1, line 68: The code looks generally correct but has a few minor points to consider:

1. **Handling of `meta` Dictionary**: The current implementation assumes that the 'meta' key exists in the data dictionary, which is safe if it's always present according to your use case. If there's any chance that this might not be true, you should add checks to handle such scenarios gracefully (e.g., using try-except blocks).

   ```python
   meta = self.data.get('meta', {})
   ```

2. **Use of `uuid3` Instead of `uuid1`**: While both UUID versions are valid, `uuid3` uses a namespace, which can make them easier to read compared to `uuid4`. However, the choice between `uuid5`, `uuid`, or `uuid4` would depend on what specific properties you want when generating unique IDs.

   ```python
   file_id = meta.get('file_id', uuid.uuid3(uuid.NAMESPACE_DNS, str(file)))
   # Or simply: file_id = meta.get('file_id', uuid.uuid4())  depending on needs
   ```

These changes ensure robustness against missing keys and clarity in how ID generation works.

apps/dataset/serializers/file_serializers.py
```diff
@@ -62,7 +62,7 @@ class FileSerializer(serializers.Serializer):
     def upload(self, with_valid=True):
         if with_valid:
             self.is_valid(raise_exception=True)
-        meta = self.data.get('meta')
+        meta = self.data.get('meta', {'debug': True})
         file_id = meta.get('file_id', uuid.uuid1())
         file = File(id=file_id, file_name=self.data.get('file').name, meta=meta)
         file.save(self.data.get('file').read())
```
shaohuzhang1, line 68: The code appears to be mostly correct, but there are a few improvements or notes that can be made:

1. **Handling `None` Values**: Currently, `self.data.get('meta')` will raise a KeyError if 'meta' is not present in the request data. You might want to add a check to ensure it exists before accessing its properties.

   ```python
   meta = self.data.get('meta', {})
   ```

2. **Defaulting Debug Mode**: The default value for `'debug'` is set to `True`. If you want this behavior only under certain conditions (e.g., testing/debugging), consider adding another condition to check the debug mode setting.

3. **File Handling**: Ensure that any exceptions related to reading from the file are properly handled. Using context managers (`with open(...) as f:`) would also improve error handling.

4. **Logging**: For debugging purposes, consider logging some key information during the process of uploading the file, such as the file ID and whether the validation was successful.

5. **Documentation**: Although not shown here, make sure to document all functions and methods appropriately, especially since they modify sensitive user files.

6. **Validation**: Depending on your specific requirements, ensure that after retrieving `meta`, further validation logic is applied if necessary.

By making these minor enhancements, you can make the code more robust, maintainable, and future-proof.

PR #2310, opened 2025-02-18
feat: Generate problem support for title variable

shaohuzhang1:
feat: Generate problem support for title variable

### DIFF
--- a/apps/dataset/task/generate.py
+++ b/apps/dataset/task/generate.py
@@ -29,7 +29,8 @@ def generate_problem_by_paragraph(paragraph, llm_model, prompt):
     try:
         ListenerManagement.update_status(QuerySet(Paragraph).filter(id=paragraph.id), TaskType.GENERATE_PROBLEM,
                                          State.STARTED)
-        res = llm_model.invoke([HumanMessage(content=prompt.replace('{data}', paragraph.content))])
+        res = llm_model.invoke(
+            [HumanMessage(content=prompt.replace('{data}', paragraph.content).replace('{title}', paragraph.title))])
         if (res.content is None) or (len(res.content) == 0):
             return
         problems = res.content.split('\n')


apps/dataset/task/generate.py
```diff
@@ -29,7 +29,8 @@ def generate_problem_by_paragraph(paragraph, llm_model, prompt):
     try:
         ListenerManagement.update_status(QuerySet(Paragraph).filter(id=paragraph.id), TaskType.GENERATE_PROBLEM,
                                          State.STARTED)
-        res = llm_model.invoke([HumanMessage(content=prompt.replace('{data}', paragraph.content))])
+        res = llm_model.invoke(
+            [HumanMessage(content=prompt.replace('{data}', paragraph.content).replace('{title}', paragraph.title))])
         if (res.content is None) or (len(res.content) == 0):
             return
         problems = res.content.split('\n')
```
shaohuzhang1, line 36: The given code has two main issues:

1. **Multiple `replace` functions**: The original `prompt.replace('{data}', paragraph.content)` replaces all occurrences of the placeholder `{data}` with the paragraph's content. However, if the input `prompt` also contains a placeholder for `'{title}'`, this line will replace both placeholders unnecessarily.

   ```python
   # Incorrect usage due to double replacement
   Prompt = "I want you to generate problems based on {data} ({title})."
   Paragraph_content = "Sample Content"
   Title = "Sample Title"
   
   prompt_with_title = Prompt.replace('{data}', Paragraph_content)  # This replaces '{data}'
   final_prompt = prompt_with_title.replace('{title}', Title)       # This replaces '{title}' again
   ```

   Instead, separate replacements should be done to avoid unnecessary replacements and make it clearer that each variable serves a specific purpose when generating text.

2. **String formatting in Python**: If there is an intention to format strings using variables from `paragraph.title`, consider using named parameters instead of keyword arguments when creating objects like `HumanMessage`. In many languages like Python and some others, using named parameters can help improve readability by clearly indicating which part of the string corresponds to which parameter value.

   ```python
   HumanMessage(content=prompt.replace('{data}', paragraph.content))
   ```
   
   becomes:

   ```python
   HumanMessage(content='{originalPrompt}'.format(originalPrompt=prompt.format(data=paragraph.content)))
   ```

Here’s how these improvements could look after addressing the first issue alone:

**Modified Code Block**

```python
try:
    ListenerManagement.update_status(QuerySet(Paragraph).filter(id=paragraph.id), TaskType.GENERATE_PROBLEM,
                                     State.STARTED)
    
    prompts_list = [
        {"message": prompt.replace('{data}', paragraph.content)},
        {"message": prompt.replace('{title}', paragraph.title)}
    ]

    for prompt_dict in prompts_list:
        res = llm_model.invoke([HumanMessage(**prompt_dict)])
    
        if (res.content is None) or (len(res.content) == 0):
                return
    
        problems = res.content.split('\n')
```

These changes ensure that each message generated uses only relevant placeholders defined in its respective dictionary key.

PR #2311, opened 2025-02-18
feat: The discriminator supports variables

shaohuzhang1:
feat: The discriminator supports variables

### DIFF
--- a/apps/application/flow/step_node/condition_node/impl/base_condition_node.py
+++ b/apps/application/flow/step_node/condition_node/impl/base_condition_node.py
@@ -36,6 +36,10 @@ class BaseConditionNode(IConditionNode):
         return all(condition_list) if condition == 'and' else any(condition_list)
 
     def assertion(self, field_list: List[str], compare: str, value):
+        try:
+            value = self.workflow_manage.generate_prompt(value)
+        except Exception as e:
+            pass
         field_value = self.workflow_manage.get_reference_field(field_list[0], field_list[1:])
         for compare_handler in compare_handle_list:
             if compare_handler.support(field_list[0], field_list[1:], field_value, compare, value):


apps/application/flow/step_node/condition_node/impl/base_condition_node.py
```diff
@@ -36,6 +36,10 @@ def branch_assertion(self, branch):
         return all(condition_list) if condition == 'and' else any(condition_list)
 
     def assertion(self, field_list: List[str], compare: str, value):
+        try:
+            value = self.workflow_manage.generate_prompt(value)
+        except Exception as e:
+            pass
         field_value = self.workflow_manage.get_reference_field(field_list[0], field_list[1:])
         for compare_handler in compare_handle_list:
             if compare_handler.support(field_list[0], field_list[1:], field_value, compare, value):
```
shaohuzhang1, line 45: The provided code snippet has a few areas that can be improved or cleaned up:

1. **Exception Handling**: The `try-except` block around the `generate_prompt` call is empty, which may not cover all potential exceptions. Consider adding more specific exception handling to manage errors.

2. **Documentation Quality**: The docstring for both methods (`branch_assertion` and `assertion`) could benefit from more detailed descriptions of what each method does, especially regarding the parameters `field_list`, `compare`, and their expected values.

3. **Code Clarity and Readability**: Ensure that the logic within the methods is clear and easy to understand. Adding comments explaining the purpose and functionality of sections of the code can enhance readability.

4. **Performance Considerations**: If repeated calls are made to `workflow_manage.get_reference_field`, consider caching results to minimize redundant computations.

Here's an updated version with suggested improvements:

```python
from typing import List

class WorkflowManage:
    # Assuming this class exists and handles reference fields generation and retrieval
    pass

def branch_assertion(branch):
    """
    Determines whether all conditions in the given list evaluate to True when combined using logical AND.
    
    :param branch: A dictionary containing field lists and comparison operations
    :return: Boolean result based on the combination of conditions
    """
    condition_list = [
        getattr(self.workflow_manage, cond)(values)
        for values, cond in zip(batch[1:], batch[::2])
    ]
    return all(condition_list) if batch[0] == 'and' else any(condition_list)

def assertion(
    workflow_manage_instance,
    field_list: List[str],
    compare: str,
    value
):
    """
    Performs assertions based on field comparisons and optional prompt generation.
    
    :param workflow_manage_instance: An instance of WorkflowManage used to generate prompts
    :param field_list: List of strings representing the target field(s) involved in the assertion
    :param compare: String indicating the type of comparison ('>' (greater than), '<' (less than), '=', etc.)
    :param value: Value against which the fields will be compared
    """
    try:
        value = workflow_manage_instance.generate_prompt(value)
    except Exception as e:
        # Handle the exception appropriately, e.g., log it or raise a custom error message
        print(f"An error occurred during prompt generation: {e}")
        value = None  # You might want to set a default or skip further checks here

    field_value = workflow_manage_instance.get_reference_field(field_list[0], field_list[1:])
    
    # Add support handlers based on available functions
    for compare_handler in compare_support_handlers:
        if compare_handler.supports(field_list[0], field_list[1:], field_value, compare, value):
            return compare_handler.evaluate(field_value, value, compare)
```

### Key Changes Made:
- Added try-except block in the `assertion` method for better error management.
- Enhanced documentation with more comprehensive explanations of method parameters and intentions.
- Removed unnecessary leading spaces to adhere to Python style conventions.

PR #2326, opened 2025-02-19
fix: Tokens cannot be obtained from the model dialogue

shaohuzhang1:
fix: Tokens cannot be obtained from the model dialogue

### DIFF
--- a/apps/setting/models_provider/impl/base_chat_open_ai.py
+++ b/apps/setting/models_provider/impl/base_chat_open_ai.py
@@ -52,7 +52,8 @@ class BaseChatOpenAI(ChatOpenAI):
             run_manager: Optional[CallbackManagerForLLMRun] = None,
             **kwargs: Any,
     ) -> Iterator[ChatGenerationChunk]:
-
+        kwargs["stream"] = True
+        kwargs["stream_options"] = {"include_usage": True}
         """Set default stream_options."""
         stream_usage = self._should_stream_usage(kwargs.get('stream_usage'), **kwargs)
         # Note: stream_options is not a valid parameter for Azure OpenAI.
@@ -63,7 +64,6 @@ class BaseChatOpenAI(ChatOpenAI):
         if stream_usage:
             kwargs["stream_options"] = {"include_usage": stream_usage}
 
-        kwargs["stream"] = True
         payload = self._get_request_payload(messages, stop=stop, **kwargs)
         default_chunk_class: Type[BaseMessageChunk] = AIMessageChunk
         base_generation_info = {}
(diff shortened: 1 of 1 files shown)


apps/setting/models_provider/impl/base_chat_open_ai.py
```diff
@@ -121,6 +118,9 @@ def _stream(
                         generation_chunk.text, chunk=generation_chunk, logprobs=logprobs
                     )
                 is_first_chunk = False
+                # custom code
+                if generation_chunk.message.usage_metadata is not None:
+                    self.usage_metadata = generation_chunk.message.usage_metadata
                 yield generation_chunk
 
     def _create_chat_result(self,
```
shaohuzhang1, line 126: The provided code snippet appears to be a part of an implementation for handling streaming responses from a chat model, specifically within a class called `Stream`. Here's a breakdown with notes on potential improvements:

### Key Issues Identified:
1. **Repeated Stream Option Setting**: The stream option (`stream` attribute) is set multiple times without considering whether it was already defined.
2. **Redundant Usage Metadata Handling**: There seems to be redundant logic around usage metadata retrieval.

### Potential Improvements:
1. **Single Stream Option Check**:
   Ensure that the `stream` option is only set once, ideally during initialization or just before initiating the stream process. This reduces ambiguity and potential bugs.

```python
# Single check for stream option
if not kwargs.get('stream', False):
    kwargs["stream"] = True
```

2. **Avoid Redundant Usage Metadata Retrieval**:
   Remove unnecessary checks and assignments related to retrieving usage metadata because it might lead to overwriting values if they are fetched more than once.

```python
# Custom code removed for clarity
```

3. **Enhance Error Handling (Optional)**:
   While not directly addressed in this snippet, consider adding error handling mechanisms to manage edge cases such as invalid inputs or timeouts when fetching response chunks.

### Suggested Changes:
Here’s how you could refactor the function based on these guidelines:

```python
def _stream(
    messages: list[base_message.BaseMessage],
    stop: Optional[list[str]] = None,
    callbacks: Optional[List[CallbackHandler]] = None,
    verbose: bool = False,
    use_cache: bool = True,  # Assuming there's a need for caching
    llm_backend=None,
    run_manager: Optional[ CallbackManagerForLLMRun] = None,
    **kwargs: Any,
) -> Iterator[ChatGenerationChunk]:
    """
    Set default stream_options and initiate streaming response.
    """
    if llm_backend == "azure":
        del kwargs["stream"]
    
    # Ensure stream option is set correctly
    kwargs["stream"] = kwargs.get("stream", False)
    
    if kwargs.get("stream"):
        # Additional setup for streaming can go here
    
    payload = self._get_request_payload(messages, stop=stop, use_cache=use_cache, **kwargs)
    default_chunk_class: Type[BaseMessageChunk] = AIMessageChunk
    base_generation_info = {}
    
    # Rest of the code remains mostly unchanged

# Example usage
async def main():
    async for chunk in client.stream(["Hello"], use_stream=True):
        print(chunk.text)

```

### Conclusion:
By ensuring the stream option is consistently managed and avoiding redundant operations concerning usage metadata, we improve both readability and robustness of the `_stream` method. These changes also enhance efficiency and reliability while maintaining consistency throughout the implementation.

PR #2347, opened 2025-02-21
fix: Child application node, input parameters too long, error reported during Q&A

shaohuzhang1:
fix: Child application node, input parameters too long, error reported during Q&A

### DIFF
--- a/apps/application/flow/step_node/application_node/impl/base_application_node.py
+++ b/apps/application/flow/step_node/application_node/impl/base_application_node.py
@@ -174,7 +174,7 @@ class BaseApplicationNode(IApplicationNode):
         current_chat_id = string_to_uuid(chat_id + application_id)
         Chat.objects.get_or_create(id=current_chat_id, defaults={
             'application_id': application_id,
-            'abstract': message
+            'abstract': message[0:1024]
         })
         if app_document_list is None:
             app_document_list = []


apps/application/flow/step_node/application_node/impl/base_application_node.py
```diff
@@ -174,7 +174,7 @@ def execute(self, application_id, message, chat_id, chat_record_id, stream, re_c
         current_chat_id = string_to_uuid(chat_id + application_id)
         Chat.objects.get_or_create(id=current_chat_id, defaults={
             'application_id': application_id,
-            'abstract': message
+            'abstract': message[0:1024]
         })
         if app_document_list is None:
             app_document_list = []
```
shaohuzhang1, line 180: The code provided appears to have several issues and potential improvements:

1. **SQL Injection**: The `string_to_uuid()` function, which seems to convert a string to a UUID, might be vulnerable to SQL injection if it's not properly sanitized or escaped.

2. **UUID Creation**: Using `chat_id + application_id` directly to create a UUID could lead to collisions if not handled carefully. It would be safer to use these values separately when generating the UUIDs.

3. **String Length Handling**: Limiting the length of `message` to 1024 characters ensures that the database can handle this size without running into storage limitations. However, consider using a field with an appropriate maximum length, such as Unicode up to 768 bytes (`max_length=384`) to cover most text data needs effectively.

4. **Optimization**:
   - If `app_document_list` can contain many items, consider storing these as separate documents rather than a single object, especially considering scalability demands.
   - Ensure that the `ChatRecord` model is also updated or created appropriately based on the chat ID, case record ID, and any other relevant information that relates back to communication records.

Here's slightly improved version of the code incorporating some of these considerations:

```python
from uuid import uuid4

def execute(self, application_id, message, chat_id, chat_record_id, stream, rec):
    # Use a unique identifier for each request instance
    current_request_uuid = str(uuid4())

    # Generate a custom UUID based on specific identifiers (not recommended, better handling through models)
    # current_chat_id = f"{current_request_uuid}-{chat_id[:8]}-{application_id}"

    chat_record = ChatRecord.objects.get_or_create(
        id=chat_record_id,
        default_values={'rec': rec}
    )

    if not isinstance(message, str) or len(message) > 1024:
        raise ValueError("Message must be a string of at most 1024 characters.")

    Chat.objects.create(
        id=f"{current_request_uuid}-ch-{chat_record.id}",
        application_id=application_id,
        abstract=message[:1024],
        chat_record=chat_record
    )
```

### Key Improvements Made:
- **Use of UUIDs**: Created a unique request UUID instead of relying on concatenating strings to generate IDs. This approach aligns more closely with best practices for creating consistent, secure IDs.
- **Validation**: Added basic validation to ensure the input string meets expected criteria (is a string and no longer than 1024 characters).
- **Improved Error Handling**: Raised an error if the message exceeds the allowed length, providing clear feedback to the caller.

These changes aim to enhance the security, stability, and performance of the code while maintaining its functionality.

PR #2372, opened 2025-02-24
fix: Workflow execution decimal cannot be serialized

shaohuzhang1:
fix: Workflow execution decimal cannot be serialized

### DIFF
--- a/apps/application/models/application.py
+++ b/apps/application/models/application.py
@@ -7,6 +7,7 @@
     @desc:
 """
 import datetime
+import decimal
 import json
 import uuid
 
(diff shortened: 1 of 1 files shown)


apps/application/models/application.py
```diff
@@ -140,6 +141,8 @@ def default(self, obj):
             return str(obj)
         if isinstance(obj, datetime.datetime):
             return obj.strftime("%Y-%m-%d %H:%M:%S")
+        if isinstance(obj, decimal.Decimal):
+            return float(obj)
         else:
             return json.JSONEncoder.default(self, obj)
```
shaohuzhang1, line 148: No significant issues were found in the provided code. However, there is a minor suggestion for improvement:

```python
+        if isinstance(obj, decimal.Decimal):
+# Use float() to convert Decimal instead of str()
+            return float(obj)
```

This change ensures that when `decimal.Decimal` objects are encountered during serialization, they are converted to a floating-point number rather than converting them to strings. This might be more suitable depending on how you want to store or manipulate these numbers later in your application.

PR #2373, opened 2025-02-24
fix: Remove vllm image cache

shaohuzhang1:
fix: Remove vllm image cache  --bug=1052365 --user=刘瑞斌 【github#2353】vllm视觉模型修改最大tokens不生效 https://www.tapd.cn/57709429/s/1657667

### DIFF
--- a/apps/setting/models_provider/impl/vllm_model_provider/model/image.py
+++ b/apps/setting/models_provider/impl/vllm_model_provider/model/image.py
@@ -18,3 +18,6 @@ class VllmImage(MaxKBBaseModel, BaseChatOpenAI):
             stream_usage=True,
             **optional_params,
         )
+
+    def is_cache_model(self):
+        return False


apps/setting/models_provider/impl/vllm_model_provider/model/image.py
```diff
@@ -18,3 +18,6 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
             stream_usage=True,
             **optional_params,
         )
+
+    def is_cache_model(self):
+        return False
```
shaohuzhang1, line 23: The provided code contains a few minor improvements:

1. Added a docstring to the `is_cache_model` method describing its purpose and functionality.

```python
+++ 
def is_cache_model(self):
    """
    Returns False indicating that this model type does not utilize caching.
    """
    return False
```

These adjustments enhance code readability and documentation. The rest of the code seems well-structured and doesn't require significant changes to ensure it functions correctly within the context described.

PR #2406, opened 2025-02-26
fix: the error in obtaining the document list

shaohuzhang1:
fix: the error in obtaining the document list

### DIFF
--- a/apps/dataset/serializers/document_serializers.py
+++ b/apps/dataset/serializers/document_serializers.py
@@ -19,7 +19,7 @@ from typing import List, Dict
 import openpyxl
 from celery_once import AlreadyQueued
 from django.core import validators
-from django.db import transaction
+from django.db import transaction, models
 from django.db.models import QuerySet, Count
 from django.db.models.functions import Substr, Reverse
 from django.http import HttpResponse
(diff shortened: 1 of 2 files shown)


apps/dataset/serializers/document_serializers.py
```diff
@@ -443,11 +443,17 @@ def get_query_set(self):
...
-                query_set = query_set.order_by('-create_time', 'id')
-            return query_set
+                order_by_query_set = order_by_query_set.order_by('-create_time', 'id')
+            return {
+                'document_custom_sql': query_set,
+                'order_by_query': order_by_query_set
+            }
 
         def list(self, with_valid=False):
             if with_valid:
```
shaohuzhang1, line 459: The code is mostly clean, but there are some minor improvements that can be made:

1. In the `get_query_set` method, you may want to ensure that the dynamic model retrieved by `get_dynamics_model` has all the fields specified (`'char_length'`, `'paragraph_count'`, `"update_time"`, `'create_time'`) defined and accessible.

2. When returning the results, instead of using a dictionary with keys `'document_custom_sql'` and `'order_by_query'`, it might make more sense to directly define these variables within the function scope for clarity:
```python
def get_query_set(self):
        # existing logic...

        return query_set, order_by_query_set
```

3. If possible, consider moving away from using hardcoded field names like `'status'`. You could set them up dynamically based on configuration settings so they're easier to change later without modifying many places in your codebase.

These changes won't significantly impact functionality but will improve readability and maintainability slightly.

PR #2409, opened 2025-02-26
fix: URL encoding parameters not decompiled

shaohuzhang1:
fix: URL encoding parameters not decompiled

### DIFF
--- a/ui/src/components/ai-chat/component/user-form/index.vue
+++ b/ui/src/components/ai-chat/component/user-form/index.vue
@@ -286,17 +286,41 @@ const checkInputParam = () => {
   let msg = []
   for (let f of apiInputFieldList.value) {
     if (!api_form_data_context.value[f.field]) {
-      api_form_data_context.value[f.field] = route.query[f.field]
+      let _value = route.query[f.field]
+      if (_value != null) {
+        if (_value instanceof Array) {
+          _value = _value
+            .map((item) => {
+              if (item != null) {
+                return decodeQuery(item)
+              }
+              return null
+            })
+            .filter((item) => item != null)
+        } else {
+          _value = decodeQuery(_value)
+        }
+        api_form_data_context.value[f.field] = _value
+      }
     }
     if (f.required && !api_form_data_context.value[f.field]) {
       msg.push(f.field)
     }
   }
   if (msg.length > 0) {
-    MsgWarning(`${t('chat.tip.inputParamMessage1')} ${msg.join('、')}${t('chat.tip.inputParamMessage2')}`)
+    MsgWarning(
+      `${t('chat.tip.inputParamMessage1')} ${msg.join('、')}${t('chat.tip.inputParamMessage2')}`
+    )
     return false
   }
   return true
+}
+const decodeQuery = (query: string) => {
+  try {
+    return decodeURIComponent(query)
+  } catch (e) {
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/user-form/index.vue
```diff
@@ -286,18 +286,42 @@ const checkInputParam = () => {
...
+const decodeQuery = (query: string) => {
+  try {
+    return decodeURIComponent(query)
+  } catch (e) {
+    return query
+  }
+}
 defineExpose({ checkInputParam })
 onMounted(() => {
   handleInputFieldList()
```
shaohuzhang1, line 327: The provided code has some minor issues that can be addressed:

### Issues Identified

1. **Duplicate Code**: The `decodeQuery` function is declared twice, which should not duplicate functionality.
2. **Comment Formatting**: The comments need to be formatted consistently and properly.
3. **Variable Naming**: There's a redundant `const _value = route.query[f.field];` line that can be removed before the assignment.

### Optimization Suggestions

1. **Avoid Unnecessary Mapping**: If you know the type of array elements (e.g., objects), avoid filtering out `null` values when mapping them back.

   ```javascript
   if (_value instanceof Array) {
     _value = _value.map((item) => decodeURIComponent(item));
     // Remove duplicates if they exist
     // _value = [...new Set(_value)];
   } else {
     _value = decodeURIComponent(_value);
   }
   ```

2. **Handle Edge Cases**: Ensure that edge cases like empty strings or non-string query parameters are correctly handled during decoding.

### Revised Code

Here's the revised version with these suggestions applied:

```javascript
@@ -286,28 +286,45 @@ const checkInputParam = () => {
     for (let f of apiInputFieldList.value) {
         if (!api_form_data_context.value[f.field]) {
             let _value = route.query[f.field];
             if (_value != null) {
                 if (_value instanceof Array) {
                     _value = _value.map((item) => decodeURIComponent(item));
                     api_form_data_context.value[f.field] = [...uniqueValues];
                 } else {
                     try {
                         _value = decodeURIComponent(_value);
                     } catch (e) {
                         console.error('Error decoding query parameter:', e);
                         _value = null;
                     }
                 }
             } else {
                 _value = route.query[f.field]
             }
             api_form_data_context.value[f.field] = _value;
         }

         if (f.required && !api_form_data_context.value[f.field]) {
             msg.push(f.field)
         }
     }
 
     if (msg.length > 0) {
-        MsgWarning(`${t('chat.tip.inputParamMessage1')} ${msg.join('、')}${t('chat.tip.inputParamMessage2')}`)
+        MsgWarning(
+            `${t('chat.tip.inputParamMessage1')} ${msg.join('、')}${t('chat.tip.inputParamMessage2')}`
+        );
         return false
     }
     return true;
 }

const decodeQuery = (query: string) => {
    let decodedValue = '';
    try {
        decodedValue = decodeURIComponent(query);
    } catch (e) {
        console.error('Error decoding URI component', e);
    }
    return decodedValue;
}

defineExpose({checkInputParam})
onMounted(() => {
    handleInputFieldList();
});
```

This revision ensures that the `codeDecodeUriComponent()` is only called once and also handles potential errors in URL decoding more gracefully. Additionally, it uses `[]([...array].concat(array))` to remove duplicates from an array.

PR #2412, opened 2025-02-26
fix: VLLM Supplier Dialogue Verification

shaohuzhang1:
fix: VLLM Supplier Dialogue Verification

### DIFF
--- a/apps/setting/models_provider/impl/vllm_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/vllm_model_provider/credential/llm.py
@@ -48,9 +48,11 @@ class VLLMModelCredential(BaseForm, BaseModelCredential):
         model = provider.get_model(model_type, model_name, model_credential, **model_params)
         try:
             res = model.invoke([HumanMessage(content=gettext('Hello'))])
-            print(res)
         except Exception as e:
-            print(e)
+            raise AppApiException(ValidCode.valid_error.value,
+                                  gettext(
+                                      'Verification failed, please check whether the parameters are correct: {error}').format(
+                                      error=str(e)))
         return True
 
     def encryption_dict(self, model_info: Dict[str, object]):


apps/setting/models_provider/impl/vllm_model_provider/credential/llm.py
```diff
@@ -48,9 +48,11 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
...
-            print(res)
         except Exception as e:
-            print(e)
+            raise AppApiException(ValidCode.valid_error.value,
+                                  gettext(
+                                      'Verification failed, please check whether the parameters are correct: {error}').format(
+                                      error=str(e)))
         return True
 
     def encryption_dict(self, model_info: Dict[str, object]):
```
shaohuzhang1, line 58: The code looks mostly well-written, but there's one issue that needs improvement: the exception handling for invalid calls to `get_model` might not be sufficient for all cases.

**Improvement Suggestions:**
1. **Ensure Proper Error Handling:** Add checks after calling `get_model()` to verify if `model` is not null or returns an error value before attempting to invoke it. This can help prevent unexpected failures from unhandled errors.
   ```python
   def is_valid(self, model_type: str, model_name, model_credential: Dict[str, object], model_params):
       # Ensure proper model retrieval and initial validation
       if model is None:  # Handle case where model retrieval fails
           raise AppApiException(ValidCode.model_not_found.value,
                                gettext('Model not found'))

       try:
           res = model.invoke([HumanMessage(content=gettext('Hello'))])
          # Print response for debugging purposes (optional)
       except Exception as e:
           raise AppApiException(ValidCode.valid_error.value,
                               gettext('Verification failed, please check whether the parameters are correct: {error}')
                                   .format(error=str(e)))

       return True

   ```

By adding these modifications, you improve robustness by checking if the model was successfully retrieved before proceeding with other operations, which can reduce runtime exceptions due to missing models. Also note that using `None` here is more precise than relying on truthiness for `provider.get_model()` responses.

Issue #2389, opened 2025-02-25
[Bug] Create a function library. When the result is 0, the debug output is -


spritelyl:
### Contact Information

14707070227

### MaxKB Version

v1.10.1

### Problem Description

创建函数库，当结果为0的时候，调试输出为-

### Steps to Reproduce

1. 创建一个函数库，里面是简单的减法运算

<img width="302" alt="Image" src="https://github.com/user-attachments/assets/673fedfd-69a5-454f-ba93-f0c1fe682797" />

2. 当结果为0的时候，调试输出为-

<img width="504" alt="Image" src="https://github.com/user-attachments/assets/f19de6ce-039b-40ef-b411-8ee181c40123" />

### The expected correct result

当结果为0的时候，调试输出正常的数字0

### Related log output

```shell
无
```

### Additional Information

无

baixin513:
感谢反馈，后续版本中修复。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, fixed in subsequent versions.

baixin513:
v1.10.2-lts已发布。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.2-lts has been released.

PR #2413, opened 2025-02-26
fix: Resolve the issue of a function returning 0 as a result (#2389)

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/function-lib/component/FunctionDebugDrawer.vue
+++ b/ui/src/views/function-lib/component/FunctionDebugDrawer.vue
@@ -84,7 +84,7 @@
           shadow="never"
           style="max-height: 350px; overflow: scroll"
         >
-          {{ result || '-' }}
+          {{ String(result) == '0' ? 0 : result || '-' }}
         </el-card>
       </div>
     </div>


ui/src/views/function-lib/component/FunctionDebugDrawer.vue
```diff
@@ -84,7 +84,7 @@
           shadow="never"
           style="max-height: 350px; overflow: scroll"
         >
-          {{ result || '-' }}
+          {{ String(result) == '0' ? 0 : result || '-' }}
         </el-card>
       </div>
     </div>
```
shaohuzhang1, line 90: The code has no immediate logical errors but can be improved and optimized:

1. **Use Template Literals**: The use of single quotes for strings inside the template expression `{{ result || '-' }}` is unnecessary when you're using the `.toString()` function later on in the same expression.

2. **Boolean Check**: The use of strict equality (`===`) with `'0'` is redundant because it will return `false` even if `result` is a number that equals zero (e.g., `0`, `-0`). A simple check with loose equality (`==`) suffices.

3. **Code Readability**: While minor, using descriptive variable names could improve readability, such as changing `result` to something like `responseData`.

Here’s an updated version of the code:

```html
<div class="card-container">
  <vue-custom-scrollbar max-height="350px" overflow-y-scroll>
    <el-card :shadow="'never'" :style="{ maxHeight: '350px', overflowY: 'scroll' }">
      {{ Number(responseData) === 0 ? 0 : responseData || '-' }}
    </el-card>
  </vue-custom-scrollbar>
</div>
```

This change ensures better JavaScript logic and improves readability while maintaining functionality.

PR #2418, opened 2025-02-26
fix: Fix variable-assign-node cannot publish

shaohuzhang1:
fix: Fix variable-assign-node cannot publish  --bug=1052464 --user=刘瑞斌 【应用】-编排应用中存在变量赋值节点，发布应用失败 https://www.tapd.cn/57709429/s/1659401

### DIFF
--- a/ui/src/workflow/nodes/variable-assign-node/index.vue
+++ b/ui/src/workflow/nodes/variable-assign-node/index.vue
@@ -209,11 +209,14 @@ const nodeCascaderRef = ref()
 const nodeCascaderRef2 = ref()
 const validate = async () => {
   // console.log(replyNodeFormRef.value.validate())
-  return Promise.all([
+  let ps = [
     replyNodeFormRef.value?.validate(),
-    ...nodeCascaderRef.value.map((item: any) => item.validate()),
-    ...nodeCascaderRef2.value.map((item: any) => item.validate())
-  ]).catch((err: any) => {
+    ...nodeCascaderRef.value.map((item: any) => item.validate())
+  ]
+  if (nodeCascaderRef2.value) {
+    ps = [...ps, ...nodeCascaderRef.value.map((item: any) => item.validate())]
+  }
+  return Promise.all(ps).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }


ui/src/workflow/nodes/variable-assign-node/index.vue
```diff
@@ -209,11 +209,14 @@ const nodeCascaderRef = ref()
...
-  ]).catch((err: any) => {
+    ...nodeCascaderRef.value.map((item: any) => item.validate())
+  ]
+  if (nodeCascaderRef2.value) {
+    ps = [...ps, ...nodeCascaderRef.value.map((item: any) => item.validate())]
+  }
+  return Promise.all(ps).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
   })
 }
```
shaohuzhang1, line 222: The provided code is mostly correct, but there's an issue with the structure of the `nodeCascaderRef2` mapping function when it might be undefined. It uses `...[]` which can lead to unexpected behavior due to type widening if `nodeCascaderRef2.value` is not defined.

Here are some recommendations:

1. **Check for Undefined Value**: Before attempting to map on `nodeCascaderRef2`, ensure that it exists. This prevents errors from occurring when calling `.map()` on an undefined value.

2. **Optimization**: The current implementation will execute `validate()` calls serially instead of parallel for both `replyNodeFormRef` and each option in `nodeCascaderRef`. If you want to execute them in series, keep the existing `Promise.all([...]).catch()`. However, this ensures they are executed one after another without changing their order.

3. **Example Code**:

```javascript
  let errorCount = 0;

  const allValidationsPromises = [
    replyNodeFormRef.value?.validate().then(() => {
      // Count validation success for replyNodeFormRef
      errorCount++;
    }).catch(err => {
      console.error('Validation failed:', err);
      errorCount++;
    }),

    ...nodeCascaderRef.value?.map(item => item.validate()).map(validateResult => {
      if (!validateResult.validated) { // Assuming each cascader node has a validated property indicating successful validation
        // Mark error for cascader nodes
        errorCount++;
      }
      return validateResult;
    }) || []

  ];

  await Promise.all(allValidationsPromises);

  if (errorCount === 0) {
    logger.info(`All validations passed.`);
  } else {
    throw new Error(`Failed to verify the form details: ${errorCount} errors found.`);
  }
};
```

4. **Error Handling**: Ensure proper handling within `catch()` blocks. In the example above, if there are no validations failures (`errorCount === 0`) it logs a success message; otherwise, it throws an exception detailing how many checks failed.

These improvements help manage edge cases such as uninitialised references more gracefully, avoiding runtime TypeScript errors during evaluation.

PR #2423, opened 2025-02-27
refactor: Support the generated images horizontal tiling display

shaohuzhang1:
refactor: Support the generated images horizontal tiling display  --story=1017907 --user=刘瑞斌 【越秀农牧】图片生成节点，生成的图片，希望可以支持横向平铺展示 https://www.tapd.cn/57709429/s/1659636

### DIFF
--- a/apps/application/flow/step_node/image_generate_step_node/impl/base_image_generate_node.py
+++ b/apps/application/flow/step_node/image_generate_step_node/impl/base_image_generate_node.py
@@ -47,7 +47,7 @@ class BaseImageGenerateNode(IImageGenerateNode):
             file_url = FileSerializer(data={'file': file, 'meta': meta}).upload()
             file_urls.append(file_url)
         self.context['image_list'] = [{'file_id': path.split('/')[-1], 'url': path} for path in file_urls]
-        answer = '\n'.join([f"![Image]({path})" for path in file_urls])
+        answer = ' '.join([f"![Image]({path})" for path in file_urls])
         return NodeResult({'answer': answer, 'chat_model': tti_model, 'message_list': message_list,
                            'image': [{'file_id': path.split('/')[-1], 'url': path} for path in file_urls],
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/image_generate_step_node/impl/base_image_generate_node.py
```diff
@@ -47,7 +47,7 @@ def execute(self, model_id, prompt, negative_prompt, dialogue_number, dialogue_t
             file_url = FileSerializer(data={'file': file, 'meta': meta}).upload()
             file_urls.append(file_url)
         self.context['image_list'] = [{'file_id': path.split('/')[-1], 'url': path} for path in file_urls]
-        answer = '\n'.join([f"![Image]({path})" for path in file_urls])
+        answer = ' '.join([f"![Image]({path})" for path in file_urls])
         return NodeResult({'answer': answer, 'chat_model': tti_model, 'message_list': message_list,
                            'image': [{'file_id': path.split('/')[-1], 'url': path} for path in file_urls],
                            'history_message': history_message, 'question': question}, {})
```
shaohuzhang1, line 53: The provided code looks mostly correct for its intended purpose of generating and returning image URLs associated with a given `model_id`, `prompt`, and other parameters. However, there is one small change you can make to optimize it:

**Improvement: Remove the newline character**

The current implementation uses `'` instead of `\n` when joining the list comprehension that creates the HTML image tags. While this difference typically doesn't affect rendering for single-image responses (like what should be happening here), using `/n` ensures consistency across different environments where multiple newline characters might be treated differently. Here's the updated line:

```python
answer = ' '.join([f"![Image]({path})" for path in file_urls])
```

This makes sure all paths and links start on a new line within the generated text.

Overall, this change enhances readability by keeping each image link on a separate line while also maintaining compatibility with various rendering contexts.

PR #2427, opened 2025-02-27
fix: Fix variable assign num not working

shaohuzhang1:
fix: Fix variable assign num not working  --bug=1052497 --user=刘瑞斌 【变量赋值】-将全局变量赋值为num类型，输出时报错 https://www.tapd.cn/57709429/s/1659733

### DIFF
--- a/apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
+++ b/apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
@@ -30,9 +30,13 @@ class BaseVariableAssignNode(IVariableAssignNode):
                             val = json.loads(variable['value'])
                         self.workflow_manage.context[variable['fields'][1]] = val
                         result['output_value'] = variable['value'] = val
-                    else:
+                    elif variable['type'] == 'string':
                         # 变量解析 例如：{{global.xxx}}
                         val = self.workflow_manage.generate_prompt(variable['value'])
+                        self.workflow_manage.context[variable['fields'][1]] = val
+                        result['output_value'] = val
+                    else:
+                        val = variable['value']
                         self.workflow_manage.context[variable['fields'][1]] = val
                         result['output_value'] = val
                 else:


apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
```diff
@@ -30,11 +30,15 @@ def execute(self, variable_list, stream, **kwargs) -> NodeResult:
...
                         val = self.workflow_manage.generate_prompt(variable['value'])
                         self.workflow_manage.context[variable['fields'][1]] = val
                         result['output_value'] = val
+                    else:
+                        val = variable['value']
+                        self.workflow_manage.context[variable['fields'][1]] = val
+                        result['output_value'] = val
                 else:
                     reference = self.get_reference_content(variable['reference'])
                     self.workflow_manage.context[variable['fields'][1]] = reference
```
shaohuzhang1, line 44: The provided code snippet appears to be a function within a class called `execute`. It's designed to handle different types of variables during execution. Here are a few comments on the code:

1. **Empty Default Case**: The commented-out `else` block after checking if the type is `'string'` suggests that there might have been a default case intended, but it's currently not used. Consider removing this commented-out block or making it functional.

2. **Type Check for 'string' Only**: Since the condition checks only for `'str'`, all other types (like `dict` or any custom object) will fall into the default `if` block or remain unchanged by the current logic. This can lead to potential issues when trying to process complex variable values.

3. **Variable Value Assignment**: Assigning both `result['output_value']` and `variable['value']` inside the same condition might not be necessary unless specifically required in the context of some operation you're missing from the comment above.

   - **Avoid Repeated Context Updates**: Ensure that updating the workflow context is done efficiently without redundant operations.
   - **Validate Variable Type Before Parsing**: Although your parsing based on type seems correct now, consider adding additional validation before attempting to convert strings to JSON, dictionaries, or other objects.

Here’s an improved version with these considerations taken into account:

```python
def execute(self, variable_list, stream, **kwargs):
    result = {'output_value': ''}
        if variable['type'].lower() == 'json':
            try:
                val = json.loads(variable['value'])
            except (JSONDecodeError, TypeError):
                print(f"Failed to decode JSON: {variable['value']}")
                continue
        elif variable['type'].upper() == 'STRING':
            # Parse string variables using self.workflow_manage.generate_prompt
            val = self.workflow_manage.generate_prompt(variable['value'])
        # Assuming no conversion needed for scalar primitives (int, float)
        else:
            val = variable['value']

        self.workflow_manage.context[variable['fields'][1]] = val

        # Optionally, update output value
        result['output_value'] = val

    return NodeResult(result=result, status_code=0)  # Add appropriate status codes
```

This revision includes type-checking explicitly for `"json"` and `"STRING"`, uses exceptions to manage `JSONDecodeError`/`TypeError` cases gracefully, ensures consistent variable handling, and improves code readability by reducing duplicate assignments and improving conditional structure.

PR #2431, opened 2025-02-27
refactor: Cancel last autoplay audio

shaohuzhang1:
refactor: Cancel last autoplay audio  --story=1017953 --user=刘瑞斌 【优化】多轮对话，语音识别自动播放，会重叠播放 https://www.tapd.cn/57709429/s/1660001

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -86,6 +86,7 @@ import applicationApi from '@/api/application'
 import { datetimeFormat } from '@/utils/time'
 import { MsgError } from '@/utils/message'
 import { t } from '@/locales'
+import bus from '@/bus'
 const route = useRoute()
 const {
   params: { id }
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -277,6 +278,11 @@ const pausePlayAnswerText = () => {
...
 
 onMounted(() => {
+  bus.on('pause-autoplay', () => {
+    pausePlayAnswerText()
+    // console.log(1234)
+  })
+  bus.emit('pause-autoplay')
   // 第一次回答后自动播放， 打开历史记录不自动播放
   if (props.tts && props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
     playAnswerText(buttonData.value.answer_text)
```
shaohuzhang1, line 288: There are no irregularities in the provided code, but there is one potential issue related to the handling of `onMounted`:

```javascript
if (props.tts && props.tts_autoplay && buttonData.value.write_ed && !buttonData.value.update_time) {
```

This line checks conditions that might lead to unexpected behavior when used with asynchronous functions or if `buttonData` changes later on. It would be more robust to place this logic inside an effect (`watch`, `computed`, etc.) instead.

Additionally, ensure that `bus.emit('pause-autoplay')` doesn't cause issues with event listeners already attached. If multiple components are listening for the same event, it's important to make sure you're not overwriting existing listeners unintentionally. 

Optimization suggestions could include avoiding unnecessary operations within lifecycle hooks like `onMounted`.

PR #2448, opened 2025-02-28
fix: Document name sorting

shaohuzhang1:

### DIFF
--- a/ui/src/views/document/index.vue
+++ b/ui/src/views/document/index.vue
@@ -64,12 +64,7 @@
           :storeKey="storeKey"
         >
           <el-table-column type="selection" width="55" :reserve-selection="true" />
-          <el-table-column
-            prop="name"
-            :label="$t('views.document.table.name')"
-            min-width="280"
-            sortable
-          >
+          <el-table-column prop="name" :label="$t('views.document.table.name')" min-width="280">
             <template #default="{ row }">
               <ReadWrite
                 @change="editName($event, row.id)"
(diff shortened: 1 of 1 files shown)


ui/src/views/document/index.vue
```diff
@@ -829,7 +819,7 @@ function getList(bool?: boolean) {
   const param = {
     ...(filterText.value && { name: filterText.value }),
     ...filterMethod.value,
-    order_by: orderBy.value,
+    order_by: orderBy.value
   }
   documentApi
     .getDocument(id as string, paginationConfig.value, param, bool ? undefined : loading)
```
shaohuzhang1, line 825: The provided code looks generally correct for an El Table component setup. Here's a concise review with some minor suggestions:

### Potential Issues
1. **Duplicate Code**: The "Create Time" column has both `sortable` set to `false` and no `filters` property defined in the template slot (`#default`). This might be unintentional.

2. **Parameter Merge Logic in `getList` Function**:
    - Line 829 is almost identical to line 826 but uses arrow functions instead of traditional callbacks. Ensure this behavior aligns with your design preferences.

3. **TypeScript Typing**: There are type annotations like `(hitHandlingMethod[row.hit_handling_method as keyof typeof hitHandlingMethod])` which should ensure that the lookup using `row.hit_handling_method` returns a valid index into the dictionary `hitHandlingMethod`. If it could return null/undefined, additional error handling might be necessary.

### Optimization Suggestions

1. **Simplified Filter Handling**: Consider consolidating duplicate logic around filters by creating a reusable utility function or merging conditional checks within a single method call where applicable (e.g., checking if `filterText.value` exists before adding its key-value pair).

2. **Consistent Arrow Functions Usage**: While not mandatory here, consistent use of either traditional callback methods or arrow functions can improve readability and consistency.

3. **Error Handling in Type Annotations**: For better developer experience, consider more descriptive type definitions. For example, replace `"views.document.table.name"` and `"common.createTime"` with their respective imported keys from `$t`.

Overall, the overall structure seems functional given the context, but these points might aid optimization and maintainability.

PR #2451, opened 2025-02-28
perf: optimization codemirror component

wangdan-fit2cloud:

### DIFF
--- a/ui/src/components/codemirror-editor/index.vue
+++ b/ui/src/components/codemirror-editor/index.vue
@@ -1,16 +1,50 @@
 <template>
-  <Codemirror
-    v-bind="$attrs"
-    ref="cmRef"
-    :extensions="extensions"
-    :style="codemirrorStyle"
-    :tab-size="4"
-    :autofocus="true"
-  />
+  <div class="codemirror-editor w-full">
+    <Codemirror
+      v-model="data"
+      ref="cmRef"
+      :extensions="extensions"
+      :style="codemirrorStyle"
+      :tab-size="4"
+      :autofocus="true"
+      v-bind="$attrs"
+    />
+
+    <div class="codemirror-editor__footer">
+      <el-button text type="info" @click="openCodemirrorDialog" class="magnify">
+        <AppIcon iconName="app-magnify" style="font-size: 16px"></AppIcon>
+      </el-button>
+    </div>
+    <!-- Codemirror 弹出层 -->
+    <el-dialog
+      v-model="dialogVisible"
+      :title="$t('views.functionLib.functionForm.form.param.code')"
+      append-to-body
+      fullscreen
+    >
+      <Codemirror
+        v-model="cloneContent"
+        :extensions="extensions"
+        :style="codemirrorStyle"
+        :tab-size="4"
+        :autofocus="true"
+        style="
+          height: calc(100vh - 160px) !important;
+          border: 1px solid #bbbfc4;
+          border-radius: 4px;
+        "
+      />
+      <template #footer>
+        <div class="dialog-footer mt-24">
+          <el-button type="primary" @click="submitDialog"> {{ $t('common.confirm') }}</el-button>
+        </div>
+      </template>
+    </el-dialog>
+  </div>
 </template>
 
 <script setup lang="ts">
-import { ref } from 'vue'
+import { ref, computed, watch } from 'vue'
 import { Codemirror } from 'vue-codemirror'
 import { python } from '@codemirror/lang-python'
 import { oneDark } from '@codemirror/theme-one-dark'
@@ -18,6 +52,19 @@ import { linter, type Diagnostic } from '@codemirror/lint'
 import FunctionApi from '@/api/function-lib'
 
 defineOptions({ name: 'CodemirrorEditor' })
+
+const props = defineProps<{
+  modelValue: any
+}>()
+const emit = defineEmits(['update:modelValue', 'submitDialog'])
+const data = computed({
+  set: (value) => {
+    emit('update:modelValue', value)
+  },
+  get: () => {
+    return props.modelValue
+  }
+})
 
 function getRangeFromLineAndColumn(state: any, line: number, column: number, end_column?: number) {
   const l = state.doc.line(line)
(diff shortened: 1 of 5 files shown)


ui/src/workflow/nodes/variable-assign-node/index.vue
```diff
@@ -212,6 +203,7 @@ function submitDialog(val: string) {
 const replyNodeFormRef = ref()
 const nodeCascaderRef = ref()
 const nodeCascaderRef2 = ref()
+
 const validate = async () => {
   // console.log(replyNodeFormRef.value.validate())
   let ps = [
```
shaohuzhang1, line 209: The provided code has several issues and improvements that can be made:

1. **`hide-required-asterisk` Property**: This property was added to an incorrect context (`require-asterisk-position="right"` is on a `b-form-group`, but here it's being set on the outer `el-form`). It should be removed since it doesn't have any effect in this scenario.

2. **Conditional Rendering of Variable Assignation Options**:
   - The custom variable input fields are wrapped inside two conditions that prevent rendering if `item.source === 'referring'`. However, when the user chooses `'reference'`, they still need to see at least the placeholder and cascading selection widget, so remove these conditional renders.
   ```vue
   <!-- Remove the below lines -->
   <!-- <div v-if="item.source === 'referring'" class="flex"> ... </div> -->
   <!-- <NodeCascader ref="nodeCascaderRef2" ... /> -->

   ```

3. **Custom Input Field Type Selection Logic**:
   - Set default values for different types to ensure proper behavior:
     - For `"string"`, initialize with an empty string rather than null.
       ```vue
       <el-input v-model="item.value" :placeholder="$t('common.inputPlaceholder')" show-word-limit clearable @wheel="wheel"></el-input>
       ```
     - For `"num"`, initialize with zero.
       ```vue
       <el-input-number v-model="item.value"></el-input-number>
       ```
   ```vue
   <!-- Ensure correct initialization before changing value -->
   item.type === "string" && (item.value = ""),
   item.type === "num" && (item.value = 0),
   item.type === "json" && (item.value = '{}'),
   ```

4. **Error Handling for Invalid JSON**:
   - Improve error handling for invalid JSON by providing more specific validation logic. Instead of always using `new Error` to indicate failure, handle parsing errors correctly based on the actual reason.
   ```typescript
   const isValidJSON = (val: string): boolean => {
     try {
       JSON.parse(val);
       return true;
     } catch (error) {
       console.error(`Invalid JSON format: ${error.message}`);
       return false;
     }
   };
   ```

5. **Improve UX for Custom JSON Input**:
   - Allow users to quickly paste JSON into the field directly or provide copy-paste buttons if applicable.
   ```html
   <el-input
     v-model="item.value"
     type="textarea"
     autosize
     @paste="handlePaste"
     @focus="(val) => { focusedJsonInput = val }"
     @blur="resetFocusJsonInput()"
   ></el-input>
   <el-button @click="copyToJsonString()">Copy to String</el-button>
   ```

6. **Add Event Listeners to Node Cascaders**:
   - Ensure all necessary event listeners such as `select` events are properly initialized when creating components.
   ```jsx
   // Initialize node cascade component after creating refs
   watch(() => form_data.variable_list, () => {
     nodeCascaderRef.value?.refresh(); // Re-initiate cascader when variable list changes
   });
   ```

7. **Optimize Form Validation Execution**:
   - Run validation only when needed instead of always executing on blur like `v-blur`.
   ```javascript
   const validateField = async (fieldKey: string) => {
     await replyNodeFormRef.value.validate([fieldKey]);
   };

   const addVariableListenerToInputs = () => {
     [...document.querySelectorAll('.el-form-item__input-wrap .el-input')]
       .filter(el => el instanceof HTMLElement)
       .forEach(inputEl => {
         inputEl.addEventListener('input', async () => {
           const key = Array.from(
             document
               .querySelectorAll<HTMLInputElement>(
                 `.el-form-item[data-index="${indexOfItemInVariablesArray}"].el-form-item`
               )
             ).map(inputWrapper => inputWrapper.querySelector(':scope > div:nth-child(2)'))[0]
               ?.getAttribute('data-key');
           
           if (!key || !form_data[key]) {
             return; 
           }

           await validateField(key); // Call validation only on change
         });
       });

       [...document.querySelectorAll('.el-form-item__content div:last-child .el-form'])
         .filter((wrapper: HTMLDivElement) => wrapper.children.length >= 19)
         .forEach(form =>
           [".el-form", ".el-select-dropdown", ".el-date-picker"] &&
             Array.from(form.childNodes).some(node => node.matches(".el-form"))
             ? form.addEventListener("scroll", () =>
               requestAnimationFrame(validateField)
             )
             : null
         );
   };
```

These suggestions will help improve the readability and functionality of the Vue.js application while addressing common design principles.

File: ui/src/components/codemirror-editor/index.vue [hunk 2]
```diff
@@ -52,9 +99,39 @@ const regexpLinter = linter(async (view) => {
...
+<style lang="scss" scoped>
+.codemirror-editor {
+  position: relative;
+  &__footer {
+    position: absolute;
+    bottom: 10px;
+    right: 10px;
+  }
+}
+</style>
```
shaohuzhang1, line 137: The code you've provided appears clean and organized for use with Vue.js along with `vue-codemirror`. It introduces several improvements over the original:

1. **State Management Using `Computed` and `Watch`**: The `data` property is now managed using both Computed properties (`computed`) and Watchers to efficiently keep the codemirror value synced between props and internal states.

2. **Fullscreen Dialog Integration**: A new El-dialog component (El-dialog's "full-screen" mode) has been added to display the same Codemirror instance in an overlay when triggered through a button click.

3. **Data Cloning and Update Handling**: In the fullscreen view, the current value of the CodeMirror editor is cloned into another variable (`cloneContent`). When editing stops, this content is then passed back via events to update the parent component.

4. Enhanced Styling**: Added CSS to ensure that only the main container takes up the full width while maintaining readability within its boundaries.

While these changes make the implementation more flexible and interactive, remember it might be beneficial to thoroughly test all user flows involving the pop-up dialog to verify smooth working and proper error handling in case anything goes wrong during runtime.

ui/src/workflow/nodes/function-node/index.vue
```diff
@@ -244,13 +211,4 @@ onMounted(() => {
...
-.function-CodemirrorEditor__footer {
-  position: absolute;
-  bottom: 10px;
-  right: 10px;
-}
-.function-CodemirrorEditor {
-  position: relative;
-}
-</style>
+<style lang="scss"></style>
```
shaohuzhang1, line 214: The code looks generally correct but has a few improvements that can be made:

  * Remove the `append-to-body` attribute from the `<el-dialog>` component. It's not necessary in this context.
  * Use CSS grid instead of flexbox to align the form items and buttons more cleanly.
  * Add some padding to the `.mb-8` class to make it look better.
  * Consider adding validation error messages directly inside the input fields, as opposed to using separate message placeholders.

Here is updated version of the code with these improvements:
```css

<style lang="scss">

.mb-8 {
  margin-bottom: 1rem;
}

.field-item {
  display: grid;
  gap: 0.5rem;

  & > .form-label-group {
    width: 35%;
  }

  & > .form-input-group {
    width: 65%;
  }
}

</style>

```

```html
... (rest of the template remains the same)
</template>
<script setup lang="ts">
...
</script>

PR #2456, opened 2025-03-03
refactor: rename function lib export subfix

shaohuzhang1:
refactor: rename function lib export subfix  --bug=1052638 --user=刘瑞斌 函数库导出后缀名优化 https://www.tapd.cn/57709429/s/1660837

### DIFF
--- a/apps/function_lib/serializers/function_lib_serializer.py
+++ b/apps/function_lib/serializers/function_lib_serializer.py
@@ -249,7 +249,7 @@ class FunctionLibSerializer(serializers.Serializer):
                 mk_instance = FlibInstance(application_dict, 'v1')
                 application_pickle = pickle.dumps(mk_instance)
                 response = HttpResponse(content_type='text/plain', content=application_pickle)
-                response['Content-Disposition'] = f'attachment; filename="{function_lib.name}.flib"'
+                response['Content-Disposition'] = f'attachment; filename="{function_lib.name}.fx"'
                 return response
             except Exception as e:
                 return result.error(str(e), response_status=status.HTTP_500_INTERNAL_SERVER_ERROR)
--- a/ui/src/api/function-lib.ts
+++ b/ui/src/api/function-lib.ts
@@ -105,7 +105,7 @@ const exportFunctionLib = (
   loading?: Ref<boolean>
 ) => {
   return exportFile(
-    name + '.flib',
+    name + '.fx',
     `${prefix}/${id}/export`,
     undefined,
     loading


apps/function_lib/serializers/function_lib_serializer.py
```diff
@@ -249,7 +249,7 @@ def export(self, with_valid=True):
                 mk_instance = FlibInstance(application_dict, 'v1')
                 application_pickle = pickle.dumps(mk_instance)
                 response = HttpResponse(content_type='text/plain', content=application_pickle)
-                response['Content-Disposition'] = f'attachment; filename="{function_lib.name}.flib"'
+                response['Content-Disposition'] = f'attachment; filename="{function_lib.name}.fx"'
                 return response
             except Exception as e:
                 return result.error(str(e), response_status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
shaohuzhang1, line 255: The code looks mostly correct but has a small suggestion for improvement:

1. The file extension used in the `filename` parameter of `response['Content-Disposition']` should be consistent. The original code uses `.flib`, but since the function is named `export_fx`, it might make sense to change this to reflect that.

**Optimization Suggestion:**
Change:
```python
response['Content-Disposition'] = f'attachment; filename="{function_lib.name}.fx"'
```

To:
```python
response['Content-Disposition'] = f'attachment; filename="{function_lib.name.lower()}.fx"'
```
This ensures the file name matches case conventions for files on different platforms and operating systems, which can be important when handling downloads from browsers.

Additionally, ensure that any error handling logic around pickling exceptions is robust, such as catching specific exceptions related to pickling errors if they exist in Django's version you're using.

ui/src/api/function-lib.ts
```diff
@@ -105,7 +105,7 @@ const exportFunctionLib = (
   loading?: Ref<boolean>
 ) => {
   return exportFile(
-    name + '.flib',
+    name + '.fx',
     `${prefix}/${id}/export`,
     undefined,
     loading
```
shaohuzhang1, line 111: The provided code snippet appears to be a function definition for an `exportFunctionLib`. It includes some small adjustments that could impact performance or readability:

1. **Consistent Naming**: 
   - The file extension from `.flib` to `.fx` is intended, but it's important to maintain consistency throughout your codebase unless there is a specific purpose for this change.

2. **Unused Parameter**:
   - There is no apparent use of the `loading` parameter within the `return` statement of this function. Although it has been included, it might not be necessary depending on how you intend to manage loading states elsewhere.

3. **Suggested Changes**:
   - Ensure that maintaining the current naming convention (`.flib`) doesn't affect other parts of your application where the `.flib` extension is used.
   - Consider removing the unused `loading` parameter if it does not serve a practical purpose in this context.

Here's the cleaned-up version with the minor changes suggested:

```javascript
const exportFunctionLib = (
  name: string,
  prefix: string,
  id: string
) => {
  return exportFile(
    name + '.fx',
    `${prefix}/${id}/export`,
    undefined
  );
};
```

Make sure these checks align with the overall architecture and requirements of your project.

PR #2462, opened 2025-03-03
fix: record log content error

shaohuzhang1:
fix: record log content error  --bug=1052670 --user=王孝刚 应用-企业微信对接应用后，在企业微信中与应用对话，对话日志中<think></think>思考过程输出了两次 https://www.tapd.cn/57709429/s/1661186

### DIFF
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -301,7 +301,7 @@ class BaseChatStep(IChatStep):
                 reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get(
                     'reasoning_content')
             post_response_handler.handler(chat_id, chat_record_id, paragraph_list, problem_text,
-                                          chat_result.content, manage, self, padding_problem_text, client_id,
+                                          content, manage, self, padding_problem_text, client_id,
                                           reasoning_content=reasoning_content if reasoning_content_enable else '')
             add_access_num(client_id, client_type, manage.context.get('application_id'))
             return manage.get_base_to_response().to_block_response(str(chat_id), str(chat_record_id),


apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
```diff
@@ -301,7 +301,7 @@ def execute_block(self, message_list: List[BaseMessage],
                 reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get(
                     'reasoning_content')
             post_response_handler.handler(chat_id, chat_record_id, paragraph_list, problem_text,
-                                          chat_result.content, manage, self, padding_problem_text, client_id,
+                                          content, manage, self, padding_problem_text, client_id,
                                           reasoning_content=reasoning_content if reasoning_content_enable else '')
             add_access_num(client_id, client_type, manage.context.get('application_id'))
             return manage.get_base_to_response().to_block_response(str(chat_id), str(chat_record_id),
```
shaohuzhang1, line 307: Based on the code snippet you provided, there is no noticeable irregularity, issue, or optimization suggestion that stands out from a cursory review. It appears to be a Python function designed to execute a block of messages with potentially additional reasoning output:

```python
def execute_block(self, message_list: List[BaseMessage],
                 paragraph_list=None, problem_text=None,
                 manage=None, client_id: str = None):
```

### Key Points Reviewed:
1. **Function Name**: The method name `execute_block` is straightforward, but it could benefit from being more descriptive for better understanding.

2. **Parameters**:
   - `message_list`: A list of `BaseMessage` objects.
   - `paragraph_list`, `problem_text`: These parameters may not always be used within the function body, considering their absence of default values and lack of usage in most scenarios.
   - `manage`: An instance of some class which might provide context about processing.
   - `client_id`: A string representing an identifier for a client.
   - `reasoning_content`: This parameter seems to be conditionally handled (not included when `reasoning_content_enable` is False).

3. **Logic Flow**:
   The logic inside involves calling `post_response_handler.handler()` with several arguments, including `chat_id`, `chat_record_id`, various lists (`paragraph_list`, `problem_text`, etc.), the main response content, the `manage` object, itself, and other variables like `padding_problem_text` and `client_id`. If `reasoning_content` exists and `reasoning_content_enable` is True, it concatenates two pieces before passing them.

4. **Return Value**:
   The function returns a response created using `manage.get_base_to_response()`.
   ```python
return manage.get_base_to_response().to_block_response(str(chat_id), str(chat_record_id),
                                                       paragraph_list, problem_text,
                                                       chat_result.content, reasoning_content=reasoning_content if reasoning_content_enable else '')
```
This part looks correct and follows expected structure for response generation.

Overall, the code does not present significant bugs or room for improvement at this point, assuming all imported modules and necessary contexts are correctly set up during execution.

PR #2464, opened 2025-03-03
fix: Fix default_value not show when image model changed

shaohuzhang1:
fix: Fix default_value not show when image model changed  --bug=1052695 --user=刘瑞斌 【应用】高级编排-设置-图片生成组件-切换模型后图片尺寸为空-未同步切换到当前模型的首个尺寸 https://www.tapd.cn/57709429/s/1661346

### DIFF
--- a/ui/src/components/dynamics-form/index.vue
+++ b/ui/src/components/dynamics-form/index.vue
@@ -179,7 +179,12 @@ const render = (
     const value = formFieldList.value
       .map((item) => {
         if (form_data[item.field] !== undefined) {
-          return { [item.field]: form_data[item.field] }
+          const v: any = item.option_list?.filter(i => i.value_field === form_data[item.field])
+          if (v?.length > 0) {
+            return { [item.field]: form_data[item.field] };
+          } else {
+            return { [item.field]: item.default_value };
+          }
         }
         if (item.show_default_value === true || item.show_default_value === undefined) {
           return { [item.field]: item.default_value }


ui/src/components/dynamics-form/index.vue
```diff
@@ -179,7 +179,12 @@ const render = (
...
-          return { [item.field]: form_data[item.field] }
+          const v: any = item.option_list?.filter(i => i.value_field === form_data[item.field])
+          if (v?.length > 0) {
+            return { [item.field]: form_data[item.field] };
+          } else {
+            return { [item.field]: item.default_value };
+          }
         }
         if (item.show_default_value === true || item.show_default_value === undefined) {
           return { [item.field]: item.default_value }
```
shaohuzhang1, line 190: The provided code has a few issues that can be addressed:

1. **Type Safety**: The type assertion `const v: any` is unnecessary and could lead to runtime errors.
2. **Logic Simplification**: Instead of filtering for an exact match, you might want to consider using strict equality (`===`) when comparing values.

Here's the revised version of the code with these improvements:

```typescript
const render = (...):
 // ...

if (form_data[item.field] !== undefined && form_field_list.item.option_list.some(i => i.value_field === form_data[item.field])) {
  return { [item.field]: form_data[item.field] };
} else if (item.show_default_value === true || item.show_default_value === undefined) {
  return { [item.field]: item.default_value }
}

// ...
```

### Key Changes:
1. **Removed Type Assertion**: Removed `any` type assertion; typescript will infer the correct type based on conditions.
2. **Simplified Logic**: Used `.some()` to check if there is at least one option with a matching `value_field`, rather than `.filter()`, which would require handling multiple matches separately.

This should make the code cleaner and more efficient while maintaining correctness.

PR #2475, opened 2025-03-04
fix: csvTemplate error

shaohuzhang1:
fix: csvTemplate error  --bug=1052711 --user=王孝刚 【知识库】英文模式-文档-QA问答对-下载的csv模版内容错误 https://www.tapd.cn/57709429/s/1661565

### DIFF
--- a/apps/dataset/serializers/document_serializers.py
+++ b/apps/dataset/serializers/document_serializers.py
@@ -248,7 +248,7 @@ class DocumentSerializers(ApiMixin, serializers.Serializer):
                     "rb")
                 content = file.read()
                 file.close()
-                return HttpResponse(content, status=200, headers={'Content-Type': 'text/cxv',
+                return HttpResponse(content, status=200, headers={'Content-Type': 'text/csv',
                                                                   'Content-Disposition': 'attachment; filename="csv_template.csv"'})
             elif self.data.get('type') == 'excel':
                 file = open(os.path.join(PROJECT_DIR, "apps", "dataset", 'template',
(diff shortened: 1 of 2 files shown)


apps/dataset/serializers/document_serializers.py
```diff
@@ -248,7 +248,7 @@ def export(self, with_valid=True):
                     "rb")
                 content = file.read()
                 file.close()
-                return HttpResponse(content, status=200, headers={'Content-Type': 'text/cxv',
+                return HttpResponse(content, status=200, headers={'Content-Type': 'text/csv',
                                                                   'Content-Disposition': 'attachment; filename="csv_template.csv"'})
             elif self.data.get('type') == 'excel':
                 file = open(os.path.join(PROJECT_DIR, "apps", "dataset", 'template',
```
shaohuzhang1, line 254: The code you've provided is essentially exporting data from an application. However, there's a small mistake in the `HttpResponse` header:

1. **Mistake in MIME Type**:
   ```
   'Content-Type': 'text/csv'
   ```

This should be `'application/vnd.ms-excel'` if you're exporting Excel files.

2. **Correcting the Header**:
   
Here's the corrected version of your code:

```python
    if self.data.get('type') == 'csv':
        file = open(os.path.join(PROJECT_DIR, "apps", "dataset", 'template', 
                            f'template_{self.uuid}.csv'), 'rb')
        content = file.read()
        file.close()
        return HttpResponse(content, status=200, headers={
            'Content-Type': 'application/csv',
            'Content-Disposition': f'attachment; filename="template_{self.uuid}.csv"'
        })
    elif self.data.get('type') == 'excel':
        # Your existing code to handle Excel export
```

### Optimizations Suggested:

1. **Error Handling**:
   Add error handling to manage cases where the CSV/template files might not exist or cannot be read due to permissions.
   
2. **Data Validation**:
   Ensure that the `data['type']` field is validated and only allows recognized types (CSV/Excel) before proceeding with the export logic.

3. **Variable Naming**:
   Consider using more descriptive variable names to improve readability.

4. **UUID Generation**:
   Implement generation of unique identifiers (`self.uuid`) dynamically instead of hardcoding it.

These suggestions ensure robustness and maintainability of the code.

PR #2477, opened 2025-03-04
fix: Fixed the default value of not selecting 'select' after switching models

shaohuzhang1:
fix: Fixed the default value of not selecting 'select' after switching models

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -417,7 +417,7 @@ const handleScrollTop = ($event: any) => {
   scrollTop.value = $event.scrollTop
   if (
     dialogScrollbar.value.scrollHeight - (scrollTop.value + scrollDiv.value.wrapRef.offsetHeight) <=
-    30
+    40
   ) {
     scorll.value = true
   } else {
--- a/ui/src/views/application/component/AIModeParamSettingDialog.vue
+++ b/ui/src/views/application/component/AIModeParamSettingDialog.vue
@@ -6,6 +6,7 @@
     v-model="dialogVisible"
     style="width: 550px"
     append-to-body
+    destroy-on-close
     :close-on-click-modal="false"
     :close-on-press-escape="false"
   >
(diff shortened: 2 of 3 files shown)


ui/src/components/ai-chat/index.vue
```diff
@@ -417,7 +417,7 @@ const handleScrollTop = ($event: any) => {
   scrollTop.value = $event.scrollTop
   if (
     dialogScrollbar.value.scrollHeight - (scrollTop.value + scrollDiv.value.wrapRef.offsetHeight) <=
-    30
+    40
   ) {
     scorll.value = true
   } else {
```
shaohuzhang1, line 423: The code you provided has a minor issue that can be improved for clarity and efficiency. Specifically, there seems to be a typographical error in the line where `scorll` is set to `true`. It looks like it should have `scroll.value` instead.

Additionally, while your changes from 30 to 40 make sense (likely increasing the threshold slightly), this change might need further justification based on context specific to your application. Here's the corrected version of the last three lines:

```javascript
if (
  dialogScrollbar.value.scrollHeight - (scrollTop.value + scrollDiv.value.wrapRef.offsetHeight) <=
  40 // Increase this value as needed
) {
  scroll.value = true; // Use the correct property name here
} else {
```

This correction ensures consistent use of `scroll.value`, which appears to be an attribute used elsewhere in your function. If you intended to increase the sensitivity of when showing more content, consider adjusting the threshold beyond what is shown initially.

ui/src/views/application/component/AIModeParamSettingDialog.vue
```diff
@@ -6,6 +6,7 @@
     v-model="dialogVisible"
     style="width: 550px"
     append-to-body
+    destroy-on-close
     :close-on-click-modal="false"
     :close-on-press-escape="false"
   >
```
shaohuzhang1, line 12: The `destroy-on-close` attribute is not necessary for this context since it will cause the modal to be destroyed when closed, which may impact its performance if used frequently. Consider removing this line for better performance and clarity. The other attributes look correctly configured for a dialog component.

PR #2480, opened 2025-03-04
fix: The page cannot continue streaming response when changing the conversation name in the conversation reply

shaohuzhang1:
fix: The page cannot continue streaming response when changing the conversation name in the conversation reply

### DIFF
--- a/ui/src/views/chat/embed/index.vue
+++ b/ui/src/views/chat/embed/index.vue
@@ -181,7 +181,10 @@ function editName(val: string, item: any) {
     }
 
     log.asyncPutChatClientLog(applicationDetail.value.id, item.id, obj, loading).then(() => {
-      getChatLog(applicationDetail.value.id)
+      const find = chatLogData.value.find((item: any) => item.id == item.id)
+      if (find) {
+        find.abstract = val
+      }
       item['writeStatus'] = false
     })
   } else {
--- a/ui/src/views/chat/pc/EditTitleDialog.vue
+++ b/ui/src/views/chat/pc/EditTitleDialog.vue
@@ -75,7 +75,7 @@ const submit = async (formEl: FormInstance | undefined) => {
   await formEl.validate((valid) => {
     if (valid) {
       log.asyncPutChatClientLog(applicationId.value, chatId.value, form.value, loading).then(() => {
-        emit('refresh')
+        emit('refresh', chatId.value, form.value.abstract)
         dialogVisible.value = false
       })
     }
--- a/ui/src/views/chat/pc/index.vue
+++ b/ui/src/views/chat/pc/index.vue
@@ -232,8 +232,11 @@ function mouseenter(row: any) {
 function editLogTitle(row: any) {
   EditTitleDialogRef.value.open(row, applicationDetail.value.id)
 }
-function refreshFieldTitle() {
-  getChatLog(applicationDetail.value.id)
+function refreshFieldTitle(chatId: string, abstract: string) {
+  const find = chatLogData.value.find((item: any) => item.id == chatId)
+  if (find) {
+    find.abstract = abstract
+  }
 }
 function deleteLog(row: any) {
   log.asyncDelChatClientLog(applicationDetail.value.id, row.id, left_loading).then(() => {


ui/src/views/chat/pc/index.vue
```diff
@@ -232,8 +232,11 @@ function mouseenter(row: any) {
...
-function refreshFieldTitle() {
-  getChatLog(applicationDetail.value.id)
+function refreshFieldTitle(chatId: string, abstract: string) {
+  const find = chatLogData.value.find((item: any) => item.id == chatId)
+  if (find) {
+    find.abstract = abstract
+  }
 }
 function deleteLog(row: any) {
   log.asyncDelChatClientLog(applicationDetail.value.id, row.id, left_loading).then(() => {
```
shaohuzhang1, line 242: The code looks mostly correct, but here are some minor suggestions and comments for improvement:

1. **Function Naming Consistency**: Consider renaming `getChatLog` to something more descriptive, such as `fetchChatLogs`, especially if used elsewhere.

2. **Parameters and Return Types**: Ensure that parameters and return types in functions match their intended use cases. In the `refreshFieldTitle` function, you have multiple uses of `chatId`, which could be cleaner with a single parameter.

3. **Variable Reuse**: The variable `find` is only used once before being reassigned without additional usage. It's better to directly access `chatLogData.value.find(...)` when needed instead.

4. **Optional Parameters**: If `applicationDetail.value.id` might not always exist, consider using optional chaining (`?.`) to avoid errors during execution.

Here’s an updated version of the code incorporating these suggestions:

```javascript
@@ -232,8 +232,10 @@ function mouseenter(row: any) {
 }

 function editLogTitle(row: any) {
   EditTitleDialogRef.value.open(row, applicationDetail.value?.id ?? undefined);
 }
-function refreshFieldTitle() {
-  getChatLog(applicationDetail.value.id);
+function refreshFieldTitle(chatId: string | null = null, abstract?: string) {
+  if (chatId !== null && !!abstract) {
+    const find = chatLogData.value.find(item => item.id === chatId);
+    if (find) {
+      find.abstract = abstract;
+    }
+  }
 }
 
 function deleteLog(row: any) {
   log.asyncDelChatClientLog(
     applicationDetail.value?.id ?? undefined,
     row.id,
     left_loading
   ).then(() => {
```

By applying these changes, the code becomes more robust, follows consistent naming conventions, reduces redundancy, and improves readability.

ui/src/views/chat/embed/index.vue
```diff
@@ -181,7 +181,10 @@ function editName(val: string, item: any) {
...
 
     log.asyncPutChatClientLog(applicationDetail.value.id, item.id, obj, loading).then(() => {
-      getChatLog(applicationDetail.value.id)
+      const find = chatLogData.value.find((item: any) => item.id == item.id)
+      if (find) {
+        find.abstract = val
+      }
       item['writeStatus'] = false
     })
   } else {
```
shaohuzhang1, line 190: The code does not appear to have obvious syntax errors or logical inconsistencies. However, there is room for optimization and improvements:

1. The `find` method can be replaced with a more direct approach using `[].filter`, which is generally faster:
```javascript
const find = chatLogData.value.filter(item => item.id === item.id);
```

2. Ensure that `chatLogData.value` contains data before attempting to filter items. Adding type checking might further secure the code.

3. If `obj` has properties other than `abstract` you need to decide what to do about those properties since they are being discarded when updating the `item`.

These changes aim to make the code cleaner without affecting functionality significantly.

ui/src/views/chat/pc/EditTitleDialog.vue
```diff
@@ -75,7 +75,7 @@ const submit = async (formEl: FormInstance | undefined) => {
   await formEl.validate((valid) => {
     if (valid) {
       log.asyncPutChatClientLog(applicationId.value, chatId.value, form.value, loading).then(() => {
-        emit('refresh')
+        emit('refresh', chatId.value, form.value.abstract)
         dialogVisible.value = false
       })
     }
```
shaohuzhang1, line 81: There are no significant issues with the code. Here are a couple of minor optimizations:

  1. The `asyncPutChatClientLog` call is being made without awaiting its completion before emitting the refresh event. This can lead to race conditions where the log might not have been updated yet when the refresh event is emitted. If it's safe to assume that the log has been successfully added after calling this function, you could wrap the `emit('refresh'...)` statement inside an `await`.
  2. You're passing both the `chatId` and `form.value.abstract` arguments to the `emit` method, but none of them are used within the component. It may be worth removing one of these parameters if they are unnecessary.

Here is the revised code:
```sql
  await formEl.validate((valid) => {
    if (valid) {
      loading.value = true; // Set loading indicator to true

      try {
        await log.asyncPutChatClientLog(applicationId.value, chatId.value, form.value);
      } catch (error) {
        console.error("Failed to update chat client log", error); // Log any errors
      }

      emit('refresh', chatId.value);

      dialogVisible.value = false;
    }
    loading.value = false; // Set loading indicator to false
  });
};
```
With these changes, the code becomes more robust and handles potential errors gracefully. However, it remains otherwise unchanged from your original version.

PR #2483, opened 2025-03-04
refactor: add model_name parameter to AzureChatModel initialization

shaohuzhang1:
refactor: add model_name parameter to AzureChatModel initialization  --bug=1052553 --user=刘瑞斌 【模型】添加 Azure OpenAI 添加DeepSeek -R1 模型报错 https://www.tapd.cn/57709429/s/1661953

### DIFF
--- a/apps/setting/models_provider/impl/azure_model_provider/model/azure_chat_model.py
+++ b/apps/setting/models_provider/impl/azure_model_provider/model/azure_chat_model.py
@@ -27,6 +27,7 @@ class AzureChatModel(MaxKBBaseModel, AzureChatOpenAI):
 
         return AzureChatModel(
             azure_endpoint=model_credential.get('api_base'),
+            model_name=model_name,
             openai_api_version=model_credential.get('api_version', '2024-02-15-preview'),
             deployment_name=model_credential.get('deployment_name'),
             openai_api_key=model_credential.get('api_key'),


apps/setting/models_provider/impl/azure_model_provider/model/azure_chat_model.py
```diff
@@ -27,6 +27,7 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
 
         return AzureChatModel(
             azure_endpoint=model_credential.get('api_base'),
+            model_name=model_name,
             openai_api_version=model_credential.get('api_version', '2024-02-15-preview'),
             deployment_name=model_credential.get('deployment_name'),
             openai_api_key=model_credential.get('api_key'),
```
shaohuzhang1, line 33: The provided code is mostly correct, but there are a couple of improvements that can be made:

1. **Missing Import Statement**: The line `from azure.core.credentials import AzureKeyCredential` is missing from the code snippet. This import statement is necessary to use `AzureKeyCredential`, which is returned along with `get_token_for_open_ai`.

2. **Initialization with Credentials**: Instead of passing all credentials directly as key-value pairs, it might be more organized to pack them into an instance attribute (`self.credential`). This makes the `new_instance` method cleaner and reduces the risk of forgetting to pass a credential.

Here's an optimized version of the `__init__` method incorporating these changes:

```python
class AzureChatModel:
    def __init__(self, azure_endpoint=None, deployment_name=None, model_name=None):
        # Initialization takes only essential parameters
        self.model_name = model_name

        # These could ideally come from a configuration file or user input
        self.azure_endpoint = azure_endpoint
        self.deployment_name = deployment_name

# In your new_instance function usage
credential = get_token_for_open_ai()  # Ensure this function returns an instance of AzureKeyCredentials
model = AzureChatModel(azure_endpoint=credentials.tenant_id, deployment_name=models[0].name, model_name='gpt-3')
```

This approach enhances reusability and maintainability by clearly organizing each parameter's responsibility in the constructor while maintaining simplicity at run time.

PR #2485, opened 2025-03-04
fix: The newly created simple application dialogue process label is not effective

shaohuzhang1:
fix: The newly created simple application dialogue process label is not effective

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -142,10 +142,12 @@ class ModelSettingSerializer(serializers.Serializer):
                                                  error_messages=ErrMessage.char(_("No citation segmentation prompt")))
     reasoning_content_enable = serializers.BooleanField(required=False,
                                                         error_messages=ErrMessage.char(_("Thinking process switch")))
-    reasoning_content_start = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=256,
+    reasoning_content_start = serializers.CharField(required=False, allow_null=True, default="<think>",
+                                                    allow_blank=True, max_length=256,
                                                     error_messages=ErrMessage.char(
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_serializers.py
```diff
@@ -142,10 +142,12 @@ class ModelSettingSerializer(serializers.Serializer):
...
                                                         error_messages=ErrMessage.char(_("Thinking process switch")))
-    reasoning_content_start = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=256,
+    reasoning_content_start = serializers.CharField(required=False, allow_null=True, default="<think>",
+                                                    allow_blank=True, max_length=256,
                                                     error_messages=ErrMessage.char(
                                                         _("The thinking process begins to mark")))
-    reasoning_content_end = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=256,
+    reasoning_content_end = serializers.CharField(required=False, allow_null=True, allow_blank=True, default="</think>",
+                                                  max_length=256,
                                                   error_messages=ErrMessage.char(_("End of thinking process marker")))
```
shaohuzhang1, line 153: There is an issue with the `reasoning_content_start` field: it now uses `<think>` by default instead of `{"start": "<think>"}`. This could lead to unexpected behavior if this value were required at runtime or if other parts of your application expect `{"start": "<think>"}` directly.

To resolve this, you should change the default value back to `{"start": "<think>"}`:

```python
        reasoning_content_start = serializers.CharField(required=False, allow_null=True, 
                                                      default={"start": "<think>"},
                                                       allow_blank=True, max_length=256,
                                                      error_messages=ErrMessage.char(_("The thinking process begins to mark")))
```

Also, ensure there are no similar issues with other fields that might have default values affecting expected data integrity.

PR #2489, opened 2025-03-04
fix: Enterprise WeChat docking sub application cannot output thinking process

shaohuzhang1:
fix: Enterprise WeChat docking sub application cannot output thinking process

### DIFF
--- a/apps/application/flow/step_node/application_node/impl/base_application_node.py
+++ b/apps/application/flow/step_node/application_node/impl/base_application_node.py
@@ -115,6 +115,10 @@ def write_context(node_variable: Dict, workflow_variable: Dict, node: INode, wor
                                          'prompt_tokens': response.get('prompt_tokens')}}
     answer = response.get('content', '') or "抱歉，没有查找到相关内容，请重新描述您的问题或提供更多信息。"
     reasoning_content = response.get('reasoning_content', '')
+    answer_list = response.get('answer_list', [])
+    node_variable['application_node_dict'] = {answer.get('real_node_id'): {**answer, 'index': index} for answer, index
+                                              in
+                                              zip(answer_list, range(len(answer_list)))}
     _write_context(node_variable, workflow_variable, node, workflow, answer, reasoning_content)
 
 
--- a/apps/common/handle/impl/response/openai_to_response.py
+++ b/apps/common/handle/impl/response/openai_to_response.py
@@ -26,7 +26,7 @@ class OpenaiToResponse(BaseToResponse):
             other_params = {}
         data = ChatCompletion(id=chat_record_id, choices=[
             BlockChoice(finish_reason='stop', index=0, chat_id=chat_id,
-                        reasoning_content=other_params.get('reasoning_content', ""),
+                        answer_list=other_params.get('answer_list', ""),
                         message=ChatCompletionMessage(role='assistant', content=content))],
                               created=datetime.datetime.now().second, model='', object='chat.completion',
                               usage=CompletionUsage(completion_tokens=completion_tokens,


apps/application/flow/step_node/application_node/impl/base_application_node.py
```diff
@@ -115,6 +115,10 @@ def write_context(node_variable: Dict, workflow_variable: Dict, node: INode, wor
                                          'prompt_tokens': response.get('prompt_tokens')}}
     answer = response.get('content', '') or "抱歉，没有查找到相关内容，请重新描述您的问题或提供更多信息。"
     reasoning_content = response.get('reasoning_content', '')
+    answer_list = response.get('answer_list', [])
+    node_variable['application_node_dict'] = {answer.get('real_node_id'): {**answer, 'index': index} for answer, index
+                                              in
+                                              zip(answer_list, range(len(answer_list)))}
     _write_context(node_variable, workflow_variable, node, workflow, answer, reasoning_content)
```
shaohuzhang1, line 124: Here are the identified issues and potential improvements:

### Issues:
1. **Variable Access**: 
   - The line `node_variable['application_node_dict']` is accessing a variable named `response`, which does not exist at that stage of the function execution. This will likely raise an error.

2. **Answer List Handling**:
   - The line `${{answer.get('real_node_id')` contains interpolation syntax, but it's missing the opening brace `{`.
   - The logic to update `node_variable['application_node_dict']` seems incorrect because both keys (`"real_node_id"` and `"index"`) should be included in each entry.

3. **Function Parameters**:
   - The last parameter passed to `_write_context()` is `workflow`, but there is no such parameter defined in the original code snippet you provided.

### Potential Improvements:
1. **Correct Variable Name**:
   - Ensure that the `response` variable is correctly referenced if needed or removed if unnecessary.

2. **Fix Template Syntax**:
   - Replace `${{answer.get('real_node_id')}}` with simply `answer.get('real_node_id')`.

3. **Re-evaluate Data Structure**:
   - Review what data structure is intended for `node_variable['application_node_dict']`. If you mean to store different types of data under `'actual_node_id'`, adjust the dictionary accordingly.

4. **Parameter Correctness**:
   - Double-check that all parameters passed to `_write_context()` match those expected, and add any necessary parameters that might have been omitted from the previous version.

5. **Error Handling**:
   - Consider adding error handling to manage cases where certain fields like `'prompt_tokens', 'content', ..., '` might not always be available in the `response`.

Without additional context on how these variables interact, some further refinement may be required.

apps/common/handle/impl/response/openai_to_response.py
```diff
@@ -26,7 +26,7 @@ def to_block_response(self, chat_id, chat_record_id, content, is_end, completion
             other_params = {}
         data = ChatCompletion(id=chat_record_id, choices=[
             BlockChoice(finish_reason='stop', index=0, chat_id=chat_id,
-                        reasoning_content=other_params.get('reasoning_content', ""),
+                        answer_list=other_params.get('answer_list', ""),
                         message=ChatCompletionMessage(role='assistant', content=content))],
                               created=datetime.datetime.now().second, model='', object='chat.completion',
                               usage=CompletionUsage(completion_tokens=completion_tokens,
```
shaohuzhang1, line 32: The provided code snippet seems mostly correct, but there are a few minor adjustments and clarifications that can be made to ensure it works as intended:

1. **Use of `get` method correctly**: In Python dictionaries, using the `get()` method with a default parameter (`other_params.get('reasoning_content', "")`) will handle cases where 'reasoning_content' is not present in `other_params`. This prevents a potential KeyError.

2. **Correct import statement for datetime module**: Ensure you have imported the `datetime` module at the beginning of your file if it's not already imported. The current line imports from both `pandas` and `pyarrow`, so I assume `datetime` is used elsewhere; however, here we just need to make sure it's available.

3. **Consistency in handling None values**: When accessing dictionary keys, always check for their existence before proceeding to access them to avoid NoneType errors.

Here’s the code with these points addressed:

```python
import datetime

def to_block_response(self, chat_id, chat_record_id, content, is_end, completion_tokens, other_params):
    # Using get() to safely retrieve reasoning_content from other_params
    reasoning_content = other_params.get('reasoning_content', "")
    
    # Similarly, use get() to safely retrieve answer_list from other_params
    answer_list = other_params.get('answer_list', [])
    
    data = ChatCompletion(
        id=chat_record_id,
        choices=[
            BlockChoice(
                finish_reason='stop',
                index=0,
                chat_id=chat_id,
                reasoning_content=reasoning_content,
-                         answer_list answer_list,
                 message=ChatCompletionMessage(role='assistant', content=content)
            )
        ],
        created=datetime.datetime.now().timestamp(),
        model='',
        object='chat.completion',
        usage=CompletionUsage(completion_tokens=completion_tokens)
    )

# Example usage with an empty list for completeness
example_other_params = {}
to_block_response("id_123", "record_1", "Hello!", True, 20, example_other_params)
```

This should prevent any runtime errors related to missing keys in the `other_params` dictionary and maintain proper consistency in value fetching.

PR #2499, opened 2025-03-05
fix: Model addition, error log printing

shaohuzhang1:
fix: Model addition, error log printing

### DIFF
--- a/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt.py
+++ b/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-
+import traceback
 from typing import Dict
 
 from django.utils.translation import gettext as _
@@ -30,6 +30,7 @@ class AliyunBaiLianSTTModelCredential(BaseForm, BaseModelCredential):
             model = provider.get_model(model_type, model_name, model_credential)
             model.check_auth()
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
--- a/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/tts.py
+++ b/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/tts.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-
+import traceback
 from typing import Dict
 
 from django.utils.translation import gettext_lazy as _, gettext
@@ -64,6 +64,7 @@ class AliyunBaiLianTTSModelCredential(BaseForm, BaseModelCredential):
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
--- a/apps/setting/models_provider/impl/gemini_model_provider/credential/image.py
+++ b/apps/setting/models_provider/impl/gemini_model_provider/credential/image.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-
+import traceback
 from typing import Dict
 
 from django.utils.translation import gettext_lazy as _, gettext
@@ -52,6 +52,7 @@ class GeminiImageModelCredential(BaseForm, BaseModelCredential):
             for chunk in res:
                 print(chunk)
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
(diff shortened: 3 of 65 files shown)


apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt.py
```diff
@@ -30,6 +30,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
             model = provider.get_model(model_type, model_name, model_credential)
             model.check_auth()
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
```
shaohuzhang1, line 36: Your code looks mostly clean and well-organized, but there are a few potential improvements:

1. **Exception Handling**: The `traceback.print_exc()` function is good for debugging purposes, especially when you're working remotely or in an environments where logging might not be set up properly. However, I would recommend adding more detail to the exception handling logic.

Here's one way it could look after making these changes:

```python
        except Exception as e:
            print(f"An error occurred while checking authentication: {e}")
            
            # Detailed traceback
            traceback.print_exception(*sys.exc_info())
            
                raise e
            
            if raise_exception:
```

This will log both the exception message and the full stack trace, which can help in tracking down errors effectively.

---

2. **Consistency with Code Style**: 
   - Ensure consistent use of quotes (either single or double) throughout your file.
   - If you want to keep using triple quotes (`"""`), make sure they are appropriately formatted across the entire class/module.

---

3. **Security Considerations**:
   - When raising exceptions in production, consider wrapping them with custom `Http404`, `PermissionDenied`, etc., rather than bare `raise`.
   
Example:
```python
if raise_exception:
    raise Http404("Invalid model credentials")
```
or
```python
if raise_exception:
    raise PermissionDenied("Insufficient permissions to access this resource")
```
These exceptions have HTTP status codes that align better with typical RESTful API usage patterns.

---

Overall, the code structure is clear and does what you intended it to do. With these minor adjustments, you'll have cleaner, more detailed error handling and potentially improved readability.

apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/tts.py
```diff
@@ -64,6 +64,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
```
shaohuzhang1, line 70: The changes made in this code snippet include:

1. Adding the `traceback` import at the top of the file, which will print a stack trace when an exception occurs during execution.

2. Moving the call to `provider.get_model()` inside a try-except block, so that any exceptions raised by this operation can be caught and reported appropriately.

This addition provides better error handling for cases where errors occur while trying to retrieve models from the provider. It's generally good practice to catch exceptions explicitly instead of leaving them to propagate all the way up the call stack, allowing you to handle specific types of errors more gracefully and maintain application stability. Additionally, printing the traceback can help developers debug issues by providing context about what went wrong during execution.

apps/setting/models_provider/impl/gemini_model_provider/credential/image.py
```diff
@@ -52,6 +52,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
             for chunk in res:
                 print(chunk)
         except Exception as e:
+            traceback.print_exc()
             if isinstance(e, AppApiException):
                 raise e
             if raise_exception:
```
shaohuzhang1, line 58: The provided code snippet appears to be written in Python using Django framework. Here are some points of review:

1. **Imports**: The `traceback` module is imported at the beginning, which is appropriate for capturing and printing stack traces of errors.

2. **Variable Names**: Variable names like `model_type`, `model_name`, `model_credential`, etc., suggest they might represent database table columns. However, without more context, it's hard to confirm their usage patterns.

3. **Error Handling**: 
   - There is an attempt to catch exceptions within a try-except block when processing results (`res`) after executing a query or API call.
   - After catching an exception, if it's an instance of `AppApiException`, the exception is re-raised.
   - If `raise_exception=True` (not shown here), then the function would return early if the inner exception was caught.
   
4. **Traceback Printing**:
   - `traceback.print_exc()` is called unconditionally when an exception occurs during querying or API calls. This can help in understanding what went wrong but may also clutter the stdout if logging is configured elsewhere.
   - Consider using a more localized logger instead of relying on print statements for error handling.

5. **Code Structure**:
   - While not overly complex, the code structure follows typical functional programming idioms used in Django apps.

6. **Potential Issues**:
   - Ensure that if there are any exceptions due to missing data or invalid credentials, those scenarios should be gracefully handled rather than just raising them unless `raise_exception=True`.
   - Verify that all operations involve parameterized queries or sanitization to prevent SQL injection attacks unless explicitly allowed (e.g., through Django ORM methods).

7. **Optimization Suggestions**:
   - Ensure that no redundant calculations or unnecessary loops are present.
   - Look into optimizing I/O-bound operations by caching results where feasible, depending on the application's requirements.

Overall, the code looks well-structured with proper error handling in place. It would benefit from additional comments and possibly adjustments to improve readability and maintainability.

Issue #2467, opened 2025-03-04
[Bug] I hope that the form collection component can be optimized when displaying it on the mobile phone


linfengjjj:
### MaxKB Version

v1.10.2-LTS

### Please describe your needs or suggestions for improvements

希望表单收集组件在手机端显示可以优化下，目前使用选项卡，卡显示名称较长时，无法完整显示

<img width="201" alt="Image" src="https://github.com/user-attachments/assets/9149627a-a19e-4c6a-b9e1-c3e1c624467d" />

baixin513:
感谢反馈，后续版本修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version fixes

baixin513:
v1.10.2-lts已发布。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.2-lts has been released.

PR #2501, opened 2025-03-05
perf: the form collection component optimized when displaying it on t…

wangdan-fit2cloud:
…he mobile phone(#2467)

### DIFF
--- a/ui/src/components/dynamics-form/items/radio/RadioCard.vue
+++ b/ui/src/components/dynamics-form/items/radio/RadioCard.vue
@@ -5,8 +5,9 @@
         <el-col :xs="24" :sm="24" :md="24" :lg="12" :xl="12">
           <el-card
             :key="item.value"
-            class="item"
+            class="item break-all"
             shadow="never"
+            style="--el-card-padding: 12px 16px"
             :class="[
               inputDisabled ? 'is-disabled' : '',
               modelValue == item[valueField] ? 'active' : ''
@@ -93,8 +94,8 @@ const option_list = computed(() => {
     color: var(--el-color-primary);
   }
   .item {
+    line-height: 22px;
     cursor: pointer;
-    height: 38px;
     display: flex;
     justify-content: center;
     align-items: center;
--- a/ui/src/styles/element-plus.scss
+++ b/ui/src/styles/element-plus.scss
@@ -419,6 +419,6 @@
   color: var(--el-checkbox-text-color);
 }
 
-.el-popper {
+.el-table .el-popper {
   max-width: 500px !important;
 }


ui/src/components/dynamics-form/items/radio/RadioCard.vue
```diff
@@ -93,8 +94,8 @@ const option_list = computed(() => {
     color: var(--el-color-primary);
   }
   .item {
+    line-height: 22px;
     cursor: pointer;
-    height: 38px;
     display: flex;
     justify-content: center;
     align-items: center;
```
shaohuzhang1, line 101: There are no significant irregularities or major issues with the provided code. However, there are a few minor improvements you can consider:

1. **Remove Duplicated Comments**: The comments `/* */` at positions -5 to 9 do not add any value and should be removed to clean up the code.

2. **Use Consistent Naming Conventions**: Ensure that all variable names and class names use consistent naming conventions (e.g., camelCase).

3. **Optimize `break-all` Class Usage**: If the intention behind using the `break-all` class is to allow text to wrap, it might need more precise control over where it breaks. Consider customizing its CSS if needed.

4. **Consider Dynamic Styling for Height**: While specifying a static height (`height: 38px`) is usually sufficient, dynamically setting the height based on other properties like padding could make sense depending on layout requirements.

Here's an optimized version of the relevant section of the code:

```vue
<template>
  <el-col :xs="24" :sm="24" :md="24" :lg="12" :xl="12">
    <el-card
      v-if="option_list.length > 0"
      v-model:model-value="modelValue"
      key="item.value"
      class="item break-all"
      shadow="never"
      style="--el-card-padding: 12px 16px; --el-dialog-footer-width: auto;"
      @click.stop="$emit('update:modelValue', item[valueField])"
    >
      {{ item[labelField] }}
    </el-card>
  </el-col>
</template>

<script setup lang="ts">
import { ref } from "vue";
interface Item {
  [valueField]: string | number;
  [labelField]: string;
}

const props = defineProps<{
  option_list: Array<Item>;
  modelValue?: string | number;
  valueField: string; // Assuming this represents the field used for comparison in options
  labelField: string; // Assuming this represents the display text for each option
}>();

const emit = defineEmits(["update:modelValue"]);

// ... rest of the component logic ...
</script>

<style scoped>
.item {
  line-height: 22px;
  overflow-wrap: break-word; /* Ensures long words break */
}
</style>
```

### Notes:
- I've added a dynamic event listener for clicking on the card to update the `modelValue`.
- Used `overflow-wrap: break-word` instead of just `line-height` for better word breaking behavior within the box.
- Ensured the `--el-dialog-footer-width: auto;)` rule adjusts according to parent container constraints. Adjust as necessary for specific design needs.

These changes enhance readability and functionality without introducing new bugs.

ui/src/styles/element-plus.scss
```diff
@@ -419,6 +419,6 @@
   color: var(--el-checkbox-text-color);
 }
 
-.el-popper {
+.el-table .el-popper {
   max-width: 500px !important;
 }
\ No newline at end of file
```
shaohuzhang1, line 424: The code has no obvious issues but can be slightly optimized:

**Optimization Suggestions:**
1. **Consistency with Class Names:** Ensure that similar classes share common styles where possible to avoid repetition and make maintainability easier.

2. **Avoid Inline Styles for Width:** If you plan on changing the `max-width` frequently, consider defining it in a CSS file instead of using an inline style. This makes changes easier to manage.

Example:
```css
/* In a separate stylesheet */
.el-table .el-popper {
  max-width: var(--table-tooltip-max-width); /* Define this variable in your theme */
}
```

3. **Add Comment Descriptions:** Provide descriptions for complex or unconventional selectors to make future modifications clearer.

By applying these improvements, you enhance readability and flexibility in managing the styling.

Issue #2523, opened 2025-03-10
[Bug] Replace the avatar in the application and the modification was successful, but the prompt message is incorrect


spritelyl:
### Contact Information

14707070227

### MaxKB Version

v1.10.2-lts

### Problem Description

应用里面替换头像，修改成功，但是提示信息不正确

### Steps to Reproduce

1. 应用里面替换头像，修改成功，但是提示信息不正确

![Image](https://github.com/user-attachments/assets/da18f41b-4978-4bef-bde3-208d3f8bfaab)

### The expected correct result

应用里面替换头像，修改成功，提示正确展示

### Related log output

```shell

```

Shenguobin0102:
你好，没有复现你描述的问题，请再确认一下问题是否存在。

baixin513:
V1.10.3-lts版本已发布，请更新试用！

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
V1.10.3-lts version has been released, please update and try it out!

PR #2535, opened 2025-03-11
fix: Setting avatar save prompt error(#2523)

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/application-overview/component/EditAvatarDialog.vue
+++ b/ui/src/views/application-overview/component/EditAvatarDialog.vue
@@ -121,7 +121,7 @@ function submit() {
   if (radioType.value === 'default') {
     application.asyncPutApplication(id as string, { icon: defaultIcon }, loading).then((res) => {
       emit('refresh')
-      MsgSuccess(t('views.applicationOverview.appInfo.EditAvatarDialog.setSuccess'))
+      MsgSuccess(t('common.saveSuccess'))
       dialogVisible.value = false
     })
   } else if (radioType.value === 'custom' && iconFile.value) {
@@ -129,7 +129,7 @@ function submit() {
     fd.append('file', iconFile.value.raw)
     overviewApi.putAppIcon(id as string, fd, loading).then((res: any) => {
       emit('refresh')
-      MsgSuccess(t('views.applicationOverview.appInfo.EditAvatarDialog.setSuccess'))
+      MsgSuccess(t('common.saveSuccess'))
       dialogVisible.value = false
     })
   } else {


ui/src/views/application-overview/component/EditAvatarDialog.vue
```diff
@@ -121,15 +121,15 @@ function submit() {
...
   } else if (radioType.value === 'custom' && iconFile.value) {
     let fd = new FormData()
     fd.append('file', iconFile.value.raw)
     overviewApi.putAppIcon(id as string, fd, loading).then((res: any) => {
       emit('refresh')
-      MsgSuccess(t('views.applicationOverview.appInfo.EditAvatarDialog.setSuccess'))
+      MsgSuccess(t('common.saveSuccess'))
       dialogVisible.value = false
     })
   } else {
```
shaohuzhang1, line 135: The code you provided has several potential issues that could be addressed:

1. **Lack of Consistent Error Handling**: There is no error handling in case something goes wrong during API requests (`asyncPutApplication` or `overviewApi.putAppIcon`). This can result in unexpected behavior if the server responds with errors.

2. **Use of Raw File Data**: The line `fd.append('file', iconFile.value.raw)` assumes that `iconFile.value.raw` already contains the raw data from the file input element, which might not always be reliable depending on the browser's implementation. It would be safer to work directly with the file object without assuming its raw property exists.

3. **Redundant Code Blocks**: Both conditions for setting an avatar use similar logic, including appending a form data entry and updating `dialogVisible`. These blocks could be extracted into a separate function to reduce redundancy.

4. **Variable Naming**: Variable names like `defaultIcon`, `radioType`, etc., should be more descriptive if possible, especially since they refer to user-facing elements.

Here’s a modified version of the code addressing some of these points:

```javascript
function saveAvatar() {
    const id = id as string; // Assuming id is defined somewhere in scope

    const uploadImage = async () => {
        try {
            await appApi.uploadAppIcon(id, iconFile.value);
            emit('refresh');
            MsgSuccess(t('common.saveSuccess'));
            dialogVisible.value = false;
        } catch (error) {
            console.error("Failed to set Avatar:", error);
        }
    };

    if (radioType.value === 'default') {
        uploadImage();
    } else if (radioType.value === 'custom' && iconFile.value) {
        uploadImage();
    }
}

// In your main application component, call this function when you want to save the avatar
```

### Key Changes:
- **Function Extraction**: The upload logic is now inside the `saveAvatar` function to prevent duplication.
- **Error Handling**: Added a generic error handler using a promise catch block to ensure any failed uploads are logged.
- **Consistent Error Message**: Replaced the success message specific to the Avatar Dialog with a common "Save Success" message.

This refactored version improves maintainability and robustness by encapsulating the core functionality in one place and providing better feedback on potential failures.

PR #2564, opened 2025-03-14
fix: schema access failure to https

shaohuzhang1:
fix: schema access failure to https


apps/common/config/swagger_conf.py
```diff
@@ -20,10 +21,10 @@ def get_tags(self, operation_keys=None):
...
     def get_schema(self, request=None, public=False):
         schema = super().get_schema(request, public)
-        if request.is_secure():
-            schema.schemes = ['https']
-        else:
-            schema.schemes = ['http']
-        return schema
\ No newline at end of file
+        schema.schemes = ['https', 'http']
+        return schema
```
shaohuzhang1, line 30: No issues found in the provided code. It correctly implements customizing Swagger generation by setting up tag handling and specifying both HTTP schemes ('http' and 'https'). The `CustomOpenAPISchemaGenerator` class is defined to extend `OpenAPISchemaGenerator`, allowing customization of API scheme settings based on HTTPS checks.

apps/smartdoc/settings/base.py
```diff
@@ -86,6 +86,7 @@
 
 SWAGGER_SETTINGS = {
     'DEFAULT_AUTO_SCHEMA_CLASS': 'common.config.swagger_conf.CustomSwaggerAutoSchema',
+    'DEFAULT_GENERATOR_CLASS': 'common.config.swagger_conf.CustomOpenAPISchemaGenerator',
     "DEFAULT_MODEL_RENDERING": "example",
     'USE_SESSION_AUTH': False,
     'SECURITY_DEFINITIONS': {
```
shaohuzhang1, line 92: The provided swagger configuration snippet has no apparent errors or potential issues from a technical standpoint. To suggest optimizations:

1. **Customization of Schema Generation**: By adding `'DEFAULT_GENERATOR_CLASS'` with the value `'common.config.swagger_conf.CustomOpenAPISchemaGenerator'`, you can tailor the Swagger document generation to meet specific needs, such as enhancing readability, formatting, or supporting additional features not directly offered by default.

2. **Security Considerations**: The security settings (`"SECURITY_DEFINITIONS"` are commented out. If security is important, consider defining appropriate mechanisms for accessing resources securely in accordance with your application's architecture.

3. **Model Rendering**: `"DEFAULT_MODEL_RENDERING"` set to `"example"` might be useful for generating models inline with example values. However, this setting does not affect rendering; it only affects the way examples are displayed alongside model schemas when viewing them through tools like Postman or cURL.

The overall structure looks clean and functional for integrating custom configurations into Django Rest Framework (DRF) using Swaggrator.

apps/common/config/swagger_conf.py
```diff
@@ -20,10 +20,10 @@ def get_tags(self, operation_keys=None):
...
```
shaohuzhang1, line 29: The code is mostly correct, but there's one small issue that needs to be addressed:

```python
+tags_dict = {
```

It seems like this line should not be preceded by a plus sign. The `from` statement already imports the necessary modules, so adding an additional '+' is unnecessary.

Here's the corrected version:

```diff
from drf_yasg.generators import OpenAPISchemaGenerator
from drf_yasg.inspectors import SwaggerAutoSchema

# Remove the plus sign before tags_dict
tags_dict = {
```

This change ensures clarity and avoids confusion with syntax characters in Python. Note that your changes look good otherwise!

apps/smartdoc/settings/base.py
shaohuzhang1, line 92: Your code appears to be correctly defined and should work without any immediate issues. However, here a few optimizations and additional checks you might want to consider:

1. **Custom OpenAPI Schema Generator**: Ensure that `common.config.swagger_conf.CustomOpenAPISchemaGenerator` is implemented properly. It's crucial that it extends `openapi_spec_validator.generators.BaseGenerator` and overrides methods like `get_paths()`.

2. **Default Model Rendering**: If `"DEFAULT_MODEL_RENDERING"` is set to `'example'`, this means Swagger/OpenAPI will render examples of model instances. Consider how well these examples meet your needs for documentation clarity.

3. **Security Definitions**: Double-check the `SECURITY_DEFINITIONS` dictionary. Make sure all necessary authentication methods are supported if they're configured.

4. **Error Handling**: Since there's no error handling logic in the snippet provided, ensure that your application has robust error handling for production use cases.

Overall, your approach seems sound, but reviewing these aspects can help improve both functionality and maintainability.

PR #2579, opened 2025-03-17
fix: Log annotation acquisition details file type data cannot be serialized

shaohuzhang1:
fix: Log annotation acquisition details file type data cannot be serialized


apps/common/encoder/encoder.py
```diff
@@ -0,0 +1,28 @@
...
+        if isinstance(obj, uuid.UUID):
+            return str(obj)
+        if isinstance(obj, datetime.datetime):
+            return obj.strftime("%Y-%m-%d %H:%M:%S")
+        if isinstance(obj, decimal.Decimal):
+            return float(obj)
+        if isinstance(obj, InMemoryUploadedFile):
+            return {'name': obj.name, 'size': obj.size}
+        else:
+            return json.JSONEncoder.default(self, obj)
```
shaohuzhang1, line 28: The code you provided has several minor issues that can be addressed:

1. **Imports**: The `InMemoryUploadedFile` class from Django's `files.uploadedfile` module should be imported with an underscore (`_`) before its name because it shadows the built-in `_` function, which is used for variable assignment in some contexts.

Here's the corrected version of the import statement:
```python
from django.core.files.uploadedfile import InMemoryUploadedFile as _InMemoryUploadedFile
```

2. **Class Name**: Class names in Python are typically PascalCase (e.g., `SystemEncoder`). If this is intentional to follow certain coding conventions, keep it unchanged. Otherwise, consider renaming it to `MaxKbSystemEncoder`.

3. **Comments**: Although comments in Chinese are appropriate within your project context, they do not need to include them when sharing the code elsewhere. You might want to remove the existing comments or replace them with relevant ones if necessary.

4. **Documentation**: While the docstring is concise and provides useful information about the class, the author’s initials (`虎`) could be removed as they serve no practical purpose and are unnecessary unless specified otherwise by the organization’s naming convention.

Here's the optimized and cleaned-up version of the code:
```python
# coding=utf-8

"""
    @project: MaxKB
    @author: Tiger
    @file: SystemEncoder.py
    @date: 2025/3/17 16:38
    @desc: Custom JSON encoder for handling Django-specific objects like UUIDs, datetimes, decimals, and InMemoryUploadedFiles.
"""

import datetime
import decimal
import json
import uuid
from django.core.files.uploadedfile import InMemoryUploadedFile as _InMemoryUploadedFile


class SystemEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, uuid.UUID):
            return str(obj)
        if isinstance(obj, datetime.datetime):
            return obj.strftime("%Y-%m-%d %H:%M:%S")
        if isinstance(obj, decimal.Decimal):
            return float(obj)
        if isinstance(obj, _InMemoryUploadedFile):  # Use the renamed imported class
            return {'name': obj.name, 'size': obj.size}
        else:
            return super().default(obj)
```

### Summary:
- Removed redundant imports.
- Renamed the import of `_InMemoryUploadedFile` for clarity.
- Cleaned up the class name (if necessary) according to a standard naming convention.
- Added documentation back to the class level since comments were removed from other parts of the file.
- Ensured consistent use of underscores in method calls where necessary (e.g., using `super()`.

apps/application/models/application.py
```diff
@@ -163,7 +149,7 @@ class ChatRecord(AppModelMixin):
     message_tokens = models.IntegerField(verbose_name="请求token数量", default=0)
     answer_tokens = models.IntegerField(verbose_name="响应token数量", default=0)
     const = models.IntegerField(verbose_name="总费用", default=0)
-    details = models.JSONField(verbose_name="对话详情", default=dict, encoder=DateEncoder)
+    details = models.JSONField(verbose_name="对话详情", default=dict, encoder=SystemEncoder)
     improve_paragraph_id_list = ArrayField(verbose_name="改进标注列表",
                                            base_field=models.UUIDField(max_length=128, blank=True)
                                            , default=list)
```
shaohuzhang1, line 155: The given code contains several improvements and optimizations:

1. **Import Statements**: Removed unnecessary import statements, especially `json` which is imported implicitly if needed.
   
   ```python
   # Remove:
   # import datetime
   # import decimal
   ```

2. **UUID Encoding**: The current implementation of the `DateEncoder` encodes `uuid.UUID` objects to strings rather than keeping them as-is. This can cause some problems since UUIDs usually have more meaning and should ideally be kept unchanged.

3. **JSONDecoder Usage**: In Django's ORM JSON fields (`JSONField`), there’s no need to manually define an encoder like in Python dictionaries due to their support for automatic conversion between native types and JSON serialization/deserialization using a built-in mechanism.

Therefore, you can simplify the encoding handling by removing the custom `DateEncoder` and focusing on maintaining consistent type conversions where necessary:

```python
class ChatRecord(AppModelMixin):
    """
    对话日志 详情
    """

    message_tokens = models.IntegerField(verbose_name="请求token数量", default=0)
    answer_tokens = models.IntegerField(verbose_name="响应token数量", default=0)
    const = models.IntegerField(verbose_name="总费用", default=0)
    details = models.JSONField(
        verbose_name="对话详情",
        default=dict,
    )  # No additional encoder needed here; auto-conversion works fine.
    improve_paragraph_id_list = ArrayField(
        verbose_name="改进标注列表",
        base_field=models.UUIDField(max_length=128, blank=True),
        default=list,
    )
```
With these changes, the code remains clean and efficient while leveraging Django's built-in capabilities better.

apps/setting/migrations/0010_log.py
```diff
@@ -16,13 +18,15 @@ class Migration(migrations.Migration):
...
                 ('operate', models.CharField(max_length=128, verbose_name='操作')),
                 ('user', models.JSONField(default=dict, verbose_name='用户信息')),
                 ('status', models.IntegerField(max_length=20, verbose_name='状态')),
                 ('ip_address', models.CharField(max_length=128, verbose_name='ip地址')),
-                ('details', models.JSONField(default=dict, verbose_name='详情')),
+                ('details',
+                 models.JSONField(default=dict, encoder=common.encoder.encoder.SystemEncoder, verbose_name='详情')),
             ],
             options={
                 'db_table': 'log',
```
shaohuzhang1, line 32: The code looks generally correct but has a few minor improvements and corrections you can consider:

1. **Imports**: Ensure that all necessary imports are at the top of the file to improve readability.

2. **UUID Field**: The `id` field already has an `encoder` parameter specified as a keyword argument when using it. This is redundant here since you've used another set of parentheses immediately after the type hinting, which is causing confusion. You can remove the explicit `encoder=common.encoder.encoder.SystemEncoder` part.

3. **Consistency**: Use consistent whitespace around operators and inside brackets for better readability.

Here is the corrected version of the code:

```python
import uuid
import common.encoder.encoder

    dependencies = [
        ('setting', '0009_set_default_model_params_form'),
    ]

    operations = [
        migrations.CreateModel(
            name='Log',
            fields=[
                ('create_time', models.DateTimeField(auto_now_add=True, verbose_name='创建时间')),
                ('update_time', models.DateTimeField(auto_now=True, verbose_name='修改时间')),
                ('id', models.UUIDField(default=uuid.uuid1, editable=False, primary_key=True, serialize=False)),
                ('menu', models.CharField(max_length=128, verbose_name='操作菜单')),
                ('operate', models.CharField(max_length=128, verbose_name='操作')),
                ('user', models.JSONField(default=dict, verbose_name='用户信息')),
                ('status', models.IntegerField(max_length=20, verbose_name='状态')),
                ('ip_address', models.CharField(max_length=128, verbose_name='ip地址')),
                ('details', models.JSONField(default=dict, encoder=common.encoder.encoder.SystemEncoder, verbose_name='详情')),
            ],
            options={
                'db_table': 'log'
            },
        ),
    ]
```

By making these changes, the code will be more readable and correctly formatted while maintaining its functionality.

PR #2613, opened 2025-03-19
fix: Logging error for logged in user

shaohuzhang1:
fix: Logging error for logged in user

### DIFF
--- a/apps/common/log/log.py
+++ b/apps/common/log/log.py
@@ -6,6 +6,7 @@
     @date：2025/3/14 16:09
     @desc:
 """
+from gettext import gettext
 
 from setting.models.log_management import Log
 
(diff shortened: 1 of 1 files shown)


apps/common/log/log.py
```diff
@@ -31,6 +32,10 @@ def _get_user(request):
     @return:
     """
     user = request.user
+    if user is None:
+        return {
+            "user_name": gettext('unknown')
+        }
     return {
         "id": str(user.id),
         "email": user.email,
```
shaohuzhang1, line 41: The provided code does not contain any syntax errors or obvious issues. However, here's an additional optimization suggestion:

```python
def _get_user(request):
    user = request.user
    if user:
        return {
            "id": str(user.id),
            "email": user.email,
        }

    # Return a common placeholder for unknown users
    return {
        "user_name": gettext("unknown"),
    }
```

This slightly optimizes the function by reducing redundancy and making it more readable, although performance has minimal impact on this simple function.

PR #2623, opened 2025-03-19
feat: add operate log

shaohuzhang1:
feat: add operate log

### DIFF
--- a/ui/src/views/operate-log/index.vue
+++ b/ui/src/views/operate-log/index.vue
@@ -211,7 +211,7 @@ function changeFilterHandle(val: string) {
 function changeDayHandle(val: number | string) {
   if (val !== 'other') {
     daterange.value.start_time = beforeDay(val)
-    daterange.value.end_time = nowDate
+    daterange.value.end_time = ''
     getList()
   }
 }


ui/src/views/operate-log/index.vue
```diff
@@ -211,7 +211,7 @@ function changeFilterHandle(val: string) {
 function changeDayHandle(val: number | string) {
   if (val !== 'other') {
     daterange.value.start_time = beforeDay(val)
-    daterange.value.end_time = nowDate
+    daterange.value.end_time = ''
     getList()
   }
 }
```
shaohuzhang1, line 217: The provided code is mostly correct, but there's one small issue that could be improved:

```javascript
function changeDayHandle(val: number | string) {
  if (val !== 'other') {
    daterange.value.start_time = beforeDay(val);
    // Avoid an error when using `beforeDay` on 'other'
    daterange.value.end_time = typeof val === 'number' ? '' : nowDate;
    getList();
  }
}
```

**Explanation of changes**:
1. Added a type check (`typeof`) to ensure that the value used in `nowDate` is a number before performing arithmetic with it, which prevents any errors. If `val` is not a number ('unknown'), the end time will remain unchanged instead of being set to undefined (`undefined`), which can lead to unintended behavior.

If you have more specific requirements or optimizations needed for this function, please clarify!

PR #2626, opened 2025-03-19
feat: add operate log

shaohuzhang1:
feat: add operate log

### DIFF
--- a/ui/src/api/operate-log.ts
+++ b/ui/src/api/operate-log.ts
@@ -22,6 +22,11 @@ const getOperateLog: (
   return get(`${prefix}/${page.current_page}/${page.page_size}`, param, loading)
 }
 
+const getMenuList: () => Promise<Result<any>> = () => {
+  return get(`${prefix}/menu_operate_option/`, undefined, undefined)
+}
+
 export default {
-  getOperateLog
+  getOperateLog,
+  getMenuList
 }
--- a/ui/src/views/operate-log/index.vue
+++ b/ui/src/views/operate-log/index.vue
@@ -76,7 +76,49 @@
         @changePage="getList"
         v-loading="loading"
       >
-        <el-table-column prop="menu" :label="$t('views.operateLog.table.menu.label')" width="160" />
+        <el-table-column prop="menu" :label="$t('views.operateLog.table.menu.label')" width="160">
+          <template #header>
+            <div>
+              <span>{{ $t('views.operateLog.table.menu.label') }}</span>
+              <el-popover :width="200" trigger="click" :visible="popoverVisible">
+                <template #reference>
+                  <el-button
+                    style="margin-top: -2px"
+                    :type="operateTypeArr && operateTypeArr.length > 0 ? 'primary' : ''"
+                    link
+                    @click="popoverVisible = !popoverVisible"
+                  >
+                    <el-icon>
+                      <Filter />
+                    </el-icon>
+                  </el-button>
+                </template>
+                <div class="filter">
+                  <div class="form-item mb-16">
+                    <div @click.stop>
+                      <el-scrollbar height="300" style="margin: 0 0 0 10px;">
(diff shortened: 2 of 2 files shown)


ui/src/views/operate-log/index.vue
```diff
@@ -241,15 +286,29 @@ function getList() {
...
+  }) 
+}
+
 onMounted(() => {
+  getMenuList()
   changeDayHandle(history_day.value)
-  getList()
 })
 </script>
 <style lang="scss" scoped>
```
shaohuzhang1, line 314: The code looks good overall, but there are a few areas that could be optimized or improved:

1. **Code Duplication**: The filter logic is repeated for both `menu` and `operate`. Consider refactoring this into a separate function to make it cleaner.

2. **Empty Check in Filter Change**: Ensure that when `val` is `'clear'`, the checkbox array is properly cleared before resetting the list of options.

3. **API Call Optimization**: If fetching menu data from `/api/operate-log/getMenuList` is expensive or frequently needed, consider caching the results locally.

4. **Loading State Handling**: It might be beneficial to introduce more granular loading states based on different API requests.

Here's an updated version of the code with some optimizations:

```vue
<template>
  <!-- Table component remains mostly unchanged -->
</template>

<script setup lang="ts">
import { ref, onMounted, reactive } from 'vue'
import getOperateLog from '@/api/operate-log'
import DetailDialog from './component/DetailDialog.vue'
import { t } from '@/locales'
import { beforeDay, datetimeFormat, nowDate } from '@/utils/time'

const popoverVisible = ref(false)
const operateTypeArr = ref<any[]>([])
const DetailDialogRef = ref()
const loading = ref(false)

// Other variables remain mostly unchanged

onMounted(() => {
  getMenuList()
  changeDayHandle(history_day.value)
  getList()
})

async function fetchData(params: any): Promise<void> {
  try {
    const response = await getOperateLog.getOperateLog(paginationConfig, params, loading);
    tableData.value = response.data.records;
    paginationConfig.total = response.data.total;
  } catch (error) {
    // Handle error appropriately
  }
}

function filterChange(val: string) {
  if (val === 'clear') {
    operateTypeArr.value = [];
  } else {
    const filteredParams: Record<string, any> = {};
    
    if (filter_type.value === 'status') {
      filteredParams['status'] = filter_status.value;
    }

    if (operateTypeArr.value.length > 0 && filter_type.value === 'menu') {
      filteredParams['menu'] = JSON.stringify(operateTypeArr.value);
    }

    fetchData(filteredParams);
  }
  popoverVisible.value = false;
}

// ...
```

**Explanation of Changes:**

1. **Moved Fetch Logic to Separate Function**: Created `fetchData` to encapsulate the API call logic, making it reusable and clean.

2. **Updated Filter Functionality**: The filter function no longer needs checks for each property separately; they all use the same pattern now.

This should improve maintainability and reduce redundancy while keeping the app responsive and efficient.

ui/src/api/operate-log.ts
```diff
@@ -22,6 +22,11 @@ const getOperateLog: (
...
 
+const getMenuList: () => Promise<Result<any>> = () => {
+  return get(`${prefix}/menu_operate_option/`, undefined, undefined)
+}
+
 export default {
-  getOperateLog
+  getOperateLog,
+  getMenuList
 }
```
shaohuzhang1, line 32: The provided code looks clean and well-structured. However, there is one minor suggestion to make it more concise:

```diff
@@ -22,7 +22,6 @@ const getOperateLog: (
 export default {
-  getOperateLog,
+  ...getOperateLog,
 } 
```

This will import `...getOperateLog` from wherever it is defined, effectively including all properties of the `getOperateLog` function object into the exported module. This can be useful if you want to avoid listing each property individually.

For other checks, such as type checking or security considerations, I would need additional context about how these functions are used and what kind of operations they perform with their parameters. If the endpoint requires specific types or handles sensitive data properly, ensure that this is accounted for in your environment. Additionally, consider adding input validation before making API calls to prevent common errors like missing or incorrect parameters.

PR #2629, opened 2025-03-20
fix: type-check

shaohuzhang1:
fix: type-check

### DIFF
--- a/ui/src/request/index.ts
+++ b/ui/src/request/index.ts
@@ -1,4 +1,4 @@
-import axios, { type AxiosRequestConfig } from 'axios'
+import axios, { type InternalAxiosRequestConfig, AxiosHeaders } from 'axios'
 import { MsgError } from '@/utils/message'
 import type { NProgress } from 'nprogress'
 import type { Ref } from 'vue'
@@ -19,9 +19,9 @@ const instance = axios.create(axiosConfig)
 
 /* 设置请求拦截器 */
 instance.interceptors.request.use(
-  (config: AxiosRequestConfig) => {
+  (config: InternalAxiosRequestConfig) => {
     if (config.headers === undefined) {
-      config.headers = {}
+      config.headers = new AxiosHeaders()
     }
     const { user } = useStore()
     const token = user.getToken()


ui/src/request/index.ts
```diff
@@ -19,9 +19,9 @@ const instance = axios.create(axiosConfig)
...
 /* 设置请求拦截器 */
 instance.interceptors.request.use(
-  (config: AxiosRequestConfig) => {
+  (config: InternalAxiosRequestConfig) => {
     if (config.headers === undefined) {
-      config.headers = {}
+      config.headers = new AxiosHeaders()
     }
     const { user } = useStore()
     const token = user.getToken()
```
shaohuzhang1, line 27: There seems to be an issue with the way headers are being accessed and initialized. Here's a few points you might consider:

1. **Header Initialization**: Ensure that `config.headers` is properly set to an empty object or newly created instance of `AxiosHeaders`. The current access `if (config.headers === undefined)` does not guarantee that `headers` will be an instance of `AxiosHeaders`.

```typescript
instance.interceptors.request.use(
  (config: InternalAxiosRequestConfig) => {
    config.headers = new AxiosHeaders();
```

---

2. **Ref Type Check**: If you're using Vue Composition API refs, there should be no need to create a new `AxiosHeaders` object unless `headers` is initially empty or not set.

If `user.token` can be null or undefined and you want to handle such cases gracefully, you could check it before setting it as a header.

```typescript
const { user } = useStore();
let token = user ? user.getToken() : '';

if (token && !Object.prototype.hasOwnProperty.call(config?.headers ?? {}, 'Authorization')) {
  config.headers.Authorization = `Bearer ${token}`;
}
```

Here’s a cleaned up version of your code based on these considerations:

```typescript
// Import statements remain unchanged

// Create axios instance with configurations
const instance = axios.create(axiosConfig);

/* Set request interceptors */
instance.interceptors.request.use(
  (config: InternalAxiosRequestConfig) => {
    // Initialize headers if they aren't already present
    if (!Reflect.hasOwnProp(config.headers || {}, 'Content-Type') &&
        !Reflect.hasOwnProp(config.headers || {}, 'Accept') &&
        !Reflect.hasOwnProp(config.headers || {}, 'X-Requested-With')) {
      config.headers = new AxiosHeaders({
        Content-Type: 'application/json',
        Accept: 'application/json', 
        'X-Requested-With': 'XMLHttpRequest'
      });
    }

    // Get authentication token from store
    const { user } = useStore();

    // Add authorization token only if available
    if (user.token) {
      config.headers.Authorization = `Bearer ${user.token}`;      
    }

    return config;
  },
  
  error => Promise.reject(error),
);
```

This ensures that headers are always initialized correctly regardless of their current state and makes sure authorization tokens are appropriately added when necessary based on user data availability.

Issue #2604, opened 2025-03-18
[Bug]


zfno111:
### MaxKB Version

v1.10.2-lts (build at 2025-03-06T19:07, commit: babe13a8)

### Problem Description

We uploaded a batch of files through our script, and the status of these files changed as follows:
Failed → Indexing → Success.
I didn’t perform any operations during this process.
The "Failed" status seems more like a waiting state.

Here’s the translation in English:
We uploaded a batch of files through our script, and the status of these files changed as follows:
Failed → Indexing → Success.
I didn’t perform any operations during this process.
The "Failed" status seems more like a waiting state.

### Steps to Reproduce

we have 500files, upload thoses file by api

### Related log output

```shell

```

Shenguobin0102:
你好，感谢反馈，请问你调用的是哪个上传文件的接口呢？

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Hello, thank you for the feedback, which interface are you calling for uploading files?

PR #2633, opened 2025-03-20
fix: Application import deserialization vulnerability #2604

shaohuzhang1:
fix: Application import deserialization vulnerability #2604

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -39,7 +39,7 @@ from common.exception.app_exception import AppApiException, NotFound404, AppUnau
 from common.field.common import UploadedImageField, UploadedFileField
 from common.models.db_model_manage import DBModelManage
 from common.response import result
-from common.util.common import valid_license, password_encrypt
+from common.util.common import valid_license, password_encrypt, restricted_loads
 from common.util.field_message import ErrMessage
 from common.util.file_util import get_file_content
 from dataset.models import DataSet, Document, Image
@@ -60,6 +60,7 @@ chat_cache = cache.caches['chat_cache']
 
 
 class MKInstance:
+
     def __init__(self, application: dict, function_lib_list: List[dict], version: str):
         self.application = application
         self.function_lib_list = function_lib_list
@@ -727,7 +728,7 @@ class ApplicationSerializer(serializers.Serializer):
             user_id = self.data.get('user_id')
             mk_instance_bytes = self.data.get('file').read()
             try:
-                mk_instance = pickle.loads(mk_instance_bytes)
+                mk_instance = restricted_loads(mk_instance_bytes)
             except Exception as e:
                 raise AppApiException(1001, _("Unsupported file format"))
             application = mk_instance.application
@@ -813,7 +814,7 @@ class ApplicationSerializer(serializers.Serializer):
             return FunctionLibSerializer.Query(
                 data={'user_id': application.user_id, 'is_active': True,
                       'function_type': FunctionType.PUBLIC}
-                ).list(with_valid=True)
+            ).list(with_valid=True)
 
         def get_function_lib(self, function_lib_id, with_valid=True):
             if with_valid:
--- a/apps/common/util/common.py
+++ b/apps/common/util/common.py
@@ -10,6 +10,7 @@ import hashlib
 import importlib
 import io
 import mimetypes
+import pickle
 import re
 import shutil
 from functools import reduce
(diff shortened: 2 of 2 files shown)


apps/common/util/common.py
```diff
@@ -23,6 +24,30 @@
...
+                                     (module, name))
+
+
+def restricted_loads(s):
+    """Helper function analogous to pickle.loads()."""
+    return RestrictedUnpickler(io.BytesIO(s)).load()
+
 
 def encryption(message: str):
     """
```
shaohuzhang1, line 53: Your code looks mostly correct, with one minor suggestion regarding readability. You have defined `ALLOWED_CLASSES` using a set instead of a tuple, which makes sense since sets cannot contain duplicate elements.

Additionally, I would suggest adding comments to explain the purpose of each class and method, especially for clarity and documentation purposes.

Here's an updated version of your code:

```python
# Import necessary libraries
import importlib
import io
import mimetypes
import pickle
import re
import shutil
from functools import reduce

# Import custom models
from .exception.app_exception import AppApiException
from .models.db_model_manage import DBModelManage

# Define a dictionary to keep track of safe built-in classes
safe_builtins = {

}

# Dictionary of allowed classes that can be unpickled
ALLOWED_CLASSES = {
    ("builtins", "dict"),      # Allow Python dictionaries (e.g., {})
    ('uuid', 'UUID'),         # Allow UUIDs from the uuid module
    ("application.serializers.application_serializers", "MKInstance")  # Allow MKInstance objects from application.serializers
}


class RestrictedUnpickler(pickle.Unpickler):

    def find_class(self, module, name):
        """
        Method overridden to restrict the classes that can be unpickled.
        
        Parameters:
            module: The module containing the class.
            name: The name of the class within the module.
            
        Returns:
            If the class exists in ALLOWED_CLASSES, it returns the original find_class() result.
            Otherwise, raises a UnpicklingError indicating that the class is not allowed.
        """
        if (module, name) in ALLOWED_CLASSES:
            return super().find_class(module, name)
        raise pickle.UnpicklingError("global '%s.%s' is forbidden" % (module, name))


def restricted_loads(s):
    """
    Helper function to load pickled data securely.
    
    Parameters:
        s: A string representing thepickle'd binary data.
        
    Returns:
        An object loaded from the pickled data.
    """
    return RestrictedUnpickler(io.BytesIO(s)).load()


# Function to encrypt a message (this part could use further implementation based on encryption library specifics)
def encryption(message: str):
    """
    Encrypts a given message.
    
    Parameters:
        message: The message to be encrypted.
        
    Returns:
        The encrypted message.
    """
    # Placeholder logic for encryption
    return f"[ENCRYPTED] {message}"
```

### Key Observations:
1. **Commenting**: Added comments to describe the purpose of each class and method.
2. **Set vs Tuple**: Changed `ALLOWED_CLASSES` to a set for better performance and consistency.
3. **Security Focus**: Ensured that only specified classes can be unpickled, enhancing security against unauthorized access when dealing with serialized data.

apps/application/serializers/application_serializers.py
```diff
@@ -813,7 +814,7 @@ def list_function_lib(self, with_valid=True):
             return FunctionLibSerializer.Query(
                 data={'user_id': application.user_id, 'is_active': True,
                       'function_type': FunctionType.PUBLIC}
-                ).list(with_valid=True)
+            ).list(with_valid=True)
 
         def get_function_lib(self, function_lib_id, with_valid=True):
             if with_valid:
```
shaohuzhang1, line 820: The provided code has several minor issues:

1. **Redundant Import**: The line `from common.util.common import restricted_loads` is unnecessary because you've already imported it earlier with `import util`.

2. **Unused Imports**: The imports of `DBModelManage`, `result`, and `ErrMessage` can be removed if not used.

3. **Typographical Errors**:
   - `"AppApiException"` should be defined before use.
   - Incorrectly formatted dictionary keys like `'is_active'` instead of `'isActive'`.
   - Potential typo in `'function_type'` which might incorrectly identify function types.

4. **Comments Formatting**: There are inconsistent indentation levels for comments, making them less readable.

5. **Variable Names**: Use camelCase for variable names to follow Python's naming conventions (e.g., `user_id` -> `userId`).

Here's revised version considering these points:

```python
from common.field.common import UploadedImageField, UploadedFileField
from common.models.db_model_manage import Application, Dataset, Document, Image
from common.response import success_result
from common.util.common import valid_license, password_encrypt, restricted_loads

class MKInstance:
    def __init__(self, application: dict, function_lib_list: List[dict], version: str):
        self.application = application
        self.function_lib_list = function_lib_list

class AppApiException(Exception):
    pass

def list_function_lib_with_valid(user_id=None, is_active=None, function_type=None):
    """List all public function libraries associated with a user"""
    params = QueryDict({'user_id': user_id, 'is_active': is_active, 'function_type': function_type})
    query_set = Application.objects.filter(is_active=is_active).values('id', 'name')
    return success_result([{'applicationId': row['id'], 'name': row['name']} for row in query_set])

def list_function_lib_without_valid(**kwargs):
    return [FunctionLibSerializer.apply(instance=library) for library in Library.objects.all()]

def get_function_lib(function_lib_id, with_valid=True):
    if with_valid:
        # Check validation here
       ...
```

Ensure that all errors mentioned above are addressed when implementing the actual logic.

PR #2669, opened 2025-03-24
fix: enhance validation for required form fields

shaohuzhang1:
fix: enhance validation for required form fields  --bug=1052705 --user=刘瑞斌 【工作流】添加”开关”按钮参数， 默认值为关闭，保存之后页面还会提示“请填写所填字段” https://www.tapd.cn/57709429/s/1675323

### DIFF
--- a/ui/src/components/ai-chat/component/user-form/index.vue
+++ b/ui/src/components/ai-chat/component/user-form/index.vue
@@ -303,7 +303,9 @@ const checkInputParam = () => {
   for (let i = 0; i < inputFieldList.value.length; i++) {
     if (
       inputFieldList.value[i].required &&
-      !form_data_context.value[inputFieldList.value[i].field]
+      (form_data_context.value[inputFieldList.value[i].field] === null ||
+        form_data_context.value[inputFieldList.value[i].field] === undefined ||
+        form_data_context.value[inputFieldList.value[i].field] === '')
     ) {
       MsgWarning(t('chat.tip.requiredMessage'))
       return false


ui/src/components/ai-chat/component/user-form/index.vue
```diff
@@ -303,7 +303,9 @@ const checkInputParam = () => {
   for (let i = 0; i < inputFieldList.value.length; i++) {
     if (
       inputFieldList.value[i].required &&
-      !form_data_context.value[inputFieldList.value[i].field]
+      (form_data_context.value[inputFieldList.value[i].field] === null ||
+        form_data_context.value[inputFieldList.value[i].field] === undefined ||
+        form_data_context.value[inputFieldList.value[i].field] === '')
     ) {
       MsgWarning(t('chat.tip.requiredMessage'))
       return false
```
shaohuzhang1, line 311: The provided code checks if required fields in `inputFieldList` are present in `form_data_context`. It appears to have an issue where it incorrectly considers `null` and `undefined` values along with empty strings as valid inputs for fields that should be non-empty. This might lead to validation errors even when the field is intentionally left blank.

Here's a suggested improvement to handle this:

```javascript
if (
  inputFieldList.value[i].required &&
  !(formDataContext.value[inputFieldList.value[i].field]) // Ensure only undefined/null/empty string is considered invalid
) {
```

This change ensures that only truly absent fields (i.e., those not defined or set to `null` or `'') result in a validation error.

Additionally, ensure that other sections of your code also consider these cases correctly.

PR #2674, opened 2025-03-25
fix: lark document icon error

shaohuzhang1:
fix: lark document icon error  --bug=1053854 --user=王孝刚 【飞书】-创建完飞书知识库后，面板中飞书的icon未变 https://www.tapd.cn/57709429/s/1675539

### DIFF
--- a/ui/src/views/dataset/index.vue
+++ b/ui/src/views/dataset/index.vue
@@ -57,6 +57,15 @@
                   >
                     <img src="@/assets/icon_web.svg" style="width: 58%" alt="" />
                   </AppAvatar>
+                  <AppAvatar
+                    v-else-if="item.type === '2'"
+                    class="mr-8 avatar-purple"
+                    shape="square"
+                    :size="32"
+                    style="background: none"
+                  >
+                    <img src="@/assets/logo_lark.svg" style="width: 100%" alt="" />
+                  </AppAvatar>
                   <AppAvatar v-else class="mr-8 avatar-blue" shape="square" :size="32">
                     <img src="@/assets/icon_document.svg" style="width: 58%" alt="" />
                   </AppAvatar>


ui/src/views/dataset/index.vue
```diff
@@ -57,6 +57,15 @@
...
+                    class="mr-8 avatar-purple"
+                    shape="square"
+                    :size="32"
+                    style="background: none"
+                  >
+                    <img src="@/assets/logo_lark.svg" style="width: 100%" alt="" />
+                  </AppAvatar>
                   <AppAvatar v-else class="mr-8 avatar-blue" shape="square" :size="32">
                     <img src="@/assets/icon_document.svg" style="width: 58%" alt="" />
                   </AppAvatar>
```
shaohuzhang1, line 71: The provided code snippet appears to be configuring `AppAvatar` components within an application. Here are some observations and recommendations:

### Observations:
1. **Multiple Avatar Types**: The code supports three types of avatars based on `item.type`. This is good separation but should be documented clearly to avoid confusion among developers.
2. **Conditional Logic**:
   - There is only one condition (`type === '1'`) handling a specific avatar type without providing additional conditions for other types (e.g., `'2'`). Consider expanding the conditional logic to handle all defined types.
   - It's not clear what happens if `item.type` is neither `'1'`, `'2'`, nor another valid case since no default or fallback logic is present.

### Potential Issues and Optimizations:
1. **Default Case Handling**: Without explicit handling for unknown or missing `type` values, it might lead to unexpected behavior at runtime.
2. **Readability and Consistency**: Ensure that different scenarios are consistently handled. For example, if certain styles or images apply universally regardless of type, they can be centralized or moved away from the conditional logic.
3. **Code Duplication**: If multiple `v-else-if` blocks follow similar patterns, consider refactoring them into a single block with template literals or object destructuring for better readability.
4. **Styling Flexibility**: The styling applied in each `AppAvatar` instance is consistent across both image sizes and shapes. While this provides uniformity, flexibility could be achieved by making these properties dynamic via props or computed properties to allow customization based on specific needs.

Here’s how you might refactor or update the conditional logic slightly more explicitly:

```html
<AppAvatar
  v-for="(item, index) in items"
  :key="index"
  :class="[
    'mr-8',
    item.type !== '1'
      ? 'avatar-purple'
      : item.type === '2'
        ? 'avatar-purple square'
        : 'avatar-blue square'
  ]"
  :shape="'square'"
  :size="32"
  style="background: none"
>
  <template slot-scope="{ imgSize }">
    <!-- Dynamically choose between two icons based on item.type -->
    <img :src="getAssetUrl(item)" :style="{ width: `${imgSize}px` }" alt="" />
  </template>
</AppAvatar>

// Assuming getAssetUrl returns a path string based on item.type
function getAssetUrl(item) {
  return (
    item.type === '1'
      ? require('@/assets/icon_web.svg')
      : item.type === '2'
        ? require('@/assets/logo_lark.svg')
        : require('@/assets/icon_document.svg')
  );
}
```

This approach centralizes the logic for selecting assets and ensures that the component remains flexible and maintainable, especially when adding further supported avatar types.

Issue #2537, opened 2025-03-11
[Bug] Extra spaces in the default prompt when the language is English.


wan92hen:
### MaxKB Version

v1.10.2

### Problem Description

![Image](https://github.com/user-attachments/assets/3f3fbbe9-944a-4fca-9016-4fa7be3e9009)

### Steps to Reproduce

1. Switching the language to English.
2. Create an new application.
3. Checking the default prompt.

### Related log output

```shell

```

zyyfit:
感谢反馈，后续版本修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version fixes

baixin513:
V1.10.3-lts版本已发布，请更新试用！

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
V1.10.3-lts version has been released, please update and try it out!

PR #2690, opened 2025-03-26
perf: Prompt word alignment issue(#2537)

wangdan-fit2cloud:

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -67,6 +67,7 @@
             <slot name="operateBefore">
               <span></span>
             </slot>
+
             <el-button
               v-if="isUserInput"
               class="user-input-button mb-8"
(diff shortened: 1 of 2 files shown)


ui/src/components/ai-chat/index.vue
```diff
@@ -67,6 +67,7 @@
             <slot name="operateBefore">
               <span></span>
             </slot>
+
             <el-button
               v-if="isUserInput"
               class="user-input-button mb-8"
```
shaohuzhang1, line 73: The provided code looks mostly clean. However, here's a suggestion for improvement:

```markdown
- **Suggestion:** Consider adding `v-slot` with dynamic arguments instead of using `<slot>` directly if you expect multiple slots to be used in different places within the same component.

For example, if you have multiple custom slots that need distinct behaviors based on their names (e.g., `"header"`, `"content"`), you can use `v-slot:[name]`.

Example usage:
```html
<template>
  <div>
    <!-- Dynamic slot -->
    <component :is="'custom-slot-' + name"></component>

    <!-- Regular slots -->
    <slot></slot> <!-- Fallback/default slot -->

    <!-- Specific named slots -->
    <slot name="title">Default title</slot>
  </div>
</template>

<script setup>
import CustomSlotTitle from './CustomSlotTitle.vue'
import CustomSlotContent from './CustomSlotContent.vue'

defineProps({
  // props...
})

const componentName = () => {
  const { name } = getCurrentInstance().type;
  return name.startsWith('CustomSlot') ? name.replace(/^CustomSlot/i, '') : '';
};
</script>
```

This approach provides more flexibility and clarity about how your slots should behave across different templates. If you have a specific implementation in mind when dealing with custom slots, let me know!

PR #2691, opened 2025-03-26
fix: Audit log login did not record user

shaohuzhang1:
fix: Audit log login did not record user

### DIFF
--- a/apps/common/log/log.py
+++ b/apps/common/log/log.py
@@ -34,7 +34,7 @@ def _get_user(request):
     user = request.user
     if user is None:
         return {
-            "user_name": gettext('unknown')
+
         }
     return {
         "id": str(user.id),
--- a/apps/users/views/user.py
+++ b/apps/users/views/user.py
@@ -179,7 +179,7 @@ class Login(APIView):
                          responses=LoginSerializer().get_response_body_api(),
                          security=[],
                          tags=[_("User")])
-    @log(menu='User', operate='Log in', get_user=lambda r: {'user_name': r.data.get('username', None)},
+    @log(menu='User', operate='Log in', get_user=lambda r: {'username': r.data.get('username', None)},
          get_details=_get_details,
          get_operation_object=lambda r, k: {'name': r.data.get('username')})
     def post(self, request: Request):


apps/common/log/log.py
```diff
@@ -34,7 +34,7 @@ def _get_user(request):
     user = request.user
     if user is None:
         return {
-            "user_name": gettext('unknown')
+
         }
     return {
         "id": str(user.id),
```
shaohuzhang1, line 40: The provided code does not appear to contain any significant errors or issues, but there are a few minor improvements that can be made:

### Improvements and Suggestions

1. **Consistency with `gettext`**: The current implementation uses `'unknown'`, which should be consistent across the application. If you intended to use it for internationalization (i18n), make sure everything else in your project respects this practice.

2. **Default Parameters**: It's good practice to include default parameters when defining functions, especially if some arguments might be optional. Although not strictly necessary here, adding them can improve readability and usage patterns.

3. **Simplified Dictionary Assignment**: You can simplify the dictionary assignment inside the function using f-strings:

   ```python
   def _get_user(request):
       user = request.user
       if user is None:
           return {"user_name": "unknown"}  # Using string directly instead of gettext function
    
       return {"id": str(user.id)}
   ```

This streamlined version maintains the same functionality while being slightly more concise and adhering to common Python practices.

apps/users/views/user.py
```diff
@@ -179,7 +179,7 @@ class Login(APIView):
                          responses=LoginSerializer().get_response_body_api(),
                          security=[],
                          tags=[_("User")])
-    @log(menu='User', operate='Log in', get_user=lambda r: {'user_name': r.data.get('username', None)},
+    @log(menu='User', operate='Log in', get_user=lambda r: {'username': r.data.get('username', None)},
          get_details=_get_details,
          get_operation_object=lambda r, k: {'name': r.data.get('username')})
     def post(self, request: Request):
```
shaohuzhang1, line 185: The code is largely correct, but there are a few minor improvements and clarifications that can be made:

1. It's good practice to use the `@api_view` decorator instead of inheriting from `APIView`. This allows you to specify which HTTP methods are accepted by the view more directly.
2. The `_get_details` function is called at the end of each operation, regardless of whether an error occurs or not. Consider calling it only when no errors occurred to avoid unnecessary database access if an exception is raised.

Here's a revised version with these points addressed:

```python
from rest_framework.decorators import api_view

@api_view(['POST'])
def login(request: Request):
    serializer = LoginSerializer(data=request.data)
    
    if serializer.is_valid(raise_exception=True): 
        return login_successful(serializer.user)

@log(
    menu='User',
    operate='Log in',
    get_user=lambda r: {'user_name': r.data.get('username', None)},  
    get_details=_get_details,
    get_operation_object=lambda r, k: {'name': r.data.get('username')}
)


# Define other functions needed here...
``` 

Make sure all necessary imports (like `LoginSerializer`, `login_successful`) are correctly imported for this solution to work.

PR #2693, opened 2025-03-26
fix: Segmented filtering of paragraphs with empty parent title content

shaohuzhang1:
fix: Segmented filtering of paragraphs with empty parent title content

### DIFF
--- a/apps/common/util/split_model.py
+++ b/apps/common/util/split_model.py
@@ -344,9 +344,9 @@ class SplitModel:
                 'content' in item and len(item.get('content').strip()) > 0]
 
     def post_reset_paragraph(self, paragraph: Dict, title_list: List[str]):
-        result = self.filter_title_special_characters(paragraph)
+        result = self.content_is_null(paragraph, title_list)
+        result = self.filter_title_special_characters(result)
         result = self.sub_title(result)
-        result = self.content_is_null(result, title_list)
         return result
 
     @staticmethod


apps/common/util/split_model.py
```diff
@@ -344,9 +344,9 @@ def parse(self, text: str):
...
 
     def post_reset_paragraph(self, paragraph: Dict, title_list: List[str]):
-        result = self.filter_title_special_characters(paragraph)
+        result = self.content_is_null(paragraph, title_list)
+        result = self.filter_title_special_characters(result)
         result = self.sub_title(result)
-        result = self.content_is_null(result, title_list)
         return result
 
     @staticmethod
```
shaohuzhang1, line 352: The provided code snippet is incomplete and lacks the necessary content to properly analyze it. It seems like there's an ongoing process of refactoring or commenting out certain sections of the code, where `self.filter_title_special_characters`, `self.content_is_null`, and other methods or attributes are referenced but not defined.

To assist effectively:

1. **Complete Definition**: Ensure all required functions (e.g., `filter_title_special_characters`, `sub_title`) and their dependencies are added to your class.
  
2. **Ensure Correct Method Signatures**: Verify that all method signatures (`post_reset_paragraph`) match those expected by your class definition.

3. **Add Context**: Provide more context about how these methods interact with each other since you've mentioned they're used together in `post_reset_paragraph`.

Here’s what I mean by some general improvements:

```python
def parse(self, text: str) -> List[Dict]:
    # Your implementation here

@staticmethod
def filter_title_special_characters(text: Dict) -> Dict:
    # Remove special characters from titles
    return {
        'title': ''.join(char for char in text['title'] if char.isalnum() or char == '-'),
        'content': ''.join(
            char.lower() if char.isalpha() else (
                '' if len(str(int(char))) != 1	else char.upper()
            ) for char in text['content']
        )
    }

@staticmethod
def sub_title(text: Dict) -> Dict:
    # Implement subtitle removal/logic as needed
    pass
```

If you have specific requirements, additional logic can be implemented accordingly.

PR #2699, opened 2025-03-26
fix: Update lark icon and dialog style

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/function-lib/component/EditAvatarDialog.vue
+++ b/ui/src/views/function-lib/component/EditAvatarDialog.vue
@@ -1,21 +1,21 @@
 <template>
   <el-dialog
-      :title="`Logo ${$t('common.setting')}`"
-      v-model="dialogVisible"
-      :close-on-click-modal="false"
-      :close-on-press-escape="false"
+    :title="`Logo ${$t('common.setting')}`"
+    v-model="dialogVisible"
+    :close-on-click-modal="false"
+    :close-on-press-escape="false"
   >
-    <el-radio-group v-model="radioType" class="radio-block mb-16">
+    <el-radio-group v-model="radioType" class="radio-block-avatar mb-16">
       <div>
         <el-radio value="default">
           <p>{{ $t('views.applicationOverview.appInfo.EditAvatarDialog.default') }}</p>
           <AppAvatar
-              v-if="detail?.name"
-              :name="detail?.name"
-              pinyinColor
-              class="mt-8 mb-8"
-              shape="square"
-              :size="32"
+            v-if="detail?.name"
+            :name="detail?.name"
+            pinyinColor
+            class="mt-8 mb-8"
+            shape="square"
+            :size="32"
           />
         </el-radio>
       </div>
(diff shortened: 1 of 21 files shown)


ui/src/views/function-lib/component/EditAvatarDialog.vue
```diff
@@ -116,16 +113,30 @@ function submit() {
...
+    height: 100%;
+    width: 100%;
+  }
+  .el-radio__label {
+    width: 100%;
+    margin-top: -8px;
+    line-height: 30px;
+  }
+}
+</style>
```
shaohuzhang1, line 142: The code looks mostly correct for functionality but could benefit from some optimizations and improved styles. Here's a review of the file:

### Optimizations

1. **Imports and Exports**: The imports and exports can be slightly optimized by reducing redundancy. For example, `functionLibApi` and `cloneDeep` are used multiple times; only importing them once is sufficient.

2. **Template Ref**: Ensure that there's no chance of using an undefined ref in template operations. This can prevent runtime errors if the DOM has been manipulated or if the component re-renders unexpectedly.

3. **CSS Styling**: While existing styles are fine, consider refining margins and padding to ensure consistency with other elements.

### Potential Issues

1. **Missing Template Ref Check**: Before accessing `fileURL` and other props, it should be checked to ensure they exist to avoid errors like "Cannot read properties of null/undefined".

2. **Lack of Default Value for radioType**: If `radioType` might not have a default value when initially loaded, it would be good to handle cases where it hasn't changed after initial setting.

3. **Styling Concerns**:
   - `.radio-block-avatar`: Applying styles like `display: block` on a div doesn’t make sense since radios typically don't wrap themselves into blocks. Instead, you may need to adjust your layout strategy.
   - `.el-radio-label` heights and alignment adjustments: These might cause visual issues depending on context. Review how they affect spacing within each radio option.

### Style Suggestions

1. **Consistent Spacing and Vertical Alignment**: Apply consistent vertical spacing around labels within the radio group to improve readability.

2. **Clear CSS Classes**: Use meaningful class names such as `logo-edit`, `avatar-editor`, etc., to enhance code maintainability.

Here’s an updated version incorporating these suggestions:

```typescript
<script setup lang="ts">
// Import necessary functions and types
import { ref, watch } from 'vue';
import functionLibApi from '@/api/function-lib';
import cloneDeep from 'lodash/cloneDeep';
import { msgError, msgSuccess } from '@/utils/message';
import { defaultIcon, isAppIcon } from '@/utils/application';

const emit = defineEmits(['refresh']);
const dialogVisible = ref(false);
const detail = ref(null); // Assume this is properly initialized elsewhere
let iconFile = ref(null);

watch(() => detail.value?.id, (newId) => {
  if (iconFile.value && fileURL.value.includes(newId)) return; // Avoid re-uploading same file
  fetchDefaultAvatar();
});

async function initForm(data) {
  if (!data?.url || !isAppIcon(data.url.split('/').at(-1))) await fetchCustomAvatar(true);
}

if (!detail.value) initForm();

function fetchCustomAvatar(emitRefresh = false) {
  // Logic to fetch custom avatar URL based on data fetched from backend service
  const url = '/path/to/avatar'; // Replace with actual API endpoint
  fileURL.value = url;
  if (emitRefresh) emit('refresh');
}

function fetchDefaultAvatar() {
  // Logic to fetch default avatar based on application details
}

const radioType = ref("default");
const loading = ref(false);

function onChange(file: FileList | null) {
  if (!file) return;
  iconFile.value = file[0];
}

function submit() {
  if (radioType.value === 'default') {
    fetchCustomAvatar();
  } else if (radioType.value === 'custom' && iconFile.value) {
    let fd = new FormData();
    fd.append('file', iconFile.value.raw);
    functionLibApi.putFunctionLibIcon(detail.value.id, fd, loading)
      .then((res: any) => {
        emit('refresh', res.data);
        dialogVisible.value = false;
      });
  } else {
    msgError(t('views.applicationOverview.appInfo.EditAvatarDialog.uploadImagePrompt'));
  }
}

defineExpose({ open });
</script>

<style lang="scss" scoped>
.logo-edit {
  width: 100%;
  max-width: 20rem;
  overflow hidden;

  & > * {
    margin-bottom: 1em;
  }

  label.el-radio__label {
    display: block;
    font-weight: normal;
    line-height: 1.42857;
    color: rgba(0, 0, 0, 0.65);
    white-space: pre-wrap;
  }

  input[type="radio"] {
    position: absolute;
    opacity: 0;
    pointer-events: none;
  }

  .upload-btn {
    cursor: pointer;
  }

  .app-avatar-container {
    border-radius: var(--border-radius-default-smaller);
	margin:auto;
		width:100%;
		max-width: 30vw;
  }
  
  img {
	padding-top:10%;
	height:unset;
  }
}
</style>
```

In summary, the changes include optimizing imports & exporting, adding proper checks before accessing properties, improving styling conventions, and providing more detailed explanations for certain behaviors.

PR #2701, opened 2025-03-27
fix: update option list search logic to handle string and array form_data values

shaohuzhang1:
fix: update option list search logic to handle string and array form_data values  --bug=1053919 --user=刘瑞斌 【函数库】非必填启动参数开启启动参数后再次编辑多选置空 https://www.tapd.cn/57709429/s/1677136

### DIFF
--- a/ui/src/components/dynamics-form/index.vue
+++ b/ui/src/components/dynamics-form/index.vue
@@ -181,7 +181,13 @@ const render = (
         if (form_data[item.field] !== undefined) {
           if (item.value_field && item.option_list && item.option_list.length > 0) {
             const value_field = item.value_field
-            const find = item.option_list?.find((i) => i[value_field] === form_data[item.field])
+            const find = item.option_list?.find((i) => {
+              if (typeof form_data[item.field] === 'string') {
+                return i[value_field] === form_data[item.field]
+              } else {
+                return form_data[item.field].indexOf([value_field]) === -1
+              }
+            })
             if (find) {
               return { [item.field]: form_data[item.field] }
             }


ui/src/components/dynamics-form/index.vue
```diff
@@ -181,7 +181,13 @@ const render = (
...
+            const find = item.option_list?.find((i) => {
+              if (typeof form_data[item.field] === 'string') {
+                return i[value_field] === form_data[item.field]
+              } else {
+                return form_data[item.field].indexOf([value_field]) === -1
+              }
+            })
             if (find) {
               return { [item.field]: form_data[item.field] }
             }
```
shaohuzhang1, line 193: The code looks generally correct, but there are a few things to consider for optimization:

1. **Code Style**: The use of `const` for local variables is good practice. However, TypeScript types should be explicitly defined when using optional chaining (`?.`) because it can sometimes lead to unexpected `undefined` values.

2. **Type Checking**: Ensure that the types of inputs to your function and methods are correctly handled. If `'string'` is expected and you want more complex string matching logic (e.g., substring search), this could benefit from additional type checks or functions like `String.includes`.

3. **Performance Considerations**: Since your current implementation iterates over all options to find a match based on a single criteria, this might not scale well with large lists. Consider sorting the list or creating an index if possible.

   ```javascript
   // Example of potential performance improvement
   const optionListIndex = new Map(item.option_list.map(opt => [opt[value_field], opt]));

   if (optionListIndex.has(form_data[item.field])) {
     return { [item.field]: form_data[item.field] };
   }

4. **Functionality Clarification**: Currently, if no match is found, the returned object will be empty `{}`. For users expecting some indication, you might want to add logging or display messages indicating that no valid option was found.

Based on these points, here's a slightly improved version:

```javascript
/**
 * Renders data based on form inputs.
 * @param form_data - An object containing form input data.
 * @param items - An array of configuration objects defining how to render each field.
 * @returns A JavaScript object representing the rendered data.
 */
function render(form_data, items) {
  let result = {};

  items.forEach((item) => {
    if (form_data[item.field] !== undefined && item.value_field && item.option_list) {
      const value_field = item.value_field;
      const option_list = item.option_list;

      // Create index for faster lookup if needed
      const optionListIndex = new Map(option_list.map(opt => [opt[value_field], opt]));

      const findValue = () => {
        if (typeof form_data[item.field] === 'string') {
          return option_list.findIndex(i => i[value_field] === form_data[item.field]);
        } else {
          return typeof form_data[item.field][value_field] !== 'undefined';
        }
      };

      const index = findValue();
      if (index >= 0 || optionListIndex.has(form_data[item.field])) {
        result[item.field] = form_data[item.field];
      }
    }
  });

  return result;
}
```

This version adds basic functionality around indexing, which can improve performance. It also includes error handling for undefined indexes in case `indexOf` returns `-1`.

PR #2739, opened 2025-03-31
fix: Fix some knowledge base import errors

shaohuzhang1:
fix: Fix some knowledge base import errors

### DIFF
--- a/apps/dataset/serializers/common_serializers.py
+++ b/apps/dataset/serializers/common_serializers.py
@@ -40,6 +40,14 @@ def zip_dir(zip_path, output=None):
     zip.close()
 
 
+def is_valid_uuid(s):
+    try:
+        uuid.UUID(s)
+        return True
+    except ValueError:
+        return False
+
+
 def write_image(zip_path: str, image_list: List[str]):
     for image in image_list:
         search = re.search("\(.*\)", image)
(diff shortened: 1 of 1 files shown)


apps/dataset/serializers/common_serializers.py
```diff
@@ -58,6 +69,9 @@ def write_image(zip_path: str, image_list: List[str]):
                     f.write(file.get_byte())
             else:
                 r = text.replace('(/api/image/', '').replace(')', '')
+                r = r.strip().split(" ")[0]
+                if not is_valid_uuid(r):
+                    break
                 image_model = QuerySet(Image).filter(id=r).first()
                 if image_model is None:
                     break
```
shaohuzhang1, line 77: The provided Python code checks image references in a ZIP archive for URL patterns that include UUIDs and validates each one using `uuid.UUID`. The code handles regular expressions to extract URLs from images' descriptions and filters out those referencing non-valid UUIDs.

Here's a suggestion for improvement:

Add error handling around file operations to manage read/write errors more robustly. For example, handle exceptions during opening files for writing and ensure resources like the ZIP archive are properly closed after use even if an error occurs.

```python
import os
from typing import List

def zip_dir(zip_path: str, output='zip_files'):
    try:
        with zipfile.ZipFile(zip_path, mode="w") as zip_handle:
            # Add your logic here to add files to the zip-archive
    except Exception as e:
        print(f"An error occurred while creating or writing to {zip_path}: {e}")
```

Additionally, consider adding logging information at different points within the function (like when processing individual items) to help debug potential issues. This can be done using Python’s built-in logging module.

PR #2740, opened 2025-03-31
fix: Processing for document export sheet with more than 32 characters

shaohuzhang1:
fix: Processing for document export sheet with more than 32 characters

### DIFF
--- a/apps/dataset/serializers/document_serializers.py
+++ b/apps/dataset/serializers/document_serializers.py
@@ -703,6 +703,8 @@ class DocumentSerializers(ApiMixin, serializers.Serializer):
 
         @staticmethod
         def reset_document_name(document_name):
+            if document_name is not None:
+                document_name = document_name.strip()[0:29]
             if document_name is None or not Utils.valid_sheet_name(document_name):
                 return "Sheet"
             return document_name.strip()


apps/dataset/serializers/document_serializers.py
```diff
@@ -703,6 +703,8 @@ def merge_problem(paragraph_list: List[Dict], problem_mapping_list: List[Dict],
 
         @staticmethod
         def reset_document_name(document_name):
+            if document_name is not None:
+                document_name = document_name.strip()[0:29]
             if document_name is None or not Utils.valid_sheet_name(document_name):
                 return "Sheet"
             return document_name.strip()
```
shaohuzhang1, line 710: The function `reset_document_name` has a few improvements that can be made to enhance its readability and potentially reduce redundancy:

1. **Stripping Whitespace**: The line `document_name.strip()` is redundant inside the condition `if document_name is not None`, as it would have been stripped anyway before entering the conditional.

2. **Truncating Document Name**: While trimming the string to 29 characters might be useful for formatting in some contexts, it should be done based on user preferences rather than automatically changing lengths throughout the application's execution.

```python
class MergeProblemHelper:
    @staticmethod
    def reset_document_name(document_name):
        """Resets the document name with specific rules."""
        # Strip leading and trailing whitespace
        document_name = document_name.strip()
        
        # Check if document name is valid
        if document_name is None or not Utils.valid_sheet_name(document_name):
            return "Sheet"
        
        # Truncate document name (optional), e.g., for filename purposes
        document_name = document_name[:29].strip()
        
        return document_name
    
    # ... other static methods ...
```

This refactoring makes each step clear and removes unnecessary logic, maintaining the original functionality while being more efficient and easier to understand.

PR #2746, opened 2025-03-31
fix: When uploading files during application dialogue, there are special whitespace characters and file name parsing errors

shaohuzhang1:
fix: When uploading files during application dialogue, there are special whitespace characters and file name parsing errors

### DIFF
--- a/ui/src/components/ai-chat/component/chat-input-operate/index.vue
+++ b/ui/src/components/ai-chat/component/chat-input-operate/index.vue
@@ -392,28 +392,36 @@ const uploadFile = async (file: any, fileList: any) => {
     .then((response) => {
       fileList.splice(0, fileList.length)
       uploadImageList.value.forEach((file: any) => {
-        const f = response.data.filter((f: any) => f.name === file.name)
+        const f = response.data.filter(
+          (f: any) => f.name.replaceAll(' ', '') === file.name.replaceAll(' ', '')
+        )
         if (f.length > 0) {
           file.url = f[0].url
           file.file_id = f[0].file_id
         }
       })
       uploadDocumentList.value.forEach((file: any) => {
-        const f = response.data.filter((f: any) => f.name === file.name)
+        const f = response.data.filter(
+          (f: any) => f.name.replaceAll(' ', '') == file.name.replaceAll(' ', '')
+        )
         if (f.length > 0) {
           file.url = f[0].url
           file.file_id = f[0].file_id
         }
       })
       uploadAudioList.value.forEach((file: any) => {
-        const f = response.data.filter((f: any) => f.name === file.name)
+        const f = response.data.filter(
+          (f: any) => f.name.replaceAll(' ', '') === file.name.replaceAll(' ', '')
+        )
         if (f.length > 0) {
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/chat-input-operate/index.vue
```diff
@@ -392,28 +392,36 @@ const uploadFile = async (file: any, fileList: any) => {
...
         }
       })
       uploadVideoList.value.forEach((file: any) => {
-        const f = response.data.filter((f: any) => f.name === file.name)
+        const f = response.data.filter(
+          (f: any) => f.name.replaceAll(' ', '') === file.name.replaceAll(' ', '')
+        )
         if (f.length > 0) {
           file.url = f[0].url
           file.file_id = f[0].file_id
```
shaohuzhang1, line 427: The provided code contains two main areas for improvement:

1. **Whitespace Handling**: The comparison logic includes `.replaceAll(' ', '')` multiple times within each filter call. In JavaScript (`replace()` does not have an option to remove multiple spaces with a single space), this will result in unexpected behavior, potentially matching filenames that contain more than one consecutive space instead of removing them.

   **Suggestion**: Remove the redundant `replaceAll(' ', '')` calls to improve clarity and correctness.

2. **Potential Issues**: Although the current usage seems correct given the context (searching by exact filename including whitespace, which could be necessary), it's generally considered good practice to avoid unnecessary string operations when searching for matches.

**Optimization Suggestions**:
- Consider using regular expressions with a pattern like `\s*` to match optional whitespace around the filenames.
  
Here’s how you might update the code:

```javascript
const uploadFile = async (file: any, fileList: any) => {
  // ... rest of the function remains unchanged ...
  uploadImageList.value.forEach((file: any) => {
    const f = response.data.filter((f: any) => f.name === file.name);
    if (f.length > 0) {
      file.url = f[0].url;
      file.file_id = f[0].file_id;
    }
  });
  uploadDocumentList.value.forEach((file: any) => {
    const f = response.data.filter((f: any) => f.name.match(/^.*\s*$|^$|.*\d$/));
    if (f.length > 0) {
      file.url = f[0].url;
      file.file_id = f[0].file_id;
    }
  });
  uploadAudioList.value.forEach((file: any) => {
    const f = response.data.filter((f: any) => f.name.match(/^.*\s*$|^$|.*\d$/));
    if (f.length > 0) {
      file.url = f[0].url;
      file.file_id = f[0].file_id;
    }
  });
  uploadVideoList.value.forEach((file: any) => {
    const f = response.data.filter((f: any) => f.name.match(/^.*\s*$|^$|.*\d$/));
    if (f.length > 0) {
      file.url = f[0].url;
      file.file_id = f[0].file_id;
    }
  });
};
```

In this updated version, we use a simplified regex pattern `^.*\s*$|^$|.*\d$` to account for files where there may be some content mixed with whitespace at various positions or digits at the end of the filename. This assumes that certain formats like PDFs containing text (PDF/A) or audio/video files ending with numbers are common cases you want to handle gracefully. Adjust the pattern based on your specific requirements and data distribution.

PR #2754, opened 2025-03-31
feat: function field form support sort

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/function-lib/component/FunctionFormDrawer.vue
+++ b/ui/src/views/function-lib/component/FunctionFormDrawer.vue
@@ -116,7 +116,7 @@
           <el-icon class="mr-4"><Plus /></el-icon> {{ $t('common.add') }}
         </el-button>
       </div>
-      <el-table :data="form.init_field_list" class="mb-16">
+      <el-table ref="initFieldTableRef" :data="form.init_field_list" class="mb-16">
         <el-table-column prop="field" :label="$t('dynamicsForm.paramForm.field.label')">
           <template #default="{ row }">
             <span :title="row.field" class="ellipsis-1">{{ row.field }}</span>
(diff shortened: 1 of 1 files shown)


ui/src/views/function-lib/component/FunctionFormDrawer.vue
```diff
@@ -471,6 +508,9 @@ const open = (data: any) => {
   visible.value = true
   setTimeout(() => {
     showEditor.value = true
+    nextTick(() => {
+      onDragHandle()
+    })
   }, 100)
 }
```
shaohuzhang1, line 516: The provided code has some improvements and corrections that can be made:

1. **Import Statement**:
   - Remove redundant `import` statement for `Plus` component as it's already imported before.

2. **Template Ref Usage**:
   - The two reference variables (`initFieldTableRef` and `inputFieldTableRef`) are used but not referenced anywhere else in the template. This can be cleaned up by either using them to call methods like `.value.scrollTo({ top ... })`, or remove unused references if they're no longer needed.

3. **Sortable.js Initialization**:
   - The initialization of `Sortable` should only happen when the table is rendered, which is done by adding a check inside the `onLoad()` lifecycle hook or an appropriate event listener to ensure all DOM elements are fully available.

   ```javascript
   import { onMounted } from 'vue'

   onMounted(() => {
     onDragHandle()
   })
   ```

4. **Unused Imports/Variables**:
   - There might be unused imports/variables such as `UserFieldFormDialogRef` or properties that are set elsewhere in the component logic.

5. **Performance Considerations**:
   - If sorting becomes expensive due to large data sets, consider optimizing rendering performance or implementing batch updates in cases where changes are frequent.

Overall, addressing these points will help improve maintainability, usability, and potentially performance of the dynamic table interactions within your Vue application.

PR #2761, opened 2025-04-01
refactor: convert getList and getUserList functions to async for improved handling of asynchronous operations

shaohuzhang1:
refactor: conditionally render dropdown item based on init_field_list length  --bug=1054160 --user=刘瑞斌 【函数库】没有启用参数的函数，不展示启动参数按钮 https://www.tapd.cn/57709429/s/1678874<br>refactor: convert getList and getUserList functions to async for improved handling of asynchronous operations  --bug=1054152 --user=刘瑞斌 【函数库】筛选“我的”函数后刷新页面，还是显示的全部函数 https://www.tapd.cn/57709429/s/1678876

### DIFF
--- a/ui/src/views/function-lib/index.vue
+++ b/ui/src/views/function-lib/index.vue
@@ -184,6 +184,7 @@
                             {{ $t('common.copy') }}
                           </el-dropdown-item>
                           <el-dropdown-item
+                            v-if="item.init_field_list?.length > 0"
                             :disabled="item.permission_type === 'PUBLIC' && !canEdit(item)"
                             @click.stop="configInitParams(item)"
                           >
--- a/ui/src/views/function-lib/index.vue
+++ b/ui/src/views/function-lib/index.vue
@@ -509,7 +509,10 @@ function importFunctionLib(file: any) {
     })
 }
 
-function getList() {
+async function getList() {
+  if (userOptions.value?.length === 0) {
+    await getUserList()
+  }
   const params = {
     ...(searchValue.value && { name: searchValue.value }),
     ...(functionType.value && { function_type: functionType.value }),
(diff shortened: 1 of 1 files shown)


ui/src/views/function-lib/index.vue
```diff
@@ -544,28 +548,26 @@ function refresh(data: any) {
...
-  })
+  }
 }
 
 onMounted(() => {
-  getUserList()
+
 })
 </script>
 <style lang="scss" scoped>
```
shaohuzhang1, line 573: The provided code has a few potential issues that should be addressed:

1. **Incorrect Use of `async/await`**: In the `getList` function, you are using `then()` with an anonymous callback instead of returning the promise directly. This can lead to asynchronous behavior not being handled correctly.

2. **Potential Race Condition**: If multiple users attempt to fetch user data at once due to race conditions, it could result in inconsistent or duplicated user options.

3. **Code Duplication**: The duplicate code for fetching user data (`getUserList`) introduces redundancy.

Here's how these issues can be addressed:

1. **Use Correct Promise Handling**:
   Ensure that all API calls return promises and use `return new Promise(...)` if necessary. For example:
   ```typescript
   async function getUserList() {
     const res = await applicationApi.getUserList('FUNCTION', loading);
     if (res.data) {
       userOptions.value = res.data.map((item: any) => ({
         label: item.username,
         value: item.id
       }));
       if (user.userInfo) {
         const selectUserIdValue = localStorage.getItem(user.userInfo.id + 'function');
         if (selectUserIdValue && userOptions.value.find((v) => v.value === selectUserIdValue)) {
           selectUserId.value = selectUserIdValue;
         }
       }
     }
     return; // Ensure function returns undefined
   }
   ```

2. **Remove Duplicated Code**:
   Combine `getUserList` into a single call when needed within other functions like `getList`. For example:
   ```typescript
   async function getList() {
     if (userOptions.value.length === 0) {
       await getUserList();
     }

     const params = {
       ...(searchValue.value && { name: searchValue.value }),
       ...(functionType.value && { function_type: functionType.value })
     };

     // Other list-fetching logic here
   }
   ```

These changes ensure better consistency, maintainability, and correct handling of asynchronous operations.

PR #2767, opened 2025-04-01
perf: Optimize word segmentation retrieval

shaohuzhang1:
perf: Optimize word segmentation retrieval

### DIFF
--- a/apps/embedding/vector/pg_vector.py
+++ b/apps/embedding/vector/pg_vector.py
@@ -12,7 +12,9 @@ import uuid
 from abc import ABC, abstractmethod
 from typing import Dict, List
 
-from django.db.models import QuerySet
+import jieba
+from django.contrib.postgres.search import SearchVector
+from django.db.models import QuerySet, Value
 from langchain_core.embeddings import Embeddings
 
 from common.db.search import generate_sql_by_query_dict
@@ -68,7 +70,8 @@ class PGVector(BaseVectorStore):
                                     source_id=text_list[index].get('source_id'),
                                     source_type=text_list[index].get('source_type'),
                                     embedding=embeddings[index],
-                                    search_vector=to_ts_vector(text_list[index]['text'])) for index in
+                                    search_vector=SearchVector(Value(to_ts_vector(text_list[index]['text'])))) for
+                          index in
                           range(0, len(texts))]
         if not is_the_task_interrupted():
             QuerySet(Embedding).bulk_create(embedding_list) if len(embedding_list) > 0 else None
--- a/apps/common/util/ts_vecto_util.py
+++ b/apps/common/util/ts_vecto_util.py
@@ -12,9 +12,6 @@ from typing import List
 
 import jieba
 import jieba.posseg
-from jieba import analyse
-
-from common.util.split_model import group_by
 
 jieba_word_list_cache = [chr(item) for item in range(38, 84)]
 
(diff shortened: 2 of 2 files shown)


apps/embedding/vector/pg_vector.py
```diff
@@ -68,7 +70,8 @@ def _batch_save(self, text_list: List[Dict], embedding: Embeddings, is_the_task_
                                     source_id=text_list[index].get('source_id'),
                                     source_type=text_list[index].get('source_type'),
                                     embedding=embeddings[index],
-                                    search_vector=to_ts_vector(text_list[index]['text'])) for index in
+                                    search_vector=SearchVector(Value(to_ts_vector(text_list[index]['text'])))) for
+                          index in
                           range(0, len(texts))]
         if not is_the_task_interrupted():
             QuerySet(Embedding).bulk_create(embedding_list) if len(embedding_list) > 0 else None
```
shaohuzhang1, line 77: There are no obvious irregularities in the provided code snippet. However, there are a few suggestions for improvement:

1. **Jieba Import**: The import of `jieba` at line 16 seems unnecessary since it's not used anywhere within the function `_batch_save`. You can remove this line to simplify the code.

2. **Use of `SearchVector` Correctly**: In the SQL query generation logic starting from line 68, you're trying to create a `SearchVector` using an `EmbeddedFunctionCall`, which might not be syntactically correct depending on your database setup. Consider whether you need to use Django's built-in support for searching (e.g., PostgreSQL's full-text search capabilities).

3. **Empty Dictionary Check**: Ensure that each dictionary entry has the necessary keys `'source_id'`, `'source_type'`, and `'text'`. Using conditional statement like `index < len(text_list)` instead of assuming `len(text_lists) == len(texts)` is more robust, especially if other parameters might change in future versions.

4. **Error Handling**: There isn't any error handling in the function, which could make debugging more difficult. Adding try-except blocks around operations that interact with your database can help catch errors early.

Here is the revised version based on these points:

```python
from abc import ABC, abstractmethod
from typing import Dict, List

from langchain_core.embeddings import Embeddings

import jieba
from django.contrib.postgres.search import SearchVector
from django.db.models import Q, Value

def _batch_save(self, texts: List[Dict], embeddings: List[Embeddings], is_the_task_interrupted):
    embedding_list = [(Q(source_id=text.get('source_id'), source_type=text.get('source_type')),
                       text['text']) for text in texts]
    
    # Filter out empty entries
    filtered_embeddings = [emb for emb in embedding_list if all(val is not None for val in emb)]
    
    if not is_the_task_interrupted():
        embedding_objects = list(Embedding.from_query(Q(**{' OR '.join(f"{key}={value}" for key, value in q.as_q.items())) 
                                                     for q, _ in filtered_embeddings))
        
        batch_create_data = []
        for q, text in filtered_embeddings:
            obj = embedding_objects.pop(0)
            batch_create_data.append((Value(text), obj.id,))
            
        if batch_create_data:
            EmbeddedObject.objects.bulk_update(batch_create_data, fields=['embedding', 'search_vector'])

# Example usage of filter out invalid entries
texts_with_invalid_entries = [{'id': 1, 'source_id': '', 'text': 'A good example'}]
embeddings_with_valid_entries = ['this is a test', 'another good one']
_text_and_embedding_pairs = zip(texts_with_invalid_entries, embeddings_with_valid_entries)

filtered_pairs = [(t, e) for t, e in _text_and_embedding_pairs if all([str(v) if v != '' else None for v in t.values()])]

print(filtered_pairs)
```

This should help ensure reliability and maintainability of your code.

PR #2773, opened 2025-04-02
fix: Embedding dialogue asker parameter does not take effect

shaohuzhang1:
fix: Embedding dialogue asker parameter does not take effect

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -332,7 +332,8 @@ class ApplicationSerializer(serializers.Serializer):
                                 for field in input_field_list:
                                     if field['assignment_method'] == 'api_input' and field['variable'] in params:
                                         query += f"&{field['variable']}={params[field['variable']]}"
-
+            if 'asker' in params:
+                query += f"&asker={params.get('asker')}"
             return query
 
     class AccessTokenSerializer(serializers.Serializer):


apps/application/serializers/application_serializers.py
```diff
@@ -332,7 +332,8 @@ def get_query_api_input(self, application, params):
                                 for field in input_field_list:
                                     if field['assignment_method'] == 'api_input' and field['variable'] in params:
                                         query += f"&{field['variable']}={params[field['variable']]}"
-
+            if 'asker' in params:
+                query += f"&asker={params.get('asker')}"
             return query
 
     class AccessTokenSerializer(serializers.Serializer):
```
shaohuzhang1, line 339: The provided code snippet does not contain any obvious syntax errors or major issues. However, there is one potential optimization suggestion regarding string formatting methods:

**Suggestion for Optimization**:
Instead of using multiple `+=` statements to build the query, consider using Python's built-in list comprehension within the `join()` function, which can simplify and potentially improve performance. Here’s how you could refactor this part of the code:

```python
for field in input_field_list:
    if field['assignment_method'] == 'api_input' and field['variable'] in params:
        # Using join with list compression and f-strings can make it more concise and efficient
        query.append(f"{field['variable']}={params[field['variable']]}")
if 'asker' in params:
    query.append(f"asker={params.get('asker')}")

# Joining the list into a single string separated by '&'
return "&".join(query)
```

This approach will create a list of strings, then use `&`.join() to concatenate them efficiently. This method reduces the overhead associated with concatenation operations done by `+=`, especially when dealing with many fields.

PR #2774, opened 2025-04-02
fix: siliconCloud rerank error

shaohuzhang1:
fix: siliconCloud rerank error  --bug=1054256 --user=王孝刚 【模型】添加硅基流动的重排序模型失败 https://www.tapd.cn/57709429/s/1679612

### DIFF
--- a/apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
+++ b/apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
@@ -36,26 +36,6 @@ class SiliconCloudReranker(MaxKBBaseModel, BaseDocumentCompressor):
             top_n=model_kwargs.get('top_n', 3)
         )
 
-    def __init__(
-            self, api_base: Optional[str] = None, model: Optional[str] = None, top_n=3,
-            api_key: Optional[str] = None
-    ):
-        super().__init__()
-
-        if not api_base:
-            raise ValueError(_('Please provide server URL'))
-
-        if not model:
-            raise ValueError(_('Please provide the model'))
-
-        if not api_key:
-            raise ValueError(_('Please provide the API Key'))
-
-        self.api_base = api_base
-        self.model = model
-        self.api_key = api_key
-        self.top_n = top_n
-
     def compress_documents(self, documents: Sequence[Document], query: str, callbacks: Optional[Callbacks] = None) -> \
             Sequence[Document]:
         if not documents:


apps/setting/models_provider/impl/siliconCloud_model_provider/model/reranker.py
```diff
@@ -36,26 +36,6 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
...
-            raise ValueError(_('Please provide the API Key'))
-
-        self.api_base = api_base
-        self.model = model
-        self.api_key = api_key
-        self.top_n = top_n
-
     def compress_documents(self, documents: Sequence[Document], query: str, callbacks: Optional[Callbacks] = None) -> \
             Sequence[Document]:
         if not documents:
```
shaohuzhang1, line 41: The code you provided has a few areas that need addressing:

1. **Code Duplication**: The `__init__` method is defined twice with different parameters (one uses `model`, the other uses `api_key`).

2. **Error Handling for Missing API Base**: The error handling in the `__init__` method assumes that the missing API base would always cause a value of `None`. This might not work correctly if there are other ways to determine what the missing parameter means.

3. **Variable Naming Consistency**: There seems to be inconsistency between using snake_case (`self.model`) and camelCase (`self.Model` in another part) for variable names. It's generally recommended to use consistent naming conventions throughout the codebase.

4. **Type Hinting Issues**: While type hints exist, they are mostly limited to function parameters. The types used could be more specific based on expected usage.

Here’s an optimized version of the code with some improvements:

```python
from typing import Any, Dict, List, Optional, Sequence

class YourClass:
    MAX_TOP_N = 99  # Example maximum value for top_n

    def __new__(cls, model_type=None, model_name='default_model',
                 model_credential: Dict[str, object] = {}, **kwargs):
        return cls.new_instance(
            model_type=model_type, model_name=model_name,
            model_credential=model_credential, **kwargs
        )

    @classmethod
    def new_instance(cls, model_type, model_name, model_credential: Dict[str, object], **kwargs) -> 'YourClass':
        top_n = kwargs.setdefault('top_n', cls.MAX_TOP_N)

        instance = super().__new__(cls)
        instance.init(model_type, model_name, model_credential, top_n=top_n)
        return instance

    def init(self, api_base: Optional[str] = None, model: Optional[str] = None, top_n: int = 3,
             api_key: Optional[str] = None):
        if not api_base:
            raise ValueError("Please provide server URL")
        
        if not model:
            raise ValueError("Please provide the model name")

        if not api_key:
            raise ValueError("Please provide the API key")

        self.api_base = api_base
        self.model = model
        self.api_key = api_key
        self.top_n = min(top_n, self.MAX_TOP_N)

    def compress_documents(self, documents: Sequence[Document], query: str, callbacks: Optional['Callbacks'] = None) -> Sequence[Document]:
        """Compresses a list of documents."""
        if not documents:
            return []

        # Placeholder for document compression logic
        compressed_docs = []  # Replace this with actual implementation
        for doc in documents:
            # Simulate compression process
            compressed_docs.append(doc.compress(query))
        return compressed_docs[:]
```

### Key Changes:

- **New Instance Method**: Refactored the initialization into a separate classmethod called `new_instance`.
- **Consistent Name Usage**: Changed `ModelType` to `model_type` for consistency.
- **Default Values**: Added default values for `TopN` which helps reduce duplication.
- **Typing**: Made `max_top_n` an explicit property for better readability and type hinting.

This should make the code cleaner and easier to maintain while adhering to best practices for Python coding standards. Let me know if you have any further questions!

PR #2777, opened 2025-04-02
fix: Unfiltered special characters in voice playback

shaohuzhang1:
fix: Unfiltered special characters in voice playback

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -99,7 +99,7 @@
   </div>
 </template>
 <script setup lang="ts">
-import { nextTick, onMounted, ref } from 'vue'
+import { nextTick, onMounted, ref, onBeforeUnmount } from 'vue'
 import { useRoute } from 'vue-router'
 import { copyClick } from '@/utils/clipboard'
 import applicationApi from '@/api/application'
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -526,6 +526,11 @@ onMounted(() => {
...
   })
 })
+onBeforeUnmount(() => {
+  if (audioManage.value) {
+    audioManage.value.pause()
+  }
+})
 </script>
 <style lang="scss" scoped>
 @media only screen and (max-width: 430px) {
```
shaohuzhang1, line 536: The provided code does not have any immediate irregularities or issues that need correction. However, there are a few optimizations you might consider:

1. **Debounce Copy Click**: The `copyClick` function in the script should be debounced to prevent excessive clipboard operations during rapid clicks.

```typescript
import debounce from 'lodash/debounce';

const handleCopyClick = debounce((text: string) => {
  copy(text);
}, 300);

const bindEventListeners = () => {
  document.querySelectorAll('.copy-button').forEach(button => {
    button.addEventListener('click', (e) => {
      e.preventDefault();
      const targetElement = e.target;
      let text;

      if (targetElement.tagName === 'SPAN') {
        text = targetElement.textContent || '';
      }

      if (!text || typeof text !== "string") return;

      handleCopyClick(text);
    });
  });
};

onMounted(bindEventListeners);
```

This will limit how often the clipboard operation is triggered when multiple clicks occur quickly.

2. **Remove Unnecessary Import for `nextTick`**:
Since `nextTick` is imported but never used within this scope, it can be removed for better readability and performance.

```typescript
+// import { nextTick } from 'vue';
```

By making these minor adjustments, the code will perform more efficiently without compromising functionality.

PR #2783, opened 2025-04-02
perf: Optimize application setting switch button control

wangdan-fit2cloud:

### DIFF
--- a/ui/src/views/application-overview/index.vue
+++ b/ui/src/views/application-overview/index.vue
@@ -54,7 +54,7 @@
                   inline-prompt
                   :active-text="$t('views.applicationOverview.appInfo.openText')"
                   :inactive-text="$t('views.applicationOverview.appInfo.closeText')"
-                  @change="changeState($event)"
+                  :before-change="() => changeState(accessToken.is_active)"
                 />
               </div>
 
@@ -148,7 +148,12 @@
           {{ $t('views.applicationOverview.monitor.monitoringStatistics') }}
         </h4>
         <div class="mb-16">
-          <el-select v-model="history_day" class="mr-12" @change="changeDayHandle" style="width:180px">
+          <el-select
+            v-model="history_day"
+            class="mr-12"
+            @change="changeDayHandle"
+            style="width: 180px"
+          >
             <el-option
               v-for="item in dayOptions"
               :key="item.value"
@@ -331,13 +336,19 @@ function refreshAccessToken() {
 }
 function changeState(bool: Boolean) {
   const obj = {
-    is_active: bool
+    is_active: !bool
   }
-  const str = bool ? t('common.status.enableSuccess') : t('common.status.disableSuccess')
+  const str = obj.is_active ? t('common.status.enableSuccess') : t('common.status.disableSuccess')
   updateAccessToken(obj, str)
+    .then(() => {
+      return true
+    })
+    .catch(() => {
+      return false
+    })
 }
 
-function updateAccessToken(obj: any, str: string) {
+async function updateAccessToken(obj: any, str: string) {
   applicationApi.putAccessToken(id as string, obj, loading).then((res) => {
     accessToken.value = res?.data
     MsgSuccess(str)
--- a/ui/src/views/dataset/ImportDocumentDataset.vue
+++ b/ui/src/views/dataset/ImportDocumentDataset.vue
@@ -5,7 +5,7 @@
     </template>
     <div class="create-dataset__main flex" v-loading="loading">
       <div class="create-dataset__component main-calc-height">
-        <div class="upload-document p-24">
+        <div class="upload-document p-24" style="min-width: 850px">
           <h4 class="title-decoration-1 mb-8">
             {{ $t('views.document.feishu.selectDocument') }}
           </h4>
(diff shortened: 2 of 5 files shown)


ui/src/views/application-overview/index.vue
```diff
@@ -331,13 +336,19 @@ function refreshAccessToken() {
...
+    .catch(() => {
+      return false
+    })
 }
 
-function updateAccessToken(obj: any, str: string) {
+async function updateAccessToken(obj: any, str: string) {
   applicationApi.putAccessToken(id as string, obj, loading).then((res) => {
     accessToken.value = res?.data
     MsgSuccess(str)
```
shaohuzhang1, line 354: There are a few potential issues and areas for improvement in the provided code:

1. **TypeScript Type Annotations**: The parameters `id`, `obj`, and `loading` should be correctly typed to avoid runtime errors. For example:
   ```typescript
   import { ref } from 'vue';
   import { useTranslation } from 'vue-i18n';
   import { applicationApi, MsgSuccess } from '@/api';

   const accessToken = ref(null);
   const id = '';
   // Assume these values are properly initialized somewhere

   function changeState(bool: boolean) {
     const obj = {
       is_active: !bool,
     };
     const str = bool ? t('common.status.enableSuccess') : t('common.status.disableSuccess');
     updateAccessToken(obj, str)
       .then(() => true)
       .catch(() => false);
   }

   async function updateAccessToken(obj: any, str: string): Promise<boolean> {
       applicationApi.putAccessToken(id as string, obj)
         .then((res) => {
           accessToken.value = res?.data;
           resolve(msgSuccess(str));
         })
         .catch(reject);
     });
   }
   ```

2. **Event Handling**:
   - In the `<el-switch>` component, the event handler has been changed to call the `changeState` function with the current state of `accessToken`.
   - This might not be the desired behavior if you want the switch to toggle its own value based on the existing value.
   - Ensure that `accessToken` provides an accurate representation of whether it's currently active or inactive.

3. **Error Handling in `updateAccessToken`**:
   - The promise returned by `updateAccessToken` now includes asynchronous logic that could potentially throw exceptions without being caught (e.g., network failures).
   - This can lead to unhandled rejections which might cause unintended side effects in your application.

4. **Code Formatting**:
   - While the code is mostly clean, consider consistent formatting across the file using a linter like Prettier.
   - Consistent spacing around operators, brackets, and semicolons improves readability.

Here’s the updated code snippet after addressing some of these points:

```javascript
import { ref } from "vue";
import { useTranslation } from "vue-i18n";

const accessToken = ref(null);

function changeState() {
  const newObj = { is_active: !accessToken.value };
  const statusMessage = accessToken.value ? t("common.status.disableSuccess") : t('common.status.enableSuccess');

  updateAccessToken(newObj, statusMessage).then(successful =>
    successful &&MsgSuccess(statusMessage)
  ).catch(() => null); // Optionally handle failure with console warning or other message
}

async function updateAccessToken(obj: { is_active: boolean }, str: string): Promise<boolean> {
  try {
    const response = await applicationApi.putAccessToken("" + id, obj);
    accessToken.value = response.data;
    await msgSuccess(str);
    return true;
  } catch (error) {
    // Handle error log or notification here
    console.error("Failed to update access token:", error);
    return false;
  }
}
```
This version introduces proper TypeScript type annotations for better clarity and safety. It also uses `async/await` syntax for cleaner handling of asynchronous operations and ensures all promises either resolve successfully or gracefully fail.

ui/src/views/dataset/ImportDocumentDataset.vue
```diff
@@ -41,7 +41,7 @@
                 @change="handleAllCheckChange"
               />
             </div>
-            <div style="height: calc(100vh - 430px)">
+            <div style="height: calc(100vh - 450px)">
               <el-scrollbar>
                 <el-tree
                   :props="props"
```
shaohuzhang1, line 47: The code you provided has a few potential improvements to enhance its functionality and readability:

### Improvements:

1. **Minimum Width Adjustment**:
   - The `style` attribute on the `.upload-document` div should be moved outside of the `<div class="upload-document ..."` tag, otherwise it can cause unnecessary layout shifts.

2. **Semantic HTML Element**:
   - Consider using a semantic HTML element such as `<label>` instead of just text within an anchor for accessibility purposes. For example:
     ```html
     <a href="#" @click="handleUpload">{{ $t('views.document.feishu.uploadFile') }}</a>
     ```

3. **Height Calculation**:
   - If this component is intended to have a flexible height based solely on the user's scrollable content, then calculating the height might not be necessary unless it affects other components indirectly. Check if there are specific constraints that would benefit from adjusting this value.

4. **Consistent Spelling**:
   - Ensure consistent spelling (e.g., "calculate", "calculation") throughout the templates and variables.

Here’s how the revised version might look:

```vue
<template>
  <!-- Rest of the template remains unchanged -->
</template>

<script>
export default {
  data() {
    return {
      loading: false,
      // Tree props here
    };
  },
  methods: {
    handleAllCheckChange(val) { /* Your method implementation */ },
    handleUpload() {/* Your upload handling logic */}
  }
};
</script>

<style scoped>
.create-dataset__main.flex {
  position: relative;
}

.upload-document.p-24 {
  min-width: 850px; /* Keep inside the tag */
}

/*
... rest of styles ...
*/
</style>
```

These changes should improve the organization and potentially enhance usability without introducing new bugs or performance issues. Always ensure to thoroughly test these changes when deploying them into production environments.

PR #2787, opened 2025-04-02
fix: The interface for obtaining model metadata cannot access public models

shaohuzhang1:
fix: The interface for obtaining model metadata cannot access public models

### DIFF
--- a/apps/setting/serializers/provider_serializers.py
+++ b/apps/setting/serializers/provider_serializers.py
@@ -313,6 +313,7 @@ class ModelSerializer(serializers.Serializer):
             return ModelSerializer.model_to_dict(model)
 
         def one_meta(self, with_valid=False):
+            model = None
             if with_valid:
                 super().is_valid(raise_exception=True)
                 model = QuerySet(Model).filter(id=self.data.get("id")).first()
@@ -320,7 +321,8 @@ class ModelSerializer(serializers.Serializer):
                     raise AppApiException(500, _('Model does not exist'))
                 if model.permission_type == 'PRIVATE' and str(model.user_id) != str(self.data.get("user_id")):
                     raise Exception(_('No permission to use this model') + f"{model.name}")
-            model = QuerySet(Model).get(id=self.data.get('id'))
+            if model is None:
+                model = QuerySet(Model).get(id=self.data.get('id'))
             return {'id': str(model.id), 'provider': model.provider, 'name': model.name, 'model_type': model.model_type,
                     'model_name': model.model_name,
                     'status': model.status,


apps/setting/serializers/provider_serializers.py
```diff
@@ -313,14 +313,16 @@ def one(self, with_valid=False):
...
                 if model is None:
                     raise AppApiException(500, _('Model does not exist'))
                 if model.permission_type == 'PRIVATE' and str(model.user_id) != str(self.data.get("user_id")):
                     raise Exception(_('No permission to use this model') + f"{model.name}")
-            model = QuerySet(Model).get(id=self.data.get('id'))
+            if model is None:
+                model = QuerySet(Model).get(id=self.data.get('id'))
             return {'id': str(model.id), 'provider': model.provider, 'name': model.name, 'model_type': model.model_type,
                     'model_name': model.model_name,
                     'status': model.status,
```
shaohuzhang1, line 328: The provided code snippet has a few potential issues and optimizations:

1. **Duplicate `if` block**:
   ```python
   if model is None:
       model = QuerySet(Model).get(id=self.data.get('id'))
   ```
   This duplicated `if` condition can be removed. It's already covered by the outer `if with_valid: ...`. If `with_valid` is true, it will have checked for existence earlier.

2. **String comparison efficiency**:
   ```python
   if str(model.user_id) != str(self.data.get("user_id")):
   ```
   String comparisons like this should be avoided unless absolutely necessary because they involve type conversion. Instead, you could compare the actual values directly:
   ```python
   if model.user_id != self.data.get("user_id"):
   ```

3. **Code readability**:
   - The code uses different spaces between operators (`==`, `!=`) which might look inconsistent.
   - Adding consistent spacing can improve readability.

4. **Redundant checks**:
   ```python
   if model is None:
       raise AppApiException(500, _('Model does not exist')
   ```
   You don't always need to raise an exception when a model doesn't exist; simply returning `None` or handling it gracefully might be better depending on its context.

Here’s a revised version of the function with these improvements:

```python
@@ -313,18 +313,21 @@ def one(self, with_valid=False):
             return ModelSerializer.model_to_dict(model)

         def one_meta(self, with_valid=False):
             if with_valid:
                 model_id = self.data.get('id')
                 try:
                     model = Model.objects.filter(id=model_id).first()
                     if not model:
                         raise AppApiException(500, _('Model does not exist'))
                     if model.permission_type == 'PRIVATE' and model.user_id != self.data.get("user_id"):
                         raise Exception(_('No permission to use this model') + f" ({model.name})")
                 except Exception as e:
                     handle_exception(e)
                 model_id = self.data.get('id')
                 
             if isinstance(model_id, int):  # Ensure model_id is an integer
                 model = Model.objects.get(id=model_id)
             else:
                 raise ValueError(_('Invalid model ID'))

             return {
                 'id': str(model.id),
                 'provider': model.provider,
                 'name': model.name,
                 'model_type': model.model_type,
                 'model_name': model.model_name,
                 'status': model.status
             }
```

### Key points:
- Removed unnecessary duplicate `if` blocks.
- Simplified string comparison by avoiding type conversion unless needed.
- Added additional logic to ensure that `self.data['id']` is an integer before calling `QuerySet.get()`.
- Improved overall code readability and maintainability through consistent formatting and proper separation of concerns.

PR #2789, opened 2025-04-03
fix: After the page is redirected, the voice playback does not stop

shaohuzhang1:
fix: After the page is redirected, the voice playback does not stop

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -539,6 +539,9 @@ onBeforeUnmount(() => {
   if (audioManage.value) {
     audioManage.value.pause()
   }
+  if (window.speechSynthesis) {
+    window.speechSynthesis.cancel()
+  }
 })
 </script>
 <style lang="scss" scoped>


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -539,6 +539,9 @@ onBeforeUnmount(() => {
   if (audioManage.value) {
     audioManage.value.pause()
   }
+  if (window.speechSynthesis) {
+    window.speechSynthesis.cancel()
+  }
 })
 </script>
 <style lang="scss" scoped>
```
shaohuzhang1, line 547: The provided code has several optimizations and improvements that can be made:

```plaintext
- The `onBeforeUnmount` hook is used to perform cleanup operations when the component is unmounted. Adding checks for `window.speechSynthesis` allows for graceful handling of speech synthesis when the application is closed.

- Ensure that any resources managed by the `AudioManage`, such as media streams or timers, are released within this function to prevent memory leaks. However, since you only see the `pause` method being called on it, no explicit cleanup might be necessary unless there are other related methods or properties that need attention.

Here is the improved version with comments:

```html
<script setup lang="ts">
import { ref } from 'vue'
import AudioManage from './components/AudioManage.vue' // Assuming AudioManage is a custom component

const audioManage = ref<AudioManage>()

onMounted(async () => {
  await new Promise(resolve => setTimeout(resolve, 500)) // Example asynchronous task
})

onBeforeUnmount(() => {
  if (audioManage.value && typeof audioManage.value.pause === 'function') {
    audioManage.value.pause()
  }

  window.speechSynthesis?.cancel() // Safely cancel speech synthesis regardless of availability
})
</script>

<style scoped lang="scss">
// Add styles scoped here...
</style>
```

Make sure to replace `'./components/AudioManage.vue'` with the actual path to your component file and adjust the example asynchronous task as needed. Additionally, consider adding more robust error handling around any external APIs or components to improve user experience during runtime errors.

PR #2803, opened 2025-04-07
fix: limit chapter title length to 256 characters in pdf_split_handle.py

shaohuzhang1:
fix: limit chapter title length to 256 characters in pdf_split_handle.py  --bug=1054363 --user=刘瑞斌 【知识库】导入PDF文档，分段标题长度超长时，没有自动截断 https://www.tapd.cn/57709429/s/1681044

### DIFF
--- a/apps/common/handle/impl/pdf_split_handle.py
+++ b/apps/common/handle/impl/pdf_split_handle.py
@@ -173,14 +173,15 @@ class PdfSplitHandle(BaseSplitHandle):
 
             # Null characters are not allowed.
             chapter_text = chapter_text.replace('\0', '')
-
+            # 限制标题长度
+            real_chapter_title = chapter_title[:256]
             # 限制章节内容长度
             if 0 < limit < len(chapter_text):
                 split_text = PdfSplitHandle.split_text(chapter_text, limit)
                 for text in split_text:
-                    chapters.append({"title": chapter_title, "content": text})
+                    chapters.append({"title": real_chapter_title, "content": text})
             else:
-                chapters.append({"title": chapter_title, "content": chapter_text if chapter_text else chapter_title})
+                chapters.append({"title": real_chapter_title, "content": chapter_text if chapter_text else real_chapter_title})
             # 保存章节内容和章节标题
         return chapters
 


apps/common/handle/impl/pdf_split_handle.py
```diff
@@ -173,14 +173,15 @@ def handle_toc(doc, limit):
...
             if 0 < limit < len(chapter_text):
                 split_text = PdfSplitHandle.split_text(chapter_text, limit)
                 for text in split_text:
-                    chapters.append({"title": chapter_title, "content": text})
+                    chapters.append({"title": real_chapter_title, "content": text})
             else:
-                chapters.append({"title": chapter_title, "content": chapter_text if chapter_text else chapter_title})
+                chapters.append({"title": real_chapter_title, "content": chapter_text if chapter_text else real_chapter_title})
             # 保存章节内容和章节标题
         return chapters
```
shaohuzhang1, line 187: The code appears to be functioning correctly based on initial inspection, but here are a few suggestions and corrections:

1. **Null Characters Check**: The line `chapter_text = chapter_text.replace('\0', '')` is correct and will remove null characters from the `chapter_text`.

2. **Title Length Limitation**: Adding a character limit of 256 for `real_chapter_title` is redundant since it's already truncated to ensure length within the function, which should be sufficient unless longer titles are required.

3. **Content Handling with Empty Chapter Titles**: There seems to be an unnecessary condition in appending content when the content is empty. When chapter_text is empty, appending `chapter_title if chapter_text else ""` would work fine without explicitly checking its length again. However, there might be intended behavior where you want to append a placeholder or message if the content doesn't exist.

4. **Code Consistency**: Ensure that all similar lines follow the same format (e.g., spacing around operators), for better readability.

Overall, the code logic is sound, but these minor adjustments could improve clarity and maintainability. Here’s a slightly refined version of the relevant section:

```python
def handle_toc(doc, limit):
    # Null characters are not allowed.
    chapter_text = chapter_text.replace('\0', '')

    # Initialize real_chapter_title to avoid repetition
    if limit > 0:
        real_chapter_title = chapter_title[:limit]

    chapters = []
    
    # If the chapter title exists, add it as the key; otherwise, use 'Unknown'
    # For simplicity, let's assume that None or empty strings are meant to have no entry at all
    if chapter_title:
        # Split the text into paragraphs based on the specified limit
        split_text = PdfSplitHandle.split_text(chapter_text, limit)
        
        for text in split_text:
            chapters.append({
                "title": real_chapter_title,
                "content": text.strip()  # Strip leading/trailing whitespace from each paragraph
            })
    
    return chapters
```

These changes make the code cleaner and more consistent.

PR #2805, opened 2025-04-07
fix: The local rearrangement model cannot be loaded

shaohuzhang1:
fix: The local rearrangement model cannot be loaded


apps/setting/models_provider/impl/local_model_provider/model/reranker.py
```diff
@@ -69,7 +69,7 @@ class LocalBaseReranker(MaxKBBaseModel, BaseDocumentCompressor):
     tokenizer: Any = None
     model: Optional[str] = None
     cache_dir: Optional[str] = None
-    model_kwargs: ClassVar = {}
+    model_kwargs: Any = {}
 
     def __init__(self, model_name, cache_dir=None, **model_kwargs):
         super().__init__()
```
shaohuzhang1, line 75: The code has a minor issue with incorrect typing for `model_kwargs`. It needs to be `ClassVar` rather than `Any`, but since they are both used interchangeably here, it doesn't cause runtime errors if you choose `Any`.

```python
@@ -72,8 +72,8 @@
         tokenizer: Any = None
         model: Optional[str] = None
         cache_dir: Optional[str] = None
-        model_kwargs: ClassVar = {} # Incorrectly defined as 'ClassVar'
+        model_kwargs: Any = {} 
```

If you want to ensure strict type checking by specifying `dict` instead of using `Any`, change both occurrences to `Dict[str, Any]:

```python
class LocalBaseReranker(BaseKBBaseModel, BaseDocumentCompressor):
    tokenizer: Any = None
    model: Optional[str] = None
    cache_dir: Optional[str] = None
    model_kwargs: Dict[str, Any] = {}  # Corrected to accept only strings as keys

    def __init__(self, model_name, cache_dir=None, **model_kwargs):
        super().__init__()
```

### Suggested Optimization:
1. **Use More Pythonic Defaults**: Consider default parameters in the `__init__` method where appropriate.
2. **Ensure Type Hint Consistency**: Always use specific types (e.g., dict vs any) and avoid unnecessary variance (`*args`, `**kwargs`) unless really needed.

These suggestions make your code more readable and consistent, which can help maintainability over time.

PR #2813, opened 2025-04-07
fix: jieba 分词使用全模式

shaohuzhang1:
fix: jieba 分词使用全模式

### DIFF
--- a/apps/common/util/ts_vecto_util.py
+++ b/apps/common/util/ts_vecto_util.py
@@ -78,11 +78,11 @@ def get_key_by_word_dict(key, word_dict):
 
 def to_ts_vector(text: str):
     # 分词
-    result = jieba.lcut(text)
+    result = jieba.lcut(text, cut_all=True)
     return " ".join(result)
 
 
 def to_query(text: str):
-    extract_tags = jieba.lcut(text)
+    extract_tags = jieba.lcut(text, cut_all=True)
     result = " ".join(extract_tags)
     return result


apps/common/util/ts_vecto_util.py
```diff
@@ -78,11 +78,11 @@ def get_key_by_word_dict(key, word_dict):
...
-    result = jieba.lcut(text)
+    result = jieba.lcut(text, cut_all=True)
     return " ".join(result)
 
 
 def to_query(text: str):
-    extract_tags = jieba.lcut(text)
+    extract_tags = jieba.lcut(text, cut_all=True)
     result = " ".join(extract_tags)
     return result
```
shaohuzhang1, line 88: The provided code looks generally correct, but it can be optimized and clarified slightly:

1. **Function Naming**:
   - The function `to_ts_vector` is not necessary if you are using the same logic in `to_query`.
   - You might want to rename one of these functions to avoid redundancy.

2. **Regular Expressions**:
   - In both `to_ts_vector` and `to_query`, the default behavior of `jieba.lcut` (or Chinese word segmentation) involves splitting text into words without considering punctuation or other delimiters. This means that numbers, symbols, etc., will also be included in the tokenization unless they are part of a Chinese character sequence.
   
   If you specifically need to keep digits out during segmenting, consider using `jieba.lcut_for_search()` which splits on spaces and considers punctuation as separate tokens:
   
     ```python
     from jieba import lcut_for_search
 
     def to_query(text: str):
         extract_tags = lcut_for_search(text)
         result = " ".join(extract_tags)
         return result
     ```

3. **Code Readability**:
   - It's generally better practice to use descriptive variable names instead of single-letter variables like `result`.

Here’s an updated version with some of these suggestions applied:

```python
def get_key_by_word_dict(key, word_dict):
    # Your existing implementation here

def process_text_for_segmentation(text: str):
    return " ".join(jieba.lcut_for_search(text))

def to_query(text: str):
    """Convert text into query format."""
    processed_text = process_text_for_segmentation(text)
    return processed_text
```

These changes make the code more readable and maintainable while improving its functionality according to specific requirements.

PR #2814, opened 2025-04-07
fix: When the discriminator is true, there is no problem with the value

shaohuzhang1:
fix: When the discriminator is true, there is no problem with the value

### DIFF
--- a/ui/src/workflow/nodes/condition-node/index.vue
+++ b/ui/src/workflow/nodes/condition-node/index.vue
@@ -326,7 +326,7 @@ function deleteCondition(index: number, cIndex: number) {
 }
 
 function changeCondition(val: string, index: number, cIndex: number) {
-  if (val === 'is_null' || val === 'is_not_null') {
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/condition-node/index.vue
```diff
@@ -326,7 +326,7 @@ function deleteCondition(index: number, cIndex: number) {
 }
 
 function changeCondition(val: string, index: number, cIndex: number) {
-  if (val === 'is_null' || val === 'is_not_null') {
+  if (['is_null', 'is_not_null', 'is_true', 'is_not_true'].includes(val)) {
     const list = cloneDeep(props.nodeModel.properties.node_data.branch)
     list[index]['conditions'][cIndex].value = 1
     set(props.nodeModel.properties.node_data, 'branch', list)
```
shaohuzhang1, line 332: The code modification seems to be correcting an error in checking condition values. The updated `changeCondition` function now correctly checks against a list of allowed conditions ('is_null', 'is_not_null', 'is_true', 'is_not_true'). This improvement ensures that only these specific condition types can be modified.

PR #2817, opened 2025-04-08
fix: Missing fields in exported logs

shaohuzhang1:
fix: Missing fields in exported logs  --bug=1054500 --user=王孝刚 【应用】对话日志导出，没有携带用户信息 https://www.tapd.cn/57709429/s/1681984

### DIFF
--- a/apps/application/serializers/chat_serializers.py
+++ b/apps/application/serializers/chat_serializers.py
@@ -216,6 +216,7 @@ class ChatSerializers(serializers.Serializer):
                     "\n".join([
                         f"{improve_paragraph_list[index].get('title')}\n{improve_paragraph_list[index].get('content')}"
                         for index in range(len(improve_paragraph_list))]),
+                    row.get('asker').get('user_name'),
                     row.get('message_tokens') + row.get('answer_tokens'), row.get('run_time'),
                     str(row.get('create_time').astimezone(pytz.timezone(TIME_ZONE)).strftime('%Y-%m-%d %H:%M:%S')
                         )]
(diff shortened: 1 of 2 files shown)


apps/application/serializers/chat_serializers.py
```diff
@@ -242,7 +243,8 @@ def stream_response():
                            gettext('answer'), gettext('User feedback'),
                            gettext('Reference segment number'),
                            gettext('Section title + content'),
-                           gettext('Annotation'), gettext('Consuming tokens'), gettext('Time consumed (s)'),
+                           gettext('Annotation'), gettext('USER'), gettext('Consuming tokens'),
+                           gettext('Time consumed (s)'),
                            gettext('Question Time')]
                 for col_idx, header in enumerate(headers, 1):
                     cell = worksheet.cell(row=1, column=col_idx)
```
shaohuzhang1, line 250: There are a couple of issues in the provided code:

1. **Potential Typographical Error**: The string 'USER' is not quoted in `workbook.update_cells()`, which might lead to an issue if you intend it to be treated as a cell value rather than just plain text.

2. **Missing Line Breaks in Headers**: There's no line break added after "ANNOTATION" and before the next label ("CONSUMING TOKENS"). This can make the headers harder to read on some systems.

3. **Inconsistent Header Formatting**: Some labels use commas instead of spaces to separate words, while others do not. It would be consistent if we used a single convention across all header entries.

Here’s an improved version of the code with these considerations addressed:

```python
@@ -215,6 +215,7 @@ def to_row(row: Dict):
                        f"{improve_paragraph_list[index].get('title')}\n{improve_paragraph_list[index].get('content')}"
                         for index in range(len(improve_paragraph_list))]),
+                    row.get('asker').get('user_name'),
                     row.get('message_tokens') + row.get('answer_tokens'), row.get('run_time'),
                     str(row.get('create_time').astimezone(pytz.timezone(TIME_ZONE)).strftime('%Y-%m-%d %H:%M:%S'))
                )
                 workbook.update_cells([(row_index + 2, col_idx) for col_idx, _ in enumerate(headers)], values=row_data)

@@ -242,8 +242,9 @@ def stream_response():
                                gettext('answer'), gettext('User feedback'),
                                gettext('Reference segment number'),
                                gettext('Section title + content'),
-                               gettext('Annotation'), gettext('Consuming tokens'), gettext('Time consumed (s)'),
+                               gettext('Annotation'), gettext('USER'), gettext('Consuming tokens'),
+                               gettext('TIME CONSUMED (S)'), # Consistent header formatting
                                gettext('Question Time')]
                         for col_idx, header in enumerate(headers, 1):
                             cell = worksheet.cell(row=1, column=col_idx)
```

These changes should improve readability and consistency in your spreadsheet output.

PR #2819, opened 2025-04-08
fix: Workflow debugging for authorized applications will result in an error message indicating unauthorized access to the model

shaohuzhang1:
fix: Workflow debugging for authorized applications will result in an error message indicating unauthorized access to the model

### DIFF
--- a/apps/application/views/chat_views.py
+++ b/apps/application/views/chat_views.py
@@ -97,7 +97,7 @@ class ChatView(APIView):
                              tags=[_("Application/Chat")])
         def post(self, request: Request):
             return result.success(ChatSerializers.OpenWorkFlowChat(
-                data={**request.data, 'user_id': request.user.id}).open())
+                data={'user_id': request.user.id, **request.data}).open())
 
     class OpenTemp(APIView):
         authentication_classes = [TokenAuth]
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -268,8 +268,10 @@ const openChatId: () => Promise<string> = () => {
       })
   } else {
     if (isWorkFlow(obj.type)) {
+      console.log(obj)
       const submitObj = {
-        work_flow: obj.work_flow
+        work_flow: obj.work_flow,
+        user_id: obj.user
       }
       return applicationApi.postWorkflowChatOpen(submitObj).then((res) => {
         chartOpenId.value = res.data


ui/src/components/ai-chat/index.vue
```diff
@@ -268,8 +268,10 @@ const openChatId: () => Promise<string> = () => {
...
   } else {
     if (isWorkFlow(obj.type)) {
+      console.log(obj)
       const submitObj = {
-        work_flow: obj.work_flow
+        work_flow: obj.work_flow,
+        user_id: obj.user
       }
       return applicationApi.postWorkflowChatOpen(submitObj).then((res) => {
         chartOpenId.value = res.data
```
shaohuzhang1, line 277: There is one issue in the given code that could cause problems if `obj` doesn't have a property named `work_flow`. The error would occur because of accessing `.work_flow` on `submitObj`, assuming it exists.

To prevent this from causing errors, you should add conditions to ensure both properties (`work_flow` and `user`) exist before constructing the `submitObj`.

Here’s an optimized version with these changes:

```javascript
const openChatId: () => Promise<string> = () => {
  if (isWorkFlow(obj.type)) {
    // Check if work_flow and user exist
    if (typeof obj.work_flow === 'string' && typeof obj.user !== 'undefined') {
      const submitObj = {
        work_flow: obj.work_flow,
        user_id: obj.user
      };
      return applicationApi.postWorkflowChatOpen(submitObj).then((res) => {
        chartOpenId.value = res.data;
      });
    } else {
      throw new Error('Missing required fields: work_flow or user');
    }
  } else {
    // ...
  }
}
```

With this modification, we first check if `obj.work_flow` exists as a string using `typeof` (`"string"`), and verify that `obj.user` does not equal `undefined`. If either condition fails, an error is thrown with a descriptive message. This prevents potential runtime errors and makes your function safer and more robust.

apps/application/views/chat_views.py
```diff
@@ -97,7 +97,7 @@ class OpenWorkFlowTemp(APIView):
                              tags=[_("Application/Chat")])
         def post(self, request: Request):
             return result.success(ChatSerializers.OpenWorkFlowChat(
-                data={**request.data, 'user_id': request.user.id}).open())
+                data={'user_id': request.user.id, **request.data}).open())
 
     class OpenTemp(APIView):
         authentication_classes = [TokenAuth]
```
shaohuzhang1, line 103: The provided code has a few issues:

1. In `OpenWorkFlowTemp.post()`, the serializer is instantiated with extra arguments (`data`), which might be unnecessary if they are not used in the serialization process.

2. The method is using `kwargs`, but there's no use of unpacking them into the serializer. This can be simplified without affecting the functionality.

Here’s an updated version that addresses these points:

```python
class OpenWorkFlowTemp(APIView):
    @action(detail=False, methods=['post'], url_name='chat-list', name=_('Application/Chat'), authentication_classes=[],
            permission_classes=[], tags=[_("Application/Chat")])
    def post(self, request: Request):
        chat_data = {'user_id': request.user.id}
        # Merge other data from the request.POST (if needed)
        chat_data.update(request.data)
        
        # Create the serialized object and open it
        return result.success(ChatSerializer.OpenWorkFlowChat(data=chat_data).open())

class OpenTemp(APIView):
    authentication_classes = [TokenAuth]
```

Changes made:
1. Simplified the instantiation of the serializer to pass just `'user_id'`.
2. Removed the unnecessary dictionary spread operator before `request.data`.

These changes should resolve the potential issue and make the code cleaner. If you have additional requirements or need further modifications, let me know!

PR #2826, opened 2025-04-08
fix: When switching application models, the model parameters should be set to the default values of the current model

shaohuzhang1:
fix: When switching application models, the model parameters should be set to the default values of the current model

### DIFF
--- a/ui/src/components/model-select/index.vue
+++ b/ui/src/components/model-select/index.vue
@@ -97,6 +97,7 @@ const props = defineProps<{
 const emit = defineEmits(['update:modelValue', 'change', 'submitModel'])
 const modelValue = computed({
   set: (item) => {
+    emit('change', item)
     emit('update:modelValue', item)
   },
   get: () => {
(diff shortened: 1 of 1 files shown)


ui/src/components/model-select/index.vue
```diff
@@ -127,7 +128,6 @@ const openCreateModel = (provider?: Provider, model_type?: string) => {
   if (provider && provider.provider) {
     createModelRef.value?.open(provider, model_type)
   } else {
-    console.log(model_type)
     selectProviderRef.value?.open(model_type)
   }
 }
```
shaohuzhang1, line 133: The code looks mostly regular, but here are some minor improvements:

1. Avoid printing to the console unless necessary (`console.log` is fine for development purposes).
2. Ensure that `modelType` being passed to `selectProviderRef.open()` has a meaningful value before calling it.

Overall, your function structure and logic appear to be correct with these adjustments. If you have more specific concerns or need further optimization for this particular implementation, feel free to ask!

PR #2830, opened 2025-04-09
security: fix reverse shell vulnerability in function library.

shaohuzhang1:
security: fix reverse shell vulnerability in function library.

### DIFF
--- a/apps/common/util/function_code.py
+++ b/apps/common/util/function_code.py
@@ -88,7 +88,7 @@ except Exception as e:
             os.system(f"chown {self.user}:{self.user} {exec_python_file}")
         kwargs = {'cwd': BASE_DIR}
         subprocess_result = subprocess.run(
-            ['su', '-c', python_directory + ' ' + exec_python_file, self.user],
+            ['su', '-s', python_directory, '-c', "exec(open('" + exec_python_file + "').read())", self.user],
             text=True,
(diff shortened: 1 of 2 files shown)


apps/common/util/function_code.py
```diff
@@ -88,7 +88,7 @@ def _exec_sandbox(self, _code, _id):
             os.system(f"chown {self.user}:{self.user} {exec_python_file}")
         kwargs = {'cwd': BASE_DIR}
         subprocess_result = subprocess.run(
-            ['su', '-c', python_directory + ' ' + exec_python_file, self.user],
+            ['su', '-s', python_directory, '-c', "exec(open('" + exec_python_file + "').read())", self.user],
             text=True,
             capture_output=True, **kwargs)
         os.remove(exec_python_file)
```
shaohuzhang1, line 94:
```markdown
The provided `su` command is incorrect because it uses `-s`, which changes the shell to `/bin/sh`. To execute Python directly from a file, use `'bash'` instead of specifying the shell separately.

Here's an optimized version:

```python
subprocess.run(
    ['su', '-c', f"exec(open('{exec_python_file}').read())", self.user],
    text=True,
    capture_output=True, **kwargs)
```

This assumes that you are using Linux with Bash. If your system uses a different default shell, adjust the arguments accordingly.

PR #2833, opened 2025-04-09
fix: condition node supports drag and drop

wangdan-fit2cloud:

### DIFF
--- a/ui/src/workflow/nodes/condition-node/index.vue
+++ b/ui/src/workflow/nodes/condition-node/index.vue
@@ -248,6 +248,9 @@ function onEnd(event?: any) {
   if (oldIndex === undefined || newIndex === undefined) return
   const list = cloneDeep(props.nodeModel.properties.node_data.branch)
   if (oldIndex === list.length - 1 || newIndex === list.length - 1) {
+    list[newIndex] = list[oldIndex]
+    list[oldIndex] = clonedData
+    set(props.nodeModel.properties.node_data, 'branch', list)
     return
   }
   list[newIndex].type = list[oldIndex].type


ui/src/workflow/nodes/condition-node/index.vue
```diff
@@ -248,6 +248,9 @@ function onEnd(event?: any) {
   if (oldIndex === undefined || newIndex === undefined) return
   const list = cloneDeep(props.nodeModel.properties.node_data.branch)
   if (oldIndex === list.length - 1 || newIndex === list.length - 1) {
+    list[newIndex] = list[oldIndex]
+    list[oldIndex] = clonedData
+    set(props.nodeModel.properties.node_data, 'branch', list)
     return
   }
   list[newIndex].type = list[oldIndex].type
```
shaohuzhang1, line 256: The provided function has a few potential issues and can be optimized:

### Potential Issues:
1. **Cloning Logic**: While `cloneDeep` is used, it doesn't seem to affect the original array `props.nodeModel.properties.node_data.branch`. Ensure you understand that cloning isn't necessary here if you're only modifying existing elements.

2. **Edge Case Handling**: The condition when both `oldIndex` and `newIndex` are `list.length - 1` checks whether swapping these positions would cause an index out of bounds error. This should always handle safely without further action needed since indices wrap around arrays in JavaScript.

3. **Simplified Assignment**: Instead of creating new variables (`clonedData`, `list`) just to swap values, you could directly manipulate the existing array using array methods like `splice`.

4. **Set Operation**: Using `set()` from Redux Toolkit for updating state might not be necessary unless your implementation requires managing multiple properties separately. Direct manipulation with `Object.assign()`, `spread operator`, or even mutation might be more efficient depending on your use case.

### Optimizations:
1. **Direct Swapping**:
   ```javascript
   if (newIndex === list.length - 1 && oldIndex === list.length - 1) {
     [list[newIndex], list[oldIndex]] = [list[oldIndex], list[newIndex]];
     set(props.nodeModel.properties.node_data, 'branch', list);
     return;
   }
   ```

2. **Efficient Manipulation**:
   ```javascript
   if (newIndex === oldIndex) return;

   // Create a copy of the data at the new and old indexes
   let newDataAtNew = clone(new ListAt(newNode));
   let origDataAtOld = list[oldIndex];

   // Update the current node's branch at new index
   if (origDataAtOld.type !== List.atType) throw new Error('wrong type');
   newDataAtNew.value = [...origDataAtOld.value];
   newDataAtNew.keypath = origDataAtOld.keypath.concat([List.dataKey]);
   newDataAtNew.indexInMap = map.indexOf(oldIndex);

   // Remove the copied value at new index before adding back as newList
   delete map[newListIndex];

   // Put newList into its place in map, making sure to update keypaths
   const parentKeys = getKeyPathForValue(map, origDataAtOld.value)[0].split('.');
   let prevItemKey = '';
   try {
     map.update(keypath.join(dotSep), function(item) {
       // Update item keys based on their relative position to the old path
       if (!prevItemKey || prevItemKey.split('.').some((k) => !item[keypath][k])) newItemKey += '.' + Object.keys(item)[0];
       item[keypath]['$key'] = '$$' + (++nextId++); // Generate unique ID
       prevItemKey = newItemKey;
       // Reorder items in nested lists
       if (item.$array && item.$.size > 1) reorderArray(childLists[item.$key], item.$.sortIndex ? item.$.sortIndex : childLists[item.$key].indexOf(origDataAtOld.value),
                                                           {from: prevItemKey});
       updateChildMaps(mapToUpdateOnRemove, map, '', null);

       return item;
     });
   } catch(err) {
     console.error(`${err}`);
   }

   // Place newList after last inserted element by key so they appear correctly later
   map.set(list[list.length - 1].indexInMap + listSize + 1, newItemKey);

   map.delete(prevItemKey); // Delete the previously unused key

   this._onListInsertion({ ...dataBeforeChange, updatedNodes: [{ node: newNode }] }, context);
   nextSibling.push(appendWithKey(newDataAtNew.clone(), nextSibling[nextSibling.length - 1]));
   ```

This optimized version leverages direct swapping and ensures efficient updates while handling cases where the new index matches the old one gracefully.

PR #2840, opened 2025-04-09
fix: Historical user conversation data has been cleared

shaohuzhang1:
fix: Historical user conversation data has been cleared

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -243,6 +243,11 @@ function sendMessage(val: string, other_params_data?: any, chat?: chatType) {
         showUserInput.value = true
         return
       })
+  } else {
+    showUserInput.value = false
+    if (!loading.value && props.applicationDetails?.name) {
+      handleDebounceClick(val, other_params_data, chat)
+    }
   }
 }
 
--- a/ui/src/stores/modules/user.ts
+++ b/ui/src/stores/modules/user.ts
@@ -65,7 +65,11 @@ const useUserStore = defineStore({
       if (token) {
         return token
       }
-      return localStorage.getItem(`${this.userAccessToken}accessToken`)
+      const local_token = localStorage.getItem(`${token}accessToken`)
+      if (local_token) {
+        return local_token
+      }
+      return localStorage.getItem(`accessToken`)
     },
 
     getPermissions() {


ui/src/stores/modules/user.ts
```diff
@@ -65,7 +65,11 @@ const useUserStore = defineStore({
...
       }
-      return localStorage.getItem(`${this.userAccessToken}accessToken`)
+      const local_token = localStorage.getItem(`${token}accessToken`)
+      if (local_token) {
+        return local_token
+      }
+      return localStorage.getItem(`accessToken`)
     },
 
     getPermissions() {
```
shaohuzhang1, line 75: The provided code snippet includes some minor issues that can be addressed:

### Issues Identified:
1. **Redundancy**: The function `getToken` fetches tokens from both `localStorage` items but uses different keys (``${token}accessToken`` and ``accessToken```). This redundancy is unnecessary.
2. **Variable Naming**: Using `local_token` as an alias for `localStorage.getItem(...)` does not improve readability; using `access_token_from_storage` would be more descriptive.

### Optimization Suggestions:
1. Remove the redundant logic and use consistency in fetching tokens:
   - Use only one key for accessing tokens, regardless of whether it's provided directly or stored in `localStorage`.
   
2. Ensure clarity with variable names to make the code easier to understand:

```javascript
@@ -65,14 +65,16 @@ const useUserStore = defineStore({
       let refreshTokenFromStorage

       if (refreshToken) {
         token = refreshToken
       }

-      // Fetch access token based on the provided token or 'accessToken' from storage
+      const tokenId = token || 'accessToken'
+      
       accessTokenFromStorage = localStorage.getItem(`${tokenId}AccessToken`)

       return accessTokenFromStorage ? accessTokenFromStorage : ''
+  } else {
+    // Fallback if no valid token is available
+    console.warn('No valid token found.')
+    return ''
   }
 }
 
 function getPermissions() {
```

By making these changes, the code becomes cleaner, reduces duplication, and improves its legibility and maintainability.

ui/src/components/ai-chat/index.vue
```diff
@@ -243,6 +243,11 @@ function sendMessage(val: string, other_params_data?: any, chat?: chatType) {
         showUserInput.value = true
         return
       })
+  } else {
+    showUserInput.value = false
+    if (!loading.value && props.applicationDetails?.name) {
+      handleDebounceClick(val, other_params_data, chat)
+    }
   }
 }
```
shaohuzhang1, line 253: The provided code function `sendMessage` can be optimized and enhanced as follows:

### Functionality

The function currently handles sending messages with various parameters, including user input and optional loading status. However, there are some areas that could be improved:

1. **Loading Status Handling**: The logic for deciding whether to call `handleDebounceClick` involves checking both the `loading.value` condition and `props.applicationDetails?.name`. This redundancy should be removed.

2. **Empty Message Validation**: It might be beneficial to add a validation step to ensure that the message is not empty before attempting to send it.

3. **Conditional Logic Refinement**: The conditional logic around showing user input can be simplified.

### Optimization and Recommendations

- **Combine Conditions**: Remove the redundant check for `loading.value && props.applicationDetails?.name`.
  
```typescript
function sendMessage(val: string, other_params_data?: any, chat?: chatType) {
  if (val.trim() !== "") { // Add validation to prevent empty messages
    showUserInput.value = true
  } else {
    console.log("Message cannot be empty");
    return;
  }
}
```

- **Simplify Conditional Block**: Combine the block handling when user input needs showing into one line.
  

```typescript
if (val.trim() !== "" || !showUserInput.value) {
    showUserInput.value = val.trim() !== "";
  } else {
    console.log("Message cannot be empty.");
    return;
  }
```

### Code Cleanup

Ensure the function is clean and easy to read by following PEP8 guidelines. Here’s an updated version of the function:

```javascript
function sendMessage(val: string, other_params_data?: any, chat?: chatType) {
  if (val.trim() === "") {
    console.log("Message cannot be empty."); // Add explicit error logging
    return; // Optionally you could throw an exception if needed
  }

  const isValid = val.trim().length > 0;
  showUserInput.value = isValid;

  if (!isValid) {
    console.log("Failed to set showUserInput:", isValid);
  }
}

// Additional logic for handleDebounceClick remains unchanged
handleDebounceClick(val, other_params_data, chat);
```

In summary, these improvements reduce code duplication, make the logic clearer, and add basic value-validation checks to enhance maintainability and reliability of the function.

PR #2843, opened 2025-04-10
fix: Historical user conversation data has been cleared

shaohuzhang1:
fix: Historical user conversation data has been cleared

### DIFF
--- a/ui/src/components/ai-chat/component/chat-input-operate/index.vue
+++ b/ui/src/components/ai-chat/component/chat-input-operate/index.vue
@@ -271,7 +271,7 @@ const props = withDefaults(
     showUserInput?: boolean
     sendMessage: (question: string, other_params_data?: any, chat?: chatType) => void
     openChatId: () => Promise<string>
-    validate: () => Promise<boolean | string>
+    validate: () => Promise<any>
   }>(),
   {
     applicationDetails: () => ({}),
--- a/ui/src/stores/modules/application.ts
+++ b/ui/src/stores/modules/application.ts
@@ -97,8 +97,8 @@ const useApplicationStore = defineStore({
         applicationApi
           .postAppAuthentication(token, loading, authentication_value)
           .then((res) => {
-            localStorage.setItem(`${token}accessToken`, res.data)
-            sessionStorage.setItem(`${token}accessToken`, res.data)
+            localStorage.setItem(`${token}-accessToken`, res.data)
+            sessionStorage.setItem(`${token}-accessToken`, res.data)
             resolve(res)
           })
           .catch((error) => {
--- a/ui/src/stores/modules/user.ts
+++ b/ui/src/stores/modules/user.ts
@@ -61,11 +61,11 @@ const useUserStore = defineStore({
       return this.userType === 1 ? localStorage.getItem('token') : this.getAccessToken()
     },
     getAccessToken() {
-      const token = sessionStorage.getItem(`${this.userAccessToken}accessToken`)
+      const token = sessionStorage.getItem(`${this.userAccessToken}-accessToken`)
       if (token) {
         return token
       }
-      const local_token = localStorage.getItem(`${token}accessToken`)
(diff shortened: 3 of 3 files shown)


ui/src/stores/modules/application.ts
```diff
@@ -97,8 +97,8 @@ const useApplicationStore = defineStore({
         applicationApi
           .postAppAuthentication(token, loading, authentication_value)
           .then((res) => {
-            localStorage.setItem(`${token}accessToken`, res.data)
-            sessionStorage.setItem(`${token}accessToken`, res.data)
+            localStorage.setItem(`${token}-accessToken`, res.data)
+            sessionStorage.setItem(`${token}-accessToken`, res.data)
             resolve(res)
           })
           .catch((error) => {
```
shaohuzhang1, line 104: The provided code snippet has a couple of small optimizations you might consider:

1. **Consistent Item Name Format**: It seems that using different storage methods (localStorage and sessionStorage) with the same item name (`${token}accessToken`) can lead to unexpected behavior if keys overlap due to differences in case sensitivity cross-browsers.

   ```javascript
   // Recommended change: Use different item names for each store type
   localStorage.setItem(`${token}-accessToken`, res.data);
   sessionStorage.setItem(`${token.toUpperCase()}-accessToken`, res.data);  // Ensure consistency in casing
```

2. **Handling Error Cases More Explicitly**: The current error handling does not explicitly catch or log errors. This makes it harder to debug issues related to API responses or server-side problems.

3. **TypeScript Annotations**: If this is TypeScript code, adding appropriate types to arguments and return values could improve readability and maintainability.

```typescript
// Type annotations
const useApplicationStore = defineStore({
  state() {
    return {
      applicationApi: null as any,
      token: '' as string,
      loading: false as boolean,
    };
  },
  actions: {
    setToken(newToken: string): void {
      this.token = newToken;
    },
    postAppAuthentication(
      token: string,
      loading: boolean,
      authentication_value: string | null
    ): Promise<any> { // Add more specific typing as needed
      this.loading = true;  // You might want to start loading here too

      return this.applicationApi.postAppAuthentication(token, loading, authentication_value).then((res) => {
        localStorage.setItem(`${token.toLocaleUpperCase()}-accessToken`, res.data)  // Adjusted format for uniformity
        this.setToken(token) // Assuming you need to update your main context with the new token
        return res; // Return the response promise to allow chaining
      }).catch((error: any) => {  // Catch and handle errors appropriately
        console.error('Error during authentication:', error.message || error.toString());
        reject(error);
      });
    },
  },
});
```

By implementing these changes, the code becomes cleaner, more robust, and adheres better to best practices for both syntax and functionality.

ui/src/components/ai-chat/component/chat-input-operate/index.vue
```diff
@@ -271,7 +271,7 @@ const props = withDefaults(
     showUserInput?: boolean
     sendMessage: (question: string, other_params_data?: any, chat?: chatType) => void
     openChatId: () => Promise<string>
-    validate: () => Promise<boolean | string>
+    validate: () => Promise<any>
   }>(),
   {
     applicationDetails: () => ({}),
```
shaohuzhang1, line 277: The provided code snippet has one primary issue:

1. **Type Inconsistency in `validate` Function**: The original code specifies that `validate` returns `Promise<boolean | string>`, which means it can either resolve to a Boolean value or a String indicating an error.

2. **Updated Type for `validate`**:
   ```typescript
   validate: () => Promise<any>,
   ```
   This change allows `validate` to return any type of object or data, which is not appropriate according to the previous definition. You should ideally specify `string` as the return type if you plan on returning errors, or otherwise define a common type for both success outcomes and errors.

To maintain correctness, here is how you could update the function declaration:

```typescript
validate(): Promise<ResponseData>,


// Example response object structure:
interface ResponseData {
  success: boolean,
  errorMessage?: string // If used, optional
}

```

This update maintains consistency with the expected behavior specified in the initial `withDefaults` setup when calling `validate()`.

PR #2865, opened 2025-04-14
fix: handle line breaks in cell content for markdown table formatting

shaohuzhang1:
fix: handle line breaks in cell content for markdown table formatting  --bug=1054683 --user=刘瑞斌 【github#2831】知识库上传excel、应用编排文档内容提取节点中上传excel，单元格中有换行，导入后没有在一个单元格里显示 https://www.tapd.cn/57709429/s/1685274

### DIFF
--- a/apps/common/handle/impl/table/xls_parse_table_handle.py
+++ b/apps/common/handle/impl/table/xls_parse_table_handle.py
@@ -82,7 +82,10 @@ class XlsSplitHandle(BaseParseTableHandle):
                 for row in data:
                     # 将每个单元格中的内容替换换行符为 <br> 以保留原始格式
                     md_table += '| ' + ' | '.join(
-                        [str(cell).replace('\n', '<br>') if cell else '' for cell in row]) + ' |\n'
+                        [str(cell)
+                         .replace('\r\n', '<br>')
+                         .replace('\n', '<br>')
+                         if cell else '' for cell in row]) + ' |\n'
                 md_tables += md_table + '\n\n'
 
             return md_tables


apps/common/handle/impl/table/xls_parse_table_handle.py
```diff
@@ -82,7 +82,10 @@ def get_content(self, file, save_image):
...
                     # 将每个单元格中的内容替换换行符为 <br> 以保留原始格式
                     md_table += '| ' + ' | '.join(
-                        [str(cell).replace('\n', '<br>') if cell else '' for cell in row]) + ' |\n'
+                        [str(cell)
+                         .replace('\r\n', '<br>')
+                         .replace('\n', '<br>')
+                         if cell else '' for cell in row]) + ' |\n'
                 md_tables += md_table + '\n\n'
 
             return md_tables
```
shaohuzhang1, line 91: The provided code looks mostly correct and efficient for its intended purpose of transforming table contents to Markdown format while preserving newline characters. However, I'd make a few minor improvements:

1. **Consistent Replacement**: You can consolidate the three replacements (`\r\n`, `\n`) into one line where applicable. This reduces redundancy.

2. **Readability**: It's good practice to have comments explaining certain parts, though they might not be strictly necessary given the code is relatively straightforward. If needed, you could add more detailed explanations in inline comments.

3. **Functionality Check**: Ensure that the function handles various edge cases correctly, such as empty cells or special characters within cells.

Here's the revised code with these suggestions applied:

```python
    md_tables = ''
    
    # Process each row in the data
    for row in data:
        # Replace \r\n at both ends before joining elements, then replace any remaining newlines with <br>
        md_row = '| ' + ' | '.join(cell.replace('\r\n', '').replace('\n', '<br>') if cell else '' for cell in row) + ' |\n'
        
        md_tables += md_row
    
    return markdown_tables  # Assuming markdown_tables was misspelled here and should be md_tables
```

These changes ensure consistency and readability, making it easier to maintain and understand future modifications to this function.

PR #2871, opened 2025-04-14
refactor: enhance input handling for custom sources in form fields

shaohuzhang1:
refactor: enhance input handling for custom sources in form fields  --bug=1054775 --user=刘瑞斌 【应用编排】MCP调用节点，自定义工具参数建议支持选择数据类型 https://www.tapd.cn/57709429/s/1685483

### DIFF
--- a/ui/src/workflow/nodes/mcp-node/index.vue
+++ b/ui/src/workflow/nodes/mcp-node/index.vue
@@ -97,11 +97,24 @@
             </div>
           </template>
           <el-input
-            v-if="item.source === 'custom'"
+            v-if="item.source === 'custom' && item.input_type === 'TextInput'"
             v-model="form_data.tool_params[form_data.params_nested][item.label.label]"
           />
+          <el-input-number
+            v-else-if="item.source === 'custom' && item.input_type === 'NumberInput'"
+            v-model="form_data.tool_params[form_data.params_nested][item.label.label]"
+          />
+          <el-switch
+            v-else-if="item.source === 'custom' && item.input_type === 'SwitchInput'"
+            v-model="form_data.tool_params[form_data.params_nested][item.label.label]"
+          />
+          <el-input
+            v-else-if="item.source === 'custom' && item.input_type === 'JsonInput'"
+            v-model="form_data.tool_params[form_data.params_nested][item.label.label]"
+            type="textarea"
+          />
           <NodeCascader
-            v-else
+            v-if="item.source === 'referencing'"
             ref="nodeCascaderRef2"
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/mcp-node/index.vue
```diff
@@ -276,7 +328,7 @@ function changeTool() {
           attrs: { tooltip: args_schema.properties[item].description },
           props_info: {}
         },
-        input_type: 'TextInput',
+        input_type: input_type,
         source: 'referencing',
         required: args_schema.required?.indexOf(item) !== -1,
         props_info: {
```
shaohuzhang1, line 334: There are several improvements and optimizations that can be made to this code:

1. **Type Checking in `changeTool` Function**:
   Instead of using type checking based on the string value (`type: 'string'`, `type: 'number'`, etc.), you should directly access the `typeof` property of the schema's properties. This is more robust as it checks against JavaScript types rather than hardcoded strings.

   ```javascript
   if (schema.items && Array.isArray(schema.items)) {
     // ...
   } else if (schema.type === 'boolean') {
     // ...
   }
   ```

2. **Simplify Nested Conditional Logic**:
   The nested conditional blocks for setting input types can be simplified by using an associative array to map data types to their corresponding input components.

   ```javascript
   let inputComponentMap = new Map([
     ['string', 'el-input'],
     ['number', 'el-input-number'],
     ['boolean', 'el-switch'],
     ['object', 'el-input'], // Assuming object is treated as JSON input due to its complexity
     ['array', 'el-input']  // Assuming array is treated as JSON input due to its variability
   ]);

   const input_component = inputComponentMap.get(args_schema.properties[item].type);
   if (input_component) {
     form_data.value.tool_form_field.push({
       field: item,
       label: {
         text: args_schema.properties[item].title || item,
         props_info: {}
       },
       input_component: input_component,
       source: 'referencing',
       required: args_schema.required?.indexOf(item) !== -1
     });
```

3. **Use Proper Namespaced Imports**:
   If `nodeCascaderRef2` is part of a larger Vue component with namespacing, ensure that you use the correct import statements. For example, if it comes from a module named `cascader-module`, you would use something like:

   ```typescript
   import { cascaderModule } from './cascader-module';
   ```

4. **Improve Error Handling**:
   Add error handling around API calls and other operations to prevent unexpected behavior when encountering invalid responses or errors.

5. **Documentation Comments**:
   Ensure there are clear documentation comments explaining each section of the code, especially where variables and methods are used.

By implementing these changes, the code becomes cleaner, more efficient, and easier to maintain.

PR #2885, opened 2025-04-15
feat: load environment variables from .env file and update config loading logic

shaohuzhang1:
chore: add dependabot configuration for pip updates<br>feat: load environment variables from .env file and update config loading logic

### DIFF
--- a/apps/maxkb/const.py
+++ b/apps/maxkb/const.py
@@ -2,6 +2,8 @@
 #
 import os
 
+from dotenv import load_dotenv
+
 from .conf import ConfigManager
 
 __all__ = ['BASE_DIR', 'PROJECT_DIR', 'VERSION', 'CONFIG']
@@ -9,4 +11,12 @@ __all__ = ['BASE_DIR', 'PROJECT_DIR', 'VERSION', 'CONFIG']
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 PROJECT_DIR = os.path.dirname(BASE_DIR)
 VERSION = '2.0.0'
-CONFIG = ConfigManager.load_user_config(root_path=os.path.abspath('/opt/maxkb/conf'))
+
+# load environment variables from .env file
+load_dotenv()
+# print(os.getenv('MAXKB_CONFIG'))
+if os.getenv('MAXKB_CONFIG') is not None:
+    CONFIG = ConfigManager.load_user_config(root_path=PROJECT_DIR)
+else:
+    CONFIG = ConfigManager.load_user_config(root_path=os.path.abspath('/opt/maxkb/conf'))
+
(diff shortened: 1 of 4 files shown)


apps/maxkb/const.py
```diff
@@ -2,11 +2,21 @@
...
-CONFIG = ConfigManager.load_user_config(root_path=os.path.abspath('/opt/maxkb/conf'))
+
+# load environment variables from .env file
+load_dotenv()
+# print(os.getenv('MAXKB_CONFIG'))
+if os.getenv('MAXKB_CONFIG') is not None:
+    CONFIG = ConfigManager.load_user_config(root_path=PROJECT_DIR)
+else:
+    CONFIG = ConfigManager.load_user_config(root_path=os.path.abspath('/opt/maxkb/conf'))
+
```
shaohuzhang1, line 22: The code looks generally clean, but there are a few things you can optimize:

1. **Environment Variable Loading**: The `dotenv` library is used to load environment variables from a `.env` file. If `.env` is located where you might intend it, this step is necessary. However, be aware that if the `.env` file is not found or contains invalid data, your application may fail.

2. **Path Handling**: Ensure that the path to your configuration file (`/opt/maxkb/conf`) is correct and accessible. You may want to check if an environment variable specifies a different location, such as `MAXKB_CONFIG_PATH`.

3. **Empty Environment Variable Checks**: Using `os.getenv()` is more robust than directly checking for truthiness (i.e., using `!= ''`). This prevents errors when reading from non-existent files or directories.

Here’s an optimized version of the code with comments explaining these changes:

```python
@@ -2,11 +2,27 @@
 #
 import os

-from .conf import ConfigManager

 __all__ = ['BASE_DIR', 'PROJECT_DIR', 'VERSION', 'CONFIG']

 # Determine base directory of the project
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))

 # Location of the project root
 PROJECT_DIR = os.path.dirname(BASE_DIR)

-# Version of the software
 VERSION = '2.0.0'

+# Load user-specific configuration if MAXKB_CONFIG env var is set
+try:
+    CONFIG_FILE = os.environ.get('MAXKB_CONFIG')
+except KeyError:
+    CONFIG_FILE = '/opt/maxkb/conf'  # Default fallback to default config location
+
+# Load environment variables from .env file if present
+import dotenv
+dotenv.load_dotenv()
+
+# Attempt to load specific configuration path based on env var
+if CONFIG_FILE:
+    CONFIG = ConfigManager.load_user_config(root_path=CONFIG_FILE)
+else:
+    print("Warning: MAXKB_CONFIG environment variable not set", file=sys.stderr)
+    CONFIG = ConfigManager.load_user_config(root_path='/opt/maxkb/conf')
```

### Key Changes:
- **Default Path Check**: Added a `try-except` block to handle situations where `MAXKB_CONFIG` is not set, providing a warning message and setting a default path instead.
- **DotEnv Configuration**: Enhanced handling of loading environment variables by explicitly specifying `dotenv`, making it clear what module is being imported.

PR #2924, opened 2025-04-18
fix: improve module tree query to use Q object for filtering by workspace_id

shaohuzhang1:
fix: improve module tree query to use Q object for filtering by workspace_id

### DIFF
--- a/apps/modules/api/module.py
+++ b/apps/modules/api/module.py
@@ -105,5 +105,12 @@ class ModuleTreeReadAPI(APIMixin):
                 enum=["APPLICATION", "KNOWLEDGE", "TOOL"],
                 location='path',
                 required=True,
-            )
+            ),
+            OpenApiParameter(
+                name="name",
+                description="名称",
+                type=OpenApiTypes.STR,
+                location='query',
+                required=False,
+            ),
         ]
--- a/apps/modules/serializers/module.py
+++ b/apps/modules/serializers/module.py
@@ -95,9 +95,13 @@ class ModuleTreeSerializer(serializers.Serializer):
     workspace_id = serializers.CharField(required=True, allow_null=True, allow_blank=True, label=_('workspace id'))
     source = serializers.CharField(required=True, label=_('source'))
 
-    def get_module_tree(self):
+    def get_module_tree(self, name=None):
         self.is_valid(raise_exception=True)
         Module = get_module_type(self.data.get('source'))
-        nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id'))).get_cached_trees()
+        if name is not None:
+            nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id')) &
+                                          Q(name__contains=name)).get_cached_trees()
+        else:
+            nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id'))).get_cached_trees()
         serializer = ToolModuleTreeSerializer(nodes, many=True)
         return serializer.data  # 这是可序列化的字典
--- a/apps/modules/views/module.py
+++ b/apps/modules/views/module.py
@@ -88,4 +88,4 @@ class ModuleTreeView(APIView):
     def get(self, request: Request, workspace_id: str, source: str):
         return result.success(ModuleTreeSerializer(
             data={'workspace_id': workspace_id, 'source': source}
-        ).get_module_tree())
+        ).get_module_tree(request.query_params.get('name')))


apps/modules/api/module.py
```diff
@@ -105,5 +105,12 @@ def get_parameters():
...
-            )
+            ),
+            OpenApiParameter(
+                name="name",
+                description="名称",
+                type=OpenApiTypes.STR,
+                location='query',
+                required=False,
+            ),
         ]
```
shaohuzhang1, line 116: ### Code Review and Optimization Suggestions

The provided code snippet defines a method `get_parameters` which appears to use an external library or framework (OpenApi), but the specifics of how it's used are not clear from the context. Assuming this involves defining API parameters using OpenAPI specifications:

**Potential Irregularities:**
1. **Duplicate Parameter Definition:** The second parameter definition seems incorrect because both "description" and "type" fields have been set unnecessarily multiple times.

   ```python
    enum=["APPLICATION", "KNOWLEDGE", "TOOL"]
   ```

2. **Incorrect Enum Type Handling:** It looks like "enum" should be defined with appropriate values rather than just its keys ("application, knowledge, tool").

3. **Missing Description for Query Parameter:** While "name" is described in Chinese, no description is given for the "location". This could enhance documentation clarity.

**Issues to Address:**

1. **Correct Enumeration:** Replace `"enum"` with list form suitable for enumeration if needed:
   
   ```python
    enum=["APPLICATION", "KNOWLEDGE", "TOOL"]
   ```
      
2. **Simplify Parameter Definitions:** Since there’s only one optional query parameter here, you might consider removing redundancy and streamline them into a single dictionary entry, leveraging nested dictionaries when necessary for more complex structures.

   ```python
    [
        {
            "name": "parameter1",
            "description": "First parameter description.",
            "required": True,
            # Add type and other properties as needed
        },
        {
            "name": "optional_param",
            "description": "Second optional parameter.",
            "type": openapi_types.STR,
            'location': 'query',
            "required": False,
        }
    ]
   ```

3. **Documentation Clarification:** Ensure each parameter has a complete and concise description for better understanding by non-native speakers. Also, clarify what each element means and where these can be found if applicable.

These adjustments will improve both readability of the code and maintainability by eliminating redundant data duplication and clarifying expectations about the types and meanings of each variable passed through the API request.

apps/modules/serializers/module.py
```diff
@@ -95,9 +95,13 @@ class ModuleTreeSerializer(serializers.Serializer):
...
         self.is_valid(raise_exception=True)
         Module = get_module_type(self.data.get('source'))
-        nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id'))).get_cached_trees()
+        if name is not None:
+            nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id')) &
+                                          Q(name__contains=name)).get_cached_trees()
+        else:
+            nodes = Module.objects.filter(Q(workspace_id=self.data.get('workspace_id'))).get_cached_trees()
         serializer = ToolModuleTreeSerializer(nodes, many=True)
         return serializer.data  # 这是可序列化的字典
```
shaohuzhang1, line 107: The code provided has a few issues and optimizations that could be addressed:

1. **Potential Performance Issue**: The usage of `get_cached_trees()` can lead to performance issues if there's no specific caching backend or query parameters that ensure efficient retrieval. Consider using `filter()`, especially with more complex conditions like `name__contains`, to avoid unnecessary overhead.

2. **Code Duplication**: There is redundancy in handling the `nodes` queryset based on whether `name` is provided. This can be simplified by reducing the duplication.

Here’s an optimized version of the function:

```python
from django.db.models import Q

class ModuleTreeSerializer(serializers.Serializer):
    workspace_id = serializers.CharField(required=True, allow_null=True, allow_blank=True, label=_('workspace id'))
    source = serializers.CharField(required=True, label=_('source'))

    def get_module_tree(self, name=None):
        
        module_queryset = get_module_type(self.data['source']).objects.filter(
                Q(workspace_id=self.data['workspace_id'])
            )
        
        if name is not None:
            module_queryset = module_queryset.filter(name__icontains=name)
        
        nodes = module_queryset.get_cached_trees()
        
        serializer = ToolModuleTreeSerializer(nodes, many=True)
        return serializer.data  # 这是可序列化后的列表或字典
```

### Key Changes:
- Removed the redundant condition for checking if `.get_cached_trees()` should include `id`. If it already includes `id`, this check is unnecessary.
- Combined the filter logic into one line within the main `module_queryset`.
- Simplified the `if name:` block to just use chaining methods directly on `module_queryset`.

These changes improve readability and reduce code redundancy while maintaining functionality.

apps/modules/views/module.py
```diff
@@ -88,4 +88,4 @@ class ModuleTreeView(APIView):
     def get(self, request: Request, workspace_id: str, source: str):
         return result.success(ModuleTreeSerializer(
             data={'workspace_id': workspace_id, 'source': source}
-        ).get_module_tree())
+        ).get_module_tree(request.query_params.get('name')))
```
shaohuzhang1, line 91: The code snippet you provided has two main areas that can be checked:

1. **Potential Issues**:
   - The `get_module_tree` method within `ModuleTreeSerializer` is called without using the `request.query_params.get('name')`, which could lead to incorrect operation or missing parameters when retrieving module tree data from the database.

   - Since you mentioned there's no specific need to use any optimizations right now (as it might depend on additional context), my advice would typically focus on ensuring the integrity and reliability of the logic. However, one small improvement could be to ensure that the default behavior of `get_queryset()` in the serializer does not conflict with explicitly passing a parameter like `'name'`.

Overall, make sure to review if there are any other parts of your application where this method is used, especially considering that it relies on the presence of a query parameter named `'name'`. If not every invocation of `get_module_tree` must include this parameter, consider refactoring your code or documentation to clarify usage expectations.

PR #2937, opened 2025-04-21
fix: Model parameters are not effective

shaohuzhang1:
fix: Model parameters are not effective

### DIFF
--- a/apps/setting/models_provider/impl/openai_model_provider/model/llm.py
+++ b/apps/setting/models_provider/impl/openai_model_provider/model/llm.py
@@ -9,7 +9,6 @@
 from typing import List, Dict
 
 from langchain_core.messages import BaseMessage, get_buffer_string
-from langchain_openai.chat_models import ChatOpenAI
 
 from common.config.tokenizer_manage_config import TokenizerManage
 from setting.models_provider.base_model_provider import MaxKBBaseModel
@@ -35,9 +34,9 @@ class OpenAIChatModel(MaxKBBaseModel, BaseChatOpenAI):
             streaming = False
         azure_chat_open_ai = OpenAIChatModel(
             model=model_name,
-            openai_api_base=model_credential.get('api_base'),
-            openai_api_key=model_credential.get('api_key'),
-            **optional_params,
+            base_url=model_credential.get('api_base'),
+            api_key=model_credential.get('api_key'),
+            extra_body=optional_params,
             streaming=streaming,
             custom_get_token_ids=custom_get_token_ids
         )
--- a/apps/setting/models_provider/impl/tencent_cloud_model_provider/model/llm.py
+++ b/apps/setting/models_provider/impl/tencent_cloud_model_provider/model/llm.py
@@ -33,21 +33,7 @@ class TencentCloudChatModel(MaxKBBaseModel, BaseChatOpenAI):
             model=model_name,
             openai_api_base=model_credential.get('api_base'),
             openai_api_key=model_credential.get('api_key'),
-            **optional_params,
+            extra_body=optional_params,
             custom_get_token_ids=custom_get_token_ids
         )
         return azure_chat_open_ai
-
-    def get_num_tokens_from_messages(self, messages: List[BaseMessage]) -> int:
-        try:
-            return super().get_num_tokens_from_messages(messages)
-        except Exception as e:
-            tokenizer = TokenizerManage.get_tokenizer()
-            return sum([len(tokenizer.encode(get_buffer_string([m]))) for m in messages])
-
-    def get_num_tokens(self, text: str) -> int:
-        try:
-            return super().get_num_tokens(text)
-        except Exception as e:
-            tokenizer = TokenizerManage.get_tokenizer()
-            return len(tokenizer.encode(text))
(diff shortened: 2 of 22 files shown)


apps/setting/models_provider/impl/tencent_cloud_model_provider/model/llm.py
```diff
@@ -33,21 +33,7 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
...
-        except Exception as e:
-            tokenizer = TokenizerManage.get_tokenizer()
-            return sum([len(tokenizer.encode(get_buffer_string([m]))) for m in messages])
-
-    def get_num_tokens(self, text: str) -> int:
-        try:
-            return super().get_num_tokens(text)
-        except Exception as e:
-            tokenizer = TokenizerManage.get_tokenizer()
-            return len(tokenizer.encode(text))
```
shaohuzhang1, line 53: There are some issues in the code that need to be addressed:

1. **Duplicate Code**: The `get_num_tokens` and `get_num_tokens_from_messages` methods should ideally not duplicate each other, as they calculate token counts using similar logic but with different approaches (using encoder on individual messages vs. entire text).

2. **Exception Handling in Base Class Calls**: The `super()` calls inside `get_num_tokens` and `get_num_tokens_from_messages` do not handle exceptions properly. It's better to encapsulate this behavior if it applies universally across all subclasses.

3. **Tokenizer Management Class**: If the `TokenizerManage.get_tokenizer()` method is used extensively throughout this module, consider separating its implementation into a separate class file. This improves maintainability and reusability.

Here's an optimized version of the code, incorporating these improvements:

```python
from typing import TypeVar, Dict, Any
from langchain.llms.base import LLM
from langchain.schema.messages import BaseMessage

T = TypeVar('T')

class CustomLLM(LLM):
    def __init__(self,
                 model_type: str,
                 model_name: str,
                 model_credential: Dict[str, object],
                 **optional_params: Optional[Any]):
        super().__init__(
            model=model_name,
            extra_body=optional_params,
            custom_get_token_ids=lambda _: None  # Placeholder; replace with actual implementation
        )

    @property
    def _llm_type(self) -> str:
        return "Custom LLM"

    def generate_prompt(self, prompt_message_list: list, **kwargs):  # Placeholder; replace with actual implementation
        pass

    def num_tokens_method(
        self,
        input_texts: Union[List[str], List[list]],
    ) -> List[int]:
        tokenizer_manage = TokenizerManage()  # Ensure this instance creation is optimal here
        total_len = []
        
        for texts in input_texts:
            if isinstance(texts, list):
                tokens = sum([tokenizer_manage.tokens_encode(text) for text in texts])
            else:
                tokens = tokenizer_manage.tokens_encode(texts)
            
            total_len.append(tokens)

        return total_len

# Assuming TokenizerManage has been defined elsewhere with necessary functions
```

### Key changes made:
- Moved exception handling from within method calls up to where the `TokenizerMananger` was initialized in both `_num_tokens` methods.
- Separated the main logic of counting tokens from the error-handling, making code cleaner and more modular.
- Added a property to define the type of LLM, which can be useful in subclassing scenarios.
- Provided placeholders for the `generate_prompt` and `_num_tokens` methods based on expected usage patterns, assuming these would be implemented further in subclasses or external modules.

apps/setting/models_provider/impl/openai_model_provider/model/llm.py
```diff
@@ -35,9 +34,9 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
...
             model=model_name,
-            openai_api_base=model_credential.get('api_base'),
-            openai_api_key=model_credential.get('api_key'),
-            **optional_params,
+            base_url=model_credential.get('api_base'),
+            api_key=model_credential.get('api_key'),
+            extra_body=optional_params,
             streaming=streaming,
             custom_get_token_ids=custom_get_token_ids
         )
```
shaohuzhang1, line 42: There are no apparent issues with the existing code, so here is a summary of potential improvements:
1. The `azure_chat_open_ai` object can be assigned to the variable name that best represents its purpose in your context (e.g., chat_model).
2. If possible, add additional exception handling to manage errors better during API requests.
3. Consider using more descriptive parameter names than `optional_params`, such as `extra_settings`.
4. It's recommended to use type hints consistently across the file and update them whenever you refactor the function parameters or return types.

Here is how the refactored code could be structured:
```python
@@ -35,8 +34,9 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
     if stream:
         streaming = False

-    # Using 'chat_model' instead of 'azure_chat_open_ai'
+    chat_model = OpenAIChatModel(
     model=model_name,
-    openai_api_base=model_credential.get('api_base'), 
-    openai_api_key=model_credential.get('api_key'),
+    base_url=model_credential.get('api_base'),  
+    api_key=model_credential.get('api_key'),          
# Update this parameter name depending on what it represents in your app    
       extra_body=optional_settings_dict,

        streaming=streaming,     
        custom_get_token_ids=custom_get_token_ids        
   ) 

```
This change provides clarity about the object being created and enhances readability throughout the codebase. Remember always to test these changes thoroughly after making modifications!

PR #2943, opened 2025-04-22
fix: OpenAI interface call for session, passing asker does not take effect

shaohuzhang1:
fix: OpenAI interface call for session, passing asker does not take effect

### DIFF
--- a/apps/application/serializers/chat_message_serializers.py
+++ b/apps/application/serializers/chat_message_serializers.py
@@ -213,12 +213,21 @@ class OpenAIChatSerializer(serializers.Serializer):
         return instance.get('messages')[-1].get('content')
 
     @staticmethod
-    def generate_chat(chat_id, application_id, message, client_id):
+    def generate_chat(chat_id, application_id, message, client_id, asker=None):
         if chat_id is None:
             chat_id = str(uuid.uuid1())
         chat = QuerySet(Chat).filter(id=chat_id).first()
         if chat is None:
-            Chat(id=chat_id, application_id=application_id, abstract=message[0:1024], client_id=client_id).save()
+            asker_dict = {'user_name': '游客'}
+            if asker is not None:
+                if isinstance(asker, str):
+                    asker_dict = {
+                        'user_name': asker
+                    }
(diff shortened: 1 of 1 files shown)


apps/application/serializers/chat_message_serializers.py
```diff
@@ -232,7 +241,8 @@ def chat(self, instance: Dict, with_valid=True):
         application_id = self.data.get('application_id')
         client_id = self.data.get('client_id')
         client_type = self.data.get('client_type')
-        chat_id = self.generate_chat(chat_id, application_id, message, client_id)
+        chat_id = self.generate_chat(chat_id, application_id, message, client_id,
+                                     asker=instance.get('form_data', {}).get("asker"))
         return ChatMessageSerializer(
             data={
                 'chat_id': chat_id, 'message': message,
```
shaohuzhang1, line 248: 1. The `asker` parameter was added to the `generate_chat` method, so you need to handle it accordingly in all parts of the code where this function is called.

2. In the constructor (`__init__`) of the class, `self.client_id` should be set based on `self.data['client_id']` or `None`.

3. It's unclear whether `ChatMessageSerializer` requires additional initialization parameters beyond what it takes from the serializer itself (e.g., `data`). If so, consider adding them here.

4. Ensure that the input dictionary `self.data` does not contain null values for required keys like `chat_id`, `application_id`, and `message`. If they aren't optional, handle their absence gracefully.

5. Consider using type hints throughout the code to improve readability and maintainability. For example, change the line defining `self.data: Dict[str, Any] = {}` into `self.data: Optional[Dict[str, Union[str, int]]] = {}` for better understanding about its expected content.

6. When handling user input via APIs, remember to validate it before processing to prevent attacks such as SQL injection, cross-site scripting (XSS), and other security vulnerabilities.

7. Review how `QuerySet(Chat)` interacts with other parts of your API to ensure proper database queries are executed when retrieving `Chat` objects.

8. Since this code handles messages and potentially sensitive information, implement appropriate logging around critical operations within try-except blocks to capture errors cleanly for debugging purposes.

PR #2945, opened 2025-04-22
fix: Edit advanced orchestration applications, add application nodes, and display that some applications are unavailable after being added

shaohuzhang1:
fix: Edit advanced orchestration applications, add application nodes, and display that some applications are unavailable after being added

### DIFF
--- a/ui/src/workflow/nodes/application-node/index.vue
+++ b/ui/src/workflow/nodes/application-node/index.vue
@@ -238,41 +238,49 @@ const update_field = () => {
         const new_user_input_field_list = cloneDeep(
           ok.data.work_flow.nodes[0].properties.user_input_field_list
         )
-        const merge_api_input_field_list = new_api_input_field_list.map((item: any) => {
-          const find_field = old_api_input_field_list.find(
-            (old_item: any) => old_item.variable == item.variable
-          )
-          if (find_field) {
-            return {
-              ...item,
-              value: find_field.value,
-              label:
-                typeof item.label === 'object' && item.label != null ? item.label.label : item.label
+        const merge_api_input_field_list =
+          new_api_input_field_list ||
+          [].map((item: any) => {
+            const find_field = old_api_input_field_list.find(
+              (old_item: any) => old_item.variable == item.variable
+            )
+            if (find_field) {
+              return {
+                ...item,
+                value: find_field.value,
+                label:
+                  typeof item.label === 'object' && item.label != null
+                    ? item.label.label
+                    : item.label
+              }
+            } else {
+              return item
             }
-          } else {
-            return item
-          }
-        })
+          })
         set(
           props.nodeModel.properties.node_data,
           'api_input_field_list',
           merge_api_input_field_list
         )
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/application-node/index.vue
```diff
@@ -294,6 +302,7 @@ const update_field = () => {
       }
     })
     .catch((err) => {
+      console.log(err)
       set(props.nodeModel.properties, 'status', 500)
     })
 }
```
shaohuzhang1, line 308: The provided code appears to be a functional component for updating field properties within a workflow node data structure. The key operations involve merging `new_api_input_field_list` and `old_api_input_field_list`, handling missing fields, and similarly for `user_input_field_list`. Here are some review points to consider:

### Regularity Checks
1. **Function Definition**: The function is defined but lacks documentation about what it does or its purpose.
2. **Error Handling**: There's no logging outside of catching an error (`err`) which means the exact nature of errors might not be fully captured.

### Potential Issues
1. **Empty Lists**: If either `new_api_input_field_list` or `old_api_input_field_list` is empty, attempting to map over them without checking for emptiness can lead to undefined behavior.
   
2. **Label Assignment**: Ensure that the assignment logic for labels handles both object values and primitive types correctly.

3. **Status Update**: While using `set(props.nodeModel.properties, 'status', 500)` when there's an error is useful, it would generally be more informative to include specific error messages in logs or UI updates.

### Optimization Suggestions
1. **Avoid Nullish Coalescing Twice**:
   ```javascript
   const merge_user_input_field_list = new_user_input_field_list ||
     [].map((item: any) => {
       // Existing code
     });
   ```
   You can optimize this by removing one level of nesting.
   
2. **Use Immutability Libraries**:
   Consider using libraries like Ramda for utility functions that help with immutability better than plain JavaScript methods.

    Example usage with Ramda:
    ```javascript
    import { mergeWithKeyRight } from 'ramda';

    const merge_with_right = (key, val1, val2) =>
      typeof val1 === 'string' ? val1 : typeof val2 === 'string' ? val2 : null;

    const merge_api_input_field_list = old_api_input_field_list
      .reduce((acc, oldItem) => ({
        ...acc,
        [oldItem.variable]: newItem || oldItem
      }), {});

    const merge_user_input_field_list = old_user_input_field_list
      .reduce((acc, oldItem) => ({
        ...acc,
        [oldItem.field]: newItem || oldItem
      }), {});
      
    set(
      'api_input_field_list',
      Object.values(merge_api_input_field_list).map(item => ({...item, value: item.value}))
    );
    
    set(
      'user_input_field_list',
      Object.values(merge_user_intput_field_list).map(item => ({...item, value: item.value}))
    );
    ```

3. **Code Readability**: Break down complex conditional checks into separate helper functions or inline conditions where necessary.

4. **Consistent Logic Across Fields**:
   Ensure consistent naming and approach for finding fields across different input lists to avoid redundancy and complexity.

### Conclusion

The provided code is mostly functional, with minor improvements suggested for code readability and efficiency. Addressing these points will enhance maintainability and performance.

PR #2947, opened 2025-04-22
fix: When there is an empty string in the context, the conversation reports an error

shaohuzhang1:
fix: When there is an empty string in the context, the conversation reports an error

### DIFF
--- a/apps/application/models/application.py
+++ b/apps/application/models/application.py
@@ -11,7 +11,7 @@ import uuid
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 from langchain.schema import HumanMessage, AIMessage
-
+from django.utils.translation import gettext as _
 from common.encoder.encoder import SystemEncoder
 from common.mixins.app_model_mixin import AppModelMixin
 from dataset.models.data_set import DataSet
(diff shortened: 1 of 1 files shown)


apps/application/models/application.py
```diff
@@ -167,7 +167,11 @@ def get_human_message(self):
...
     def get_ai_message(self):
-        return AIMessage(content=self.answer_text)
+        answer_text = self.answer_text
+        if answer_text is None or len(str(answer_text).strip()) == 0:
+            answer_text = _(
+                'Sorry, no relevant content was found. Please re-describe your problem or provide more information. ')
+        return AIMessage(content=answer_text)
 
     def get_node_details_runtime_node_id(self, runtime_node_id):
         return self.details.get(runtime_node_id, None)
```
shaohuzhang1, line 177: The code appears to be well-written, but there are a couple of optimizations you might consider:

1. **String Formatting:** In `get_ai_message`, using an inline string concatenation directly within `_.format()` can be less efficient than constructing the entire string first. Consider extracting the result into a variable.

2. **Variable Naming Consistency:** Ensure that all variables (like `answer_text`) use consistent naming conventions throughout the method.

Here's revised version with some of these suggestions implemented:

```python
from django.contrib.postgres.fields import ArrayField
from django.db import models
from langchain.schema import HumanMessage, AIMessage

from common.encoder.encoder import SystemEncoder
from common.mixins.app_model_mixin import AppModelMixin
from dataset.models.data_set import DataSet


def construct_answer_message(text):
    if not text or str(text).strip() == "":
        return _(
            'Sorry, no relevant content was found. Please re-describe your problem or provide more information.'
        )
    else:
        return text

class YourClassName(AppModelMixin):  # Replace with actual class name
    problem_text = models.TextField()
    answer_text = models.TextField()

    def get_human_message(self):
        return HumanMessage(content=self.problem_text)

    def get_ai_message(self):
        return AIMessage(content=construct_answer_message(self.answer_text))

    def get_node_details_runtime_node_id(self, runtime_node_id):
        return self.details.get(runtime_node_id, None)
```

This change refactors the construction of the AI message into a separate function (`construct_answer_message`), making it easier to test and reuse if needed. The rest of the logic remains mostly unchanged.

PR #2948, opened 2025-04-22
fix: The forgot password page displays an error

shaohuzhang1:
fix: The forgot password page displays an error  --bug=1055073 --user=王孝刚 【长城开发科技】自定平台登录设置在点击忘记密码时还是默认提示 https://www.tapd.cn/57709429/s/1689108

### DIFF
--- a/ui/src/views/login/forgot-password/index.vue
+++ b/ui/src/views/login/forgot-password/index.vue
@@ -1,6 +1,10 @@
 <template>
   <login-layout>
-    <LoginContainer :subTitle="$t('views.system.theme.defaultSlogan')">
+    <LoginContainer
+      :subTitle="
+        user.themeInfo?.slogan ? user.themeInfo?.slogan : $t('views.system.theme.defaultSlogan')
+      "
+    >
       <h2 class="mb-24">{{ $t('views.login.forgotPassword') }}</h2>
       <el-form
         class="register-form"
@@ -41,15 +45,15 @@
                   isDisabled
                     ? `${$t('views.login.verificationCode.resend')}（${time}s）`
                     : $t('views.login.verificationCode.getVerificationCode')
-                }}</el-button
-              >
+                }}
+              </el-button>
             </div>
           </el-form-item>
         </div>
       </el-form>
-      <el-button size="large" type="primary" class="w-full" @click="checkCode">{{
-        $t('views.login.buttons.checkCode')
-      }}</el-button>
+      <el-button size="large" type="primary" class="w-full" @click="checkCode"
+        >{{ $t('views.login.buttons.checkCode') }}
+      </el-button>
       <div class="operate-container mt-12">
         <el-button
           class="register"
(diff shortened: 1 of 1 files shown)


ui/src/views/login/forgot-password/index.vue
```diff
@@ -72,7 +76,10 @@ import type { FormInstance, FormRules } from 'element-plus'
 import UserApi from '@/api/user'
 import { MsgSuccess } from '@/utils/message'
 import { t } from '@/locales'
+import useStore from '@/stores'
+
 const router = useRouter()
+const { user } = useStore()
 const CheckEmailForm = ref<CheckCodeRequest>({
   email: '',
   code: '',
```
shaohuzhang1, line 85: The provided code has a few areas that can be improved:

1. **Template Optimization**: The template contains two consecutive `<br>` tags at lines 3 and 48. Instead of using `</br>`, which is not standard HTML syntax, you should use line breaks for readability.

2. **Component Props Handling**: In the `<LoginContainer>` component, the sub-title is set based on `user.themeInfo.slogan` if it exists, otherwise it defaults to `$t('views.system.theme.defaultSlogan')`. This handling ensures that no error occurs if `themeInfo` or its properties do not exist.

3. **Button Disabled State**: It appears there might be an issue with how the button's disabled state is updated, especially related to the timing (`time`) variable. Ensure this logic updates correctly to prevent unexpected behavior when `resendTime` expires.

4. **Dynamic Content Handling**: When displaying dynamic content such as verification messages or buttons, consider using proper Vue.js directives like `v-if` and other controls to handle their visibility dynamically without resorting to conditional statements inside strings.

5. **Globalization Considerations**: If internationalization (i18n) is handled through external libraries like i18next or vue-i18n, ensure consistency across all usage of translations, including static texts within elements' text nodes like headings and button captions. Using consistent data structure patterns for storing translated information can help maintain clarity and reduce errors during localization efforts.

Here is a slightly reformatted version with these considerations incorporated:

```html
<template>
  <login-layout>
    <LoginContainer :subTitle="getSubtitle">
      <h2 class="mb-24"> {{ t('views.login.forgotPassword') }} </h2>

      <el-form class="register-form">
        ...

        <button-size large type="primary" class="w-full" @click="checkCode">
          {{ $t('views.login.buttons.checkCode') }}
        </button-size>

        ...
        
      </el-form>
  
      <!-- Other components -->
      
    </LoginContainer>
  </login-layout>
</template>

<script lang="ts">

// Assuming getSubtitle() function calculates the subtitle string

setup () {
  // Your setup code here...

  return {
    t,
    checkCode,
    sendAgain,
    resendTimer: undefined,
    formRef,

    calculatedSubtitle: computed(() => (
      user.themeInfo && user.themeInfo.slogan !== null &&
      typeof user.themeInfo.slogan === 'string'
        ? user.themeInfo.slogan
        : t('views.system.theme.defaultSlogan')
    ))
  }
}

</script>

<style scoped>
/* Your styles here */
</style>
```

This refactoring helps maintain better organization and avoids several common pitfalls in Vue.js development practices. Make sure to adjust `getSubtitle()` according to your actual implementation if needed.

PR #2949, opened 2025-04-22
fix: Docx segmented font title recognition

shaohuzhang1:
fix: Docx segmented font title recognition

### DIFF
--- a/apps/common/handle/impl/doc_split_handle.py
+++ b/apps/common/handle/impl/doc_split_handle.py
@@ -112,11 +112,7 @@ def get_image_id_func():
 
 title_font_list = [
     [36, 100],
-    [26, 36],
-    [24, 26],
-    [22, 24],
-    [18, 22],
-    [16, 18]
+    [30, 36]
 ]
 
 
@@ -130,7 +126,7 @@ def get_title_level(paragraph: Paragraph):
         if len(paragraph.runs) == 1:
             font_size = paragraph.runs[0].font.size
             pt = font_size.pt
-            if pt >= 16:
+            if pt >= 30:
                 for _value, index in zip(title_font_list, range(len(title_font_list))):
                     if pt >= _value[0] and pt < _value[1]:
                         return index + 1


apps/common/handle/impl/doc_split_handle.py
```diff
@@ -130,7 +126,7 @@ def get_title_level(paragraph: Paragraph):
         if len(paragraph.runs) == 1:
             font_size = paragraph.runs[0].font.size
             pt = font_size.pt
-            if pt >= 16:
+            if pt >= 30:
                 for _value, index in zip(title_font_list, range(len(title_font_list))):
                     if pt >= _value[0] and pt < _value[1]:
                         return index + 1
```
shaohuzhang1, line 132: The provided code has some minor improvements and corrections:

1. The `get_image_id` function is defined at the top but used further down, which means it might not be needed there unless you're reusing it elsewhere.

2. In the `get_title_level` function:
   - You've removed three sets of conditional checks that are essentially duplicating the check for `pt >= 30`. Only the last condition remains useful.

3. The list comprehension in `title_font_list` should likely include all available sizes rather than just smaller ones to cover all possible titles if they exist beyond the given range.

4. It's unclear why `pt >= 16` or any specific conditions (like `< 36`) were included for fonts larger than 30 points in `title_font_list`, as it would always match with `[30, 36]`.

Here's an improved version of the `get_title_level` function based on these considerations:

```python
def get_title_level(paragraph: Paragraph):
    if len(paragraph.runs) == 1:
        font_size = paragraph.runs[0].font.size
        pt = font_size.pt
        
        # Use binary search to find the appropriate title level
        left, right = 0, len(title_font_list) - 1
        while left <= right:
            mid = left + (right - left) // 2
            size_range = title_font_list[mid]
            if pt >= 30 and pt < size_range[1]:
                return mid + 1
            elif pt < size_range[0]:
                right = mid - 1
            else:
                left = mid + 1

    return 1  # Default level, typically H1
```

### Potential Optimization Suggestions:
- For better readability and maintainability, separate out each case into different functions or methods.
- If the number of title levels extends significantly, consider using a dictionary mapping instead of a list for faster lookups.
- Ensure that the logic handles edge cases correctly, such as when no relevant paragraphs are found.

PR #2955, opened 2025-04-22
fix: Change the max_token of the Qianfan large model to max_output_token

shaohuzhang1:
fix: Change the max_token of the Qianfan large model to max_output_token

### DIFF
--- a/apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
@@ -27,7 +27,7 @@ class WenxinLLMModelParams(BaseForm):
                                     _step=0.01,
                                     precision=2)
 
-    max_tokens = forms.SliderField(
+    max_output_tokens = forms.SliderField(
         TooltipLabel(_('Output the maximum Tokens'),
                      _('Specify the maximum number of tokens that the model can generate')),
         required=True, default_value=1024,


apps/setting/models_provider/impl/wenxin_model_provider/credential/llm.py
```diff
@@ -27,7 +27,7 @@ class WenxinLLMModelParams(BaseForm):
                                     _step=0.01,
                                     precision=2)
 
-    max_tokens = forms.SliderField(
+    max_output_tokens = forms.SliderField(
         TooltipLabel(_('Output the maximum Tokens'),
                      _('Specify the maximum number of tokens that the model can generate')),
         required=True, default_value=1024,
```
shaohuzhang1, line 33: The proposed change from `max_tokens` to `max_output_tokens` is valid and aligns with the current naming conventions in many APIs, which often use "output" at the end of parameter names to indicate they refer to what an API generates rather than receiving. This makes sense given the context of LLM (Language Model) parameters, especially since models typically output tokens or sentences.

### Potential Improvements:

1. **Documentation Clarity**: Consider adding more detailed documentation for both fields (`max_output_tokens` and potentially `_step`, `precision`) about their intended usage and expected values. This will help users better understand how these parameters affect the behavior of your application.

2. **Validation**: Ensure that the validation logic remains appropriate for both fields:
   - For `max_output_tokens`, set a reasonable upper limit based on typical text generation tasks.
   - Validate `_step` to ensure it meets expectations (e.g., non-negative numbers).
   - Validate `precision` similarly, ensuring it's within acceptable bounds.

3. **Consistent Naming**: If there are other slider fields involving token limits (like `max_input_tokens`), ensure consistent naming across all similar fields to maintain clarity and ease of use.

Overall, this refactoring appears to be well-intentioned and should work correctly as long as the existing codebase and business requirements are adhered to.

PR #2969, opened 2025-04-24
fix: Fixed interface parameter transmission not displaying during deb…

wangdan-fit2cloud:
…ugging

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -162,7 +162,10 @@ const initialApiFormData = ref({})
 const isUserInput = computed(
   () =>
     props.applicationDetails.work_flow?.nodes?.filter((v: any) => v.id === 'base-node')[0]
-      .properties.user_input_field_list.length > 0
+      .properties.user_input_field_list.length > 0 ||
+    (props.type === 'debug-ai-chat' &&
+      props.applicationDetails.work_flow?.nodes?.filter((v: any) => v.id === 'base-node')[0]
+        .properties.api_input_field_list.length > 0)
 )
 const showUserInputContent = computed(() => {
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/index.vue
```diff
@@ -162,7 +162,10 @@ const initialApiFormData = ref({})
...
   () =>
     props.applicationDetails.work_flow?.nodes?.filter((v: any) => v.id === 'base-node')[0]
-      .properties.user_input_field_list.length > 0
+      .properties.user_input_field_list.length > 0 ||
+    (props.type === 'debug-ai-chat' &&
+      props.applicationDetails.work_flow?.nodes?.filter((v: any) => v.id === 'base-node')[0]
+        .properties.api_input_field_list.length > 0)
 )
 const showUserInputContent = computed(() => {
   return ((isUserInput.value && firsUserInput.value) || showUserInput.value) && props.type !== 'log'
```
shaohuzhang1, line 171: There are no significant irregularities or major issues in this代码. However, here are a few points to consider:

1. **Type of `v`**: The line inside `.filter()` is using type assertion `(any)`, which can lead to runtime errors if not handled properly. It's better to specify the TypeScript types more explicitly.

2. **Complexity with Logical OR**: The last part of the logical OR condition checks for different properties based on the application type (`type`). If these conditions become too complex and nested multiple times, they might make the logic harder to read.

3. **Variable Naming**: Variables like `firstUserInput` should be defined before being used to avoid potential undefined variable usage errors during computation.

4. **Computed Property Name`: While it seems fine at first glance, ensure that dynamic computed property names will work as expected across different environments, especially if you plan to use this in frameworks that support template interpolation or other reactive contexts heavily relying on computed properties.

Overall, the code looks functional, but improving its efficiency through better typing could enhance maintainability and prevent future bugs related to invalid data processing.

PR #2971, opened 2025-04-24
fix: Thinking about process labels during simple application debugging does not take effect

shaohuzhang1:
fix: Thinking about process labels during simple application debugging does not take effect

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -148,10 +148,12 @@ class ModelSettingSerializer(serializers.Serializer):
                                                         error_messages=ErrMessage.char(_("Thinking process switch")))
     reasoning_content_start = serializers.CharField(required=False, allow_null=True, default="<think>",
                                                     allow_blank=True, max_length=256,
+                                                    trim_whitespace=False,
                                                     error_messages=ErrMessage.char(
                                                         _("The thinking process begins to mark")))
     reasoning_content_end = serializers.CharField(required=False, allow_null=True, allow_blank=True, default="</think>",
                                                   max_length=256,
+                                                  trim_whitespace=False,
                                                   error_messages=ErrMessage.char(_("End of thinking process marker")))
 
 


apps/application/serializers/application_serializers.py
```diff
@@ -148,10 +148,12 @@ class ModelSettingSerializer(serializers.Serializer):
                                                         error_messages=ErrMessage.char(_("Thinking process switch")))
     reasoning_content_start = serializers.CharField(required=False, allow_null=True, default="<think>",
                                                     allow_blank=True, max_length=256,
+                                                    trim_whitespace=False,
                                                     error_messages=ErrMessage.char(
                                                         _("The thinking process begins to mark")))
     reasoning_content_end = serializers.CharField(required=False, allow_null=True, allow_blank=True, default="</think>",
                                                   max_length=256,
+                                                  trim_whitespace=False,
                                                   error_messages=ErrMessage.char(_("End of thinking process marker")))
```
shaohuzhang1, line 159: There are no apparent irregularities or major issues in the provided code. However, there is a small concern regarding trimming whitespace with `trim_whitespace=False`. While this setting does not affect these particular fields (`reasoning_content_start` and `reasoning_content_end`) directly, it might be better to explicitly include the field's description within the error messages to make the intent more clear.

Here's an updated version of the serializer with comments explaining the rationale:

```python
class ModelSettingSerializer(serializers.Serializer):
    thinking_process_switch = serializers.BooleanField(
        required=False,
        default=False
    )

    # Start marker for thinking process
    reasoning_content_start = serializers.CharField(
        required=False,
        allow_null=True,
        default="<think>",
        allow_blank=True,
        max_length=256,
        trim_whitespace=False,
        help_text=_("This specifies the start of the thinking process."
                       " Default value '<think>'.")
    )
    
    # End marker for thinking process
    reasoning_content_end = serializers.CharField(
        required=False,
        allow_null=True,
        allow_blank=True,
        default="</think>",
        max_length=256,
        trim_whitespace=False,
        help_text=_("This specifies the end marker of the thinking process."
                        " If absent but 'thinking_process_switch' is True,"
                        " placeholders will be used.")
    )
```

In summary, ensure that any additional context or information about how the fields should be processed (e.g., using placeholders if both markers are not set) is clearly documented in the Django rest framework documentation. This helps prevent confusion during integration or maintenance.

PR #2972, opened 2025-04-24
fix: update post_cell function to handle different newline characters in cell values

shaohuzhang1:
fix: update post_cell function to handle different newline characters in cell values  --bug=1054683 --user=刘瑞斌 【github#2831】知识库上传excel、应用编排文档内容提取节点中上传excel，单元格中有换行，导入后没有在一个单元格里显示 https://www.tapd.cn/57709429/s/1690232

### DIFF
--- a/apps/common/handle/impl/xls_split_handle.py
+++ b/apps/common/handle/impl/xls_split_handle.py
@@ -14,7 +14,7 @@ from common.handle.base_split_handle import BaseSplitHandle
 
 
 def post_cell(cell_value):
-    return cell_value.replace('\n', '<br>').replace('|', '&#124;')
+    return cell_value.replace('\r\n', '<br>').replace('\n', '<br>').replace('|', '&#124;')
 
 
 def row_to_md(row):


apps/common/handle/impl/xls_split_handle.py
```diff
@@ -14,7 +14,7 @@
 
 
 def post_cell(cell_value):
-    return cell_value.replace('\n', '<br>').replace('|', '&#124;')
+    return cell_value.replace('\r\n', '<br>').replace('\n', '<br>').replace('|', '&#124;')
 
 
 def row_to_md(row):
```
shaohuzhang1, line 20: The provided code is mostly correct but has a small issue with handling line breaks (`\t`):

```python
@@ -14,7 +14,7 @@
 
 
 def post_cell(cell_value):
-    return cell_value.replace('\n', '<br>').replace('|', '&#124;')
+    return cell_value.replace('\r\n', '<br>').replace('\n', '<br>').replace('\t', '&nbsp;')  # Add this line to replace tabs with spaces
```

### Explanation:

1. **Replacement of `\n`:** Both `\n` and `\r\n` are used to denote line breaks in text. By replacing both `\n` and `\r\n`, you ensure consistency across different newline characters.

2. **Handling Tabs:** The code currently replaces tabs (\`\t`) with `&nbsp;` (non-breaking space). This may be necessary if tabular data needs to preserve its original formatting, especially when converting it to HTML or Markdown where automatic indentation might cause problems.

3. **Potential Issues:**
   - If the input data contains only `\t`s without any newlines, `post_cell` will not work as intended because it relies on `\n`.
   - Ensuring that all relevant whitespace characters (including tabs) are handled appropriately can make the function more robust for various inputs.

By adding the replacement for `\t`, you handle more edge cases and ensure that the converted markup preserves consistent spacing and structure.

PR #3018, opened 2025-04-29
fix: During the conversation, there is sensitive data in the application profile obtained by the application

shaohuzhang1:
fix: During the conversation, there is sensitive data in the application profile obtained by the application

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -1012,7 +1012,8 @@ class ApplicationSerializer(serializers.Serializer):
                  'stt_autosend': application.stt_autosend,
                  'file_upload_enable': application.file_upload_enable,
                  'file_upload_setting': application.file_upload_setting,
-                 'work_flow': application.work_flow,
+                 'work_flow': {'nodes': [node for node in application.work_flow.get('nodes') if
+                                         node.get('id') == 'base-node']},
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_serializers.py
```diff
@@ -1012,7 +1012,8 @@ def profile(self, with_valid=True):
                  'stt_autosend': application.stt_autosend,
                  'file_upload_enable': application.file_upload_enable,
                  'file_upload_setting': application.file_upload_setting,
-                 'work_flow': application.work_flow,
+                 'work_flow': {'nodes': [node for node in application.work_flow.get('nodes') if
+                                         node.get('id') == 'base-node']},
                  'show_source': application_access_token.show_source,
                  'language': application_access_token.language,
                  **application_setting_dict})
```
shaohuzhang1, line 1019: The main change in this function is removing all but one workflow node from `application.work_flow.nodes` before serializing it to JSON. This could potentially cause issues if the application relies on other nodes defined in the original workflow.

Potential improvements:
1. Remove unnecessary code: The line `if 'base-node' in [node.get('id') for node in ...]:` can be simplified or removed entirely since you're already checking against a specific ID later.
2. Consider adding error handling: It might be good to add a try-except block around operations involving parsing or accessing data within the workflow structure.

PR #3019, opened 2025-04-29
fix: Workflow application node parameter saving cannot be reflected back

shaohuzhang1:
fix: Workflow application node parameter saving cannot be reflected back

### DIFF
--- a/ui/src/workflow/nodes/application-node/index.vue
+++ b/ui/src/workflow/nodes/application-node/index.vue
@@ -238,49 +238,42 @@ const update_field = () => {
         const new_user_input_field_list = cloneDeep(
           ok.data.work_flow.nodes[0].properties.user_input_field_list
         )
-        const merge_api_input_field_list =
-          new_api_input_field_list ||
-          [].map((item: any) => {
-            const find_field = old_api_input_field_list.find(
-              (old_item: any) => old_item.variable == item.variable
-            )
-            if (find_field) {
-              return {
-                ...item,
-                value: find_field.value,
-                label:
-                  typeof item.label === 'object' && item.label != null
-                    ? item.label.label
-                    : item.label
-              }
-            } else {
-              return item
+
+        const merge_api_input_field_list = (new_api_input_field_list || []).map((item: any) => {
+          const find_field = old_api_input_field_list.find(
+            (old_item: any) => old_item.variable == item.variable
+          )
+          if (find_field) {
+            return {
+              ...item,
+              value: find_field.value,
+              label:
+                typeof item.label === 'object' && item.label != null ? item.label.label : item.label
             }
-          })
+          } else {
+            return item
+          }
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/application-node/index.vue
```diff
@@ -238,49 +238,42 @@ const update_field = () => {
...
+                typeof item.label === 'object' && item.label != null ? item.label.label : item.label
             }
-          })
+          } else {
+            return item
+          }
+        })
         set(
           props.nodeModel.properties.node_data,
           'user_input_field_list',
```
shaohuzhang1, line 279: In this code snippet, there are several potential issues that need to be addressed:

1. **Redundant Code**: The same logic is repeated for `merge_api_input_field_list` and `merge_user_input_field_list`. Consider creating a helper function or refactor these sections.

2. **Undefined Old API Input Field List**: If `old_api_input_field_list` can be undefined, it's good practice to handle that case properly.

3. **Type Annotations**: Ensure type annotations are correct for variables used throughout the function.

4. **String Literals**: Replace hardcoded string literals with constants where applicable to improve readability.

5. **Cloning Process**: The use of `cloneDeep()` from lodash might not always be necessary. You could consider using simple deep cloning if you don't rely on specific features provided by lodash.

Here's an optimized version of the code addressing some of these concerns:

```javascript
import { cloneDeep } from 'lodash';

const MERGE_FIELDS = [
  'variable', // For API input field list
  'field'      // For user input field list
];

// Helper function to create merged fields list
function mergeFields(newFieldList, oldFieldList, variableName) {
  return (
    newFieldList || []
  ).map(item => {
    const findField = oldFieldList.find(oldItem => {
      return oldItem[variableName] === item[variableName];
    });

    if (!findField) {
      return item;
    }

    return {
      ...item,
      value: findField.value,
      label:
        typeof findField.label === 'object'
          ? findField.label[label]
          : findField.label
    };
  });
}

const update_field = () => {
  const new_user_input_field_list = cloneDeep(
    ok.data.work_flow.nodes[0].properties.user_input_field_list
  );
  const new_api_input_field_list = cloneDeep(
    ok.data.work_flow.nodes[0].properties.api_input_field_list
  );

  const merge_api_input_field_list = mergeFields(
    new_api_input_field_list,
    old_api_input_field_list,
    'variable'
  );

  const merge_user_input_field_list = mergeFields(
    new_user_input_field_list,
    old_user_input_field_list,
    'field'
  );

  set(props.nodeModel.properties.node_data, 'api_input_field_list', merge_api_input_field_list);
  set(props.nodeModel.properties.node_data, 'user_input_field_list', merge_user_input_field_list);
};
```

By refactoring the code into reusable methods like `mergeFields`, we have reduced redundancy and improved maintainability. This approach also makes the code easier to test and understand.

PR #3020, opened 2025-04-29
feat: Support Alibaba Bailian supplier qwen3 model

shaohuzhang1:
feat: Support Alibaba Bailian supplier qwen3 model

### DIFF
--- a/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/llm.py
+++ b/apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/llm.py
@@ -47,7 +47,11 @@ class BaiLianLLMModelCredential(BaseForm, BaseModelCredential):
                     return False
         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=gettext('Hello'))])
+            if model_params.get('stream'):
+                for res in model.stream([HumanMessage(content=gettext('Hello'))]):
+                    pass
+            else:
+                model.invoke([HumanMessage(content=gettext('Hello'))])
         except Exception as e:
             traceback.print_exc()
             if isinstance(e, AppApiException):


apps/setting/models_provider/impl/aliyun_bai_lian_model_provider/credential/llm.py
```diff
@@ -47,7 +47,11 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
...
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=gettext('Hello'))])
+            if model_params.get('stream'):
+                for res in model.stream([HumanMessage(content=gettext('Hello'))]):
+                    pass
+            else:
+                model.invoke([HumanMessage(content=gettext('Hello'))])
         except Exception as e:
             traceback.print_exc()
             if isinstance(e, AppApiException):
```
shaohuzhang1, line 57: The code provided has a few minor improvements that can be made:

1. **Handling Stream Mode**: The `model.stream` method is only being called when `model_params.get('stream')` is truthy.
2. **Explicitly Handling None Types**: If `res` in the stream mode returns `None`, it will not throw an exception but simply go to the next iteration of the loop.

Here's the updated code with these improvements included:

```diff
@@ -47,7 +47,12 @@
                 return False

         try:
             model = provider.get_model(model_type, model_name, model_credential, **model_params)
-            model.invoke([HumanMessage(content=gettext('Hello'))])
+            if 'stream' in model_params and model_params['stream']:
+                for res in model.stream([HumanMessage(content=gettext('Hello'))]):
+                    # Handle streaming output here, e.g., print(res)
+                    pass
+            else:
+                model.invoke([HumanMessage(content=gettext('Hello'))])

         except Exception as e:
             traceback.print_exc()
             if isinstance(e, AppApiException):
```

Additionally, consider adding error handling around any operations within the stream loop or where `res` might be `None`.

PR #3021, opened 2025-04-29
fix: The similarity of knowledge base mixed retrieval should be 0-2

shaohuzhang1:
fix: The similarity of knowledge base mixed retrieval should be 0-2

### DIFF
--- a/ui/src/views/application/component/ParamSettingDialog.vue
+++ b/ui/src/views/application/component/ParamSettingDialog.vue
@@ -81,7 +81,7 @@
                 <el-input-number
                   v-model="form.dataset_setting.similarity"
                   :min="0"
-                  :max="form.search_mode === 'blend' ? 2 : 1"
+                  :max="form.dataset_setting.search_mode === 'blend' ? 2 : 1"
                   :precision="3"
                   :step="0.1"
                   :value-on-clear="0"


ui/src/views/application/component/ParamSettingDialog.vue
```diff
@@ -81,7 +81,7 @@
                 <el-input-number
                   v-model="form.dataset_setting.similarity"
                   :min="0"
-                  :max="form.search_mode === 'blend' ? 2 : 1"
+                  :max="form.dataset_setting.search_mode === 'blend' ? 2 : 1"
                   :precision="3"
                   :step="0.1"
                   :value-on-clear="0"
```
shaohuzhang1, line 87: The provided lines of code seem to be part of an implementation for a Vue.js component using Element Plus framework, with `form` being the state object containing various properties related to dataset settings. The specific concern is about the value of `form.dataset_setting.similarity`, which has a maximum limit that depends on the value of another property, `form.search_mode`.

Here’s what I recommend:

1. **Variable Consistency**: Ensure that all variables and constants used in conditionals match their expected values. In this case, both `form.search_mode` and `form.dataset_setting.search_mode` should ideally point to a single variable to avoid inconsistencies.

2. **Documentation**: It would improve readability to add comments explaining why the upper bound is 2 when `search_mode` is 'blend'. This might clarify future maintenance requirements.

   - If there are multiple places where the same calculation (`form.search_mode === 'blend' ? 2 : 1`) occurs, consider centralizing it into a helper function to maintain consistency and reduce duplication.
     
```javascript
// Helper function to determine max similarity based on search mode
function getMaxSimilarityModeBlend(searchMode) {
  return searchMode === 'blend' ? 2 : 1;
}

<el-input-number
  v-model="form.dataset_setting.similarity"
  :min="0"
  :max="getMaxSimilarityModeBlend(form.search_mode)"
  :precision="3"
  :step="0.1"
  :value-on-clear="0"
/>
```

This way, if you need to change the logic later (e.g., handling different modes differently), you only need to modify one place instead of many.

PR #3022, opened 2025-04-29
fix: Cannot copy applications with opening statement characters exceeding 4096

shaohuzhang1:
fix: Cannot copy applications with opening statement characters exceeding 4096

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -164,7 +164,7 @@ class ApplicationWorkflowSerializer(serializers.Serializer):
                                  max_length=256, min_length=1,
                                  error_messages=ErrMessage.char(_("Application Description")))
     work_flow = serializers.DictField(required=False, error_messages=ErrMessage.dict(_("Workflow Objects")))
-    prologue = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=4096,
+    prologue = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=102400,
                                      error_messages=ErrMessage.char(_("Opening remarks")))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_serializers.py
```diff
@@ -495,7 +495,7 @@ class Edit(serializers.Serializer):
                                                    min_value=0,
                                                    max_value=1024,
                                                    error_messages=ErrMessage.integer(_("Historical chat records")))
-        prologue = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=4096,
+        prologue = serializers.CharField(required=False, allow_null=True, allow_blank=True, max_length=102400,
                                          error_messages=ErrMessage.char(_("Opening remarks")))
         dataset_id_list = serializers.ListSerializer(required=False, child=serializers.UUIDField(required=True),
                                                      error_messages=ErrMessage.list(_("Related Knowledge Base"))
```
shaohuzhang1, line 501: There are a couple of issues in this code snippet:
1. The `allow_blank=True` parameter is typically used with fields that accept Unicode strings (like text areas) instead of character strings (like single-line text). So the appropriate field type should be something like `CharField`, not `Serializers.CharField`. You can remove the `allow_blank=True` if you intended to use something else.

Here's an optimized version assuming UTF-8 encoding:
```python
class Edit(serializers.Serializer):
    historical_chat_count = serializers.IntegerField(
                                required=False,
                                default=0,
                                min_value=0,
                                max_value=1024,
                                error_messages=dict(min_value=_('Minimum value: 0'),
                                                  max_value=_('Maximum value: 1024'))
                            )
    prologue = serializers Charfield(
                      required=False,
                      blank=True,
                      max_length=102400,   # Assuming maximum length of utf-8 characters
                      error_messages=dict(required=_('Prologue is required'))
                  )
    dataset_id_list = serializers.ListSerializer(
                         required=False,
                         child=serializers.UUIDField(),
                         allow_null=True,
                         error_messages=dict(max_length=_('Dataset ID list cannot be longer than allowed'))
                     )
```
Additionally, ensure all other fields follow the same logic as well, using the correct fields from Django REST Framework and adhering to best naming conventions for variables and attributes within classes.

PR #3028, opened 2025-04-29
fix: The simple application demonstration cannot be opened

shaohuzhang1:
fix: The simple application demonstration cannot be opened

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -1012,7 +1012,7 @@ class ApplicationSerializer(serializers.Serializer):
                  'stt_autosend': application.stt_autosend,
                  'file_upload_enable': application.file_upload_enable,
                  'file_upload_setting': application.file_upload_setting,
-                 'work_flow': {'nodes': [node for node in application.work_flow.get('nodes') if
+                 'work_flow': {'nodes': [node for node in ((application.work_flow or {}).get('nodes', []) or []) if
                                          node.get('id') == 'base-node']},
                  'show_source': application_access_token.show_source,
                  'language': application_access_token.language,


apps/application/serializers/application_serializers.py
```diff
@@ -1012,7 +1012,7 @@ def profile(self, with_valid=True):
                  'stt_autosend': application.stt_autosend,
                  'file_upload_enable': application.file_upload_enable,
                  'file_upload_setting': application.file_upload_setting,
-                 'work_flow': {'nodes': [node for node in application.work_flow.get('nodes') if
+                 'work_flow': {'nodes': [node for node in ((application.work_flow or {}).get('nodes', []) or []) if
                                          node.get('id') == 'base-node']},
                  'show_source': application_access_token.show_source,
                  'language': application_access_token.language,
```
shaohuzhang1, line 1018: Looks good! There are no major irregularities or issues found in the code. However, it's worth noting that accessing `application.work_flow` twice is redundant when using the ternary operator to handle potential None value scenarios. Here's a simplified version:

```python
self.config = {
    "stt_autosend": application.stt_autosend,
    "file_upload_enable": application.file_upload_enable,
    "file_upload_setting": application.file_upload_setting,
    "work_flow": {  # Handle case where work_flow might not exist or be empty
        "nodes": [
            node for node in (application.work_flow.get("nodes", [])
                             if isinstance(application.work_flow, dict)
                             and 'nodes' in application_work_flow)
                              if node.get('id') == 'base-node']
    },
    "show_source": application_access_token.show_source,
    "language": application_access_token.language
}
```

This approach ensures that you only access the dictionary once and includes checks to ensure that both `application.work_flow` and its `nodes` key exist before processing them.

PR #3089, opened 2025-05-15

shaohuzhang1:

### DIFF
--- a/apps/common/config/embedding_config.py
+++ b/apps/common/config/embedding_config.py
@@ -11,7 +11,7 @@ import time
 
 from common.cache.mem_cache import MemCache
 
-lock = threading.Lock()
+_lock = threading.Lock()
 
 
 class ModelManage:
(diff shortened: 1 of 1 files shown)


apps/common/config/embedding_config.py
```diff
@@ -20,26 +20,27 @@ class ModelManage:
...
     @staticmethod
     def clear_timeout_cache():
-        if time.time() - ModelManage.up_clear_time > 60:
-            ModelManage.cache.clear_timeout_data()
+        if time.time() - ModelManage.up_clear_time > 60 * 60:
+            threading.Thread(target=lambda: ModelManage.cache.clear_timeout_data()).start()
+            ModelManage.up_clear_time = time.time()
 
     @staticmethod
     def delete_key(_id):
```
shaohuzhang1, line 46: Here are some optimizations and improvements:

1. Use `try-with-resources` style context managers to automatically release locks:
```python
 with _lock:
```

2. Simplify cache set logic by directly using the lock within it:
   ```python
   def get_model(self, _id, get_model):
       model_instance = self.cache.get(_id)
       if not model_instance or not model_instance.is_cache_model():
           with _lock:
               model_instance = get_model(_id)
               self.cache.set(
                   _id,
                   model_instance,
                   timeout=60 * 3600,  # Increased from 30 minutes to an hour
               )
               self.clear_timeout_cache()
               return model_instance
       elif model_instance.is_cache_model():
           self.cache.touch(_id, timeout=60 * 3600)  # Increased from 30 minutes to an hour
           return model_instance
       else:
           self.delete_key(_id)
           return get_model(_id)
   ```
   
3. Move the cache clearing logic into its own method called `clear_expired_keys`, so that you can call it explicitly when needed without modifying `get_model`.

4. Ensure that `ModelManage.up_clear_time` is properly initialized before being used.

5. The `time.sleep(3)` in your original code was removed because there's no reason for sleeping between each request. If sleep is required somewhere else, ensure that it doesn't interfere with performance or functionality of this critical service. 

This optimized version maintains the necessary logic while improving readability and efficiency. Always test thoroughly after making changes to ensure stability!

PR #3091, opened 2025-05-15

shaohuzhang1:

### DIFF
--- a/apps/common/config/embedding_config.py
+++ b/apps/common/config/embedding_config.py
@@ -25,16 +25,14 @@ class ModelManage:
             with _lock:
                 model_instance = get_model(_id)
                 ModelManage.cache.set(_id, model_instance, timeout=60 * 60 * 8)
-                ModelManage.clear_timeout_cache()
-                return model_instance
         else:
             if model_instance.is_cache_model():
                 ModelManage.cache.touch(_id, timeout=60 * 60 * 8)
-                return model_instance
             else:
                 model_instance = get_model(_id)
                 ModelManage.cache.set(_id, model_instance, timeout=60 * 60 * 8)
-                return model_instance
+        ModelManage.clear_timeout_cache()
+        return model_instance
 
     @staticmethod
     def clear_timeout_cache():


apps/common/config/embedding_config.py
```diff
@@ -25,16 +25,14 @@ def get_model(_id, get_model):
...
-                return model_instance
             else:
                 model_instance = get_model(_id)
                 ModelManage.cache.set(_id, model_instance, timeout=60 * 60 * 8)
-                return model_instance
+        ModelManage.clear_timeout_cache()
+        return model_instance
 
     @staticmethod
     def clear_timeout_cache():
```
shaohuzhang1, line 38: The provided code seems correct and efficient. Here are a few minor improvements and notes:

1. **Return Value**: The function `get_model` should have only one return point without unnecessary nested conditions after the caching logic.

2. **Consistent Code Blocks**: Ensure consistent indentation throughout the code block to improve readability.

3. **Variable Naming**: Variable names like `_lock`, `_id`, `model_instance`, and others could be more descriptive for better understanding of their purpose within the context of the function.

4. **Performance Considerations**: Since you're setting timeouts for cachable instances but not retrieving them again before they expire, this pattern is efficient for ensuring hot data remains cached until it's no longer needed due to updates or expiration.

Here’s an optimized version of the code with these considerations:

```python
@@ -25,16 +25,13 @@ def get_model(_id, get_model):
             with _lock:
                 model_instance = get_model(_id)
                 ModelManage.cache.set(_id, model_instance, timeout=8 * 60 * 60)  # Set a fixed duration rather than using another method call
-                ModelManage.clear_timeout_cache()  # Only clear cache when a new instance is fetched
         else:
             if model_instance.is_cache_model():
                 ModelManage.cache.touch(_id, timeout=8 * 60 * 60)  # Extend the cache lifetime of existing cached models
             else:
                 model_instance = get_model(_id)
                 ModelManage.cache.set(_id, model_instance, timeout=8 * 60 * 60)  # Always set new instances with a fixed timeout

     @staticmethod
     def clear_timeout_cache(*args, **kwargs):  # Adjusting parameters based on expected usage
     ```

These changes simplify the logic and ensure that resources are used efficiently while maintaining clarity and simplicity in the codebase.

PR #3098, opened 2025-05-16

shaohuzhang1:

### DIFF
--- a/apps/common/util/rsa_util.py
+++ b/apps/common/util/rsa_util.py
@@ -40,15 +40,12 @@ def generate():
 def get_key_pair():
     rsa_value = rsa_cache.get(cache_key)
     if rsa_value is None:
-        lock.acquire()
-        rsa_value = rsa_cache.get(cache_key)
-        if rsa_value is not None:
-            return rsa_value
-        try:
+        with lock:
+            rsa_value = rsa_cache.get(cache_key)
+            if rsa_value is not None:
+                return rsa_value
             rsa_value = get_key_pair_by_sql()
             rsa_cache.set(cache_key, rsa_value)
-        finally:
-            lock.release()
     return rsa_value
 
 


apps/common/util/rsa_util.py
```diff
@@ -40,15 +40,12 @@ def generate():
...
-        try:
+        with lock:
+            rsa_value = rsa_cache.get(cache_key)
+            if rsa_value is not None:
+                return rsa_value
             rsa_value = get_key_pair_by_sql()
             rsa_cache.set(cache_key, rsa_value)
-        finally:
-            lock.release()
     return rsa_value
```
shaohuzhang1, line 51: There is no error in this code, but there is an inefficiency in obtaining keys. If a key pair already exists in the cache, it does not need to fetch one from the database each time because `rsa_cache.set(cache_key, rsa_value)` will always overwrite the existing value in the cache even if the key pair hasn't changed.

To optimize this function, you can modify it so that only calls to `get_key_pair_by_sql()` actually fetch data from the database:

```python
def generate():
    def get_key_pair():
        rsa_value = rsa_cache.get(cache_key)
        if rsa_value is None:
            with lock:
                rsa_value = rsa_cache.get(cache_key)
                if rsa_value is not None:
                    return rsa_value
                # Fetch keys from database only if they're missing.
                rsa_value = get_key_pair_by_sql() if not has_new_keys else None
                if rsa_value is not None:
                    rsa_cache.set(cache_key, rsa_value)
                    has_new_keys = True  # Reset flag after adding new values.
        return rsa_value

    has_new_keys = False
    generate()
```

If we know in advance when a full update is likely (for instance, during system startup), `has_new_keys` could be set once at the start of application initialization phase instead of checking every access to `generate`. This avoids unnecessary queries while still providing quick response on demand unless a full refresh might be needed sooner due to some event.

Issue #3053, opened 2025-05-07
[Bug] An error will be reported when the custom non-required parameter of a custom function is empty


hello1534:
### MaxKB Version

v1.10.5

### Problem Description

自定义函数的自定义非必填参数为空值时会报错

### Steps to Reproduce

![Image](https://github.com/user-attachments/assets/799242b2-1afc-4d71-8dfb-254abc9de554)

![Image](https://github.com/user-attachments/assets/0789e537-1104-415a-babb-9faa11b81dae)
还有在函数库中调试时也会

![Image](https://github.com/user-attachments/assets/4f70bb31-d51f-497e-b46b-1add4c03a6f8)

### The expected correct result

正常返回None

### Related log output

```shell

```

Shenguobin0102:
感谢反馈，后续版本优化

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version optimization

baixin513:
v1.10.8已发布

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.8 has been released

PR #3107, opened 2025-05-19
fix: When the function parameter is of non string type, if the parameter is empty, the function will run with an error #3053

shaohuzhang1:
fix: When the function parameter is of non string type, if the parameter is empty, the function will run with an error #3053

### DIFF
--- a/apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
+++ b/apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
@@ -65,7 +65,7 @@ def valid_reference_value(_type, value, name):
 
 
 def convert_value(name: str, value, _type, is_required, source, node):
-    if not is_required and value is None:
+    if not is_required and (value is None or (isinstance(value, str) and len(value) == 0)):
         return None
     if not is_required and source == 'reference' and (value is None or len(value) == 0):
         return None
--- a/apps/application/flow/step_node/function_node/impl/base_function_node.py
+++ b/apps/application/flow/step_node/function_node/impl/base_function_node.py
@@ -49,7 +49,7 @@ def valid_reference_value(_type, value, name):
 
 
 def convert_value(name: str, value, _type, is_required, source, node):
-    if not is_required and value is None:
+    if not is_required and (value is None or (isinstance(value, str) and len(value) == 0)):
         return None
     if source == 'reference':
         value = node.workflow_manage.get_reference_field(


apps/application/flow/step_node/function_lib_node/impl/base_function_lib_node.py
```diff
@@ -65,7 +65,7 @@ def valid_reference_value(_type, value, name):
 
 
 def convert_value(name: str, value, _type, is_required, source, node):
-    if not is_required and value is None:
+    if not is_required and (value is None or (isinstance(value, str) and len(value) == 0)):
         return None
     if not is_required and source == 'reference' and (value is None or len(value) == 0):
         return None
```
shaohuzhang1, line 71: The provided code has a few concerns:

1. **Redundant Check**: The second `if` statement checks both if `source` is `'reference'` and if the value is empty. This redundancy can be resolved by checking only one condition.

2. **String Length Consideration**: The current implementation does not explicitly handle cases where the input could contain non-string values that should also be considered "empty," although it implicitly handles them because `None` is handled separately.

3. **Immutability of List**: Although this doesn't appear to cause an issue with the given code, lists cannot be used immutably in certain contexts without explicit conversion (e.g., using tuple assignment).

Here's a revised version with these considerations addressed:

```python
def valid_reference_value(_type, value, name):
    
def convert_value(name: str, value, _type, is_required, source, node):
    if value is None and is_required:
        raise ValueError(f"Required field '{name}' missing.")

    # Convert string to title case unless required specifically
    if _type == 'title-cased':
        if isinstance(value, str) and len(value.strip()) > 0:
            return value.title()

    # Handle null or blank values appropriately based on requirement and source
    if not is_required and (value is None or 
                               (isinstance(value, str) and len(value.strip()) == 0)):
        return None

```

### Key Changes:
- Removed redundant check inside the second `if`.
- Added logic for converting strings to title case when `_type` is `'title-cased'` and there is no content.
- Raised a `ValueError` for required fields that are `None`.

These changes help improve readability, error handling, and potentially performance depending on context.

apps/application/flow/step_node/function_node/impl/base_function_node.py
```diff
@@ -49,7 +49,7 @@ def valid_reference_value(_type, value, name):
 
 
 def convert_value(name: str, value, _type, is_required, source, node):
-    if not is_required and value is None:
+    if not is_required and (value is None or (isinstance(value, str) and len(value) == 0)):
         return None
     if source == 'reference':
         value = node.workflow_manage.get_reference_field(
```
shaohuzhang1, line 55: The provided function `convert_value` has a minor issue with handling empty string conversions from references:

**Issue**: The current check `(isinstance(value, str) and len(value) == 0)` will always result in `False`, because both conditions must be `True`. If the condition fails at all (`None`), the outer check will automatically handle it.

**Suggestion**:
Remove the first condition that checks for an empty string when the value is already `None`.

Here's the updated code snippet:

```python
@@ -49,7 +49,6 @@ def valid_reference_value(_type, value, name):

def convert_value(name: str, value, _type, is_required, source, node):
    if not is_required and value is None:
        return None
    
    # Check if source is reference
    if source == 'reference':
        # Get value using workflow_manage.get_reference_field method
```

This ensures that the conversion logic only executes if the value is not `None`.

PR #3108, opened 2025-05-19
fix: Advanced orchestration defines global variables and assigns them through assignment nodes. If there is a form collection node before the reference node, the assignment node becomes invalid and the initial variable value is still referenced

shaohuzhang1:
fix: Advanced orchestration defines global variables and assigns them through assignment nodes. If there is a form collection node before the reference node, the assignment node becomes invalid and the initial variable value is still referenced

### DIFF
--- a/apps/application/flow/step_node/start_node/impl/base_start_node.py
+++ b/apps/application/flow/step_node/start_node/impl/base_start_node.py
@@ -45,6 +45,8 @@ class BaseStartStepNode(IStarNode):
         self.err_message = details.get('err_message')
         for key, value in workflow_variable.items():
             workflow_manage.context[key] = value
+        for item in details.get('global_fields', []):
+            workflow_manage.context[item.get('key')] = item.get('value')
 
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/start_node/impl/base_start_node.py
```diff
@@ -45,6 +45,8 @@ def save_context(self, details, workflow_manage):
         self.err_message = details.get('err_message')
         for key, value in workflow_variable.items():
             workflow_manage.context[key] = value
+        for item in details.get('global_fields', []):
+            workflow_manage.context[item.get('key')] = item.get('value')
 
     def get_node_params_serializer_class(self) -> Type[serializers.Serializer]:
         pass
```
shaohuzhang1, line 52: There are no notable issues or optimizations in the given code. The function `save_context` updates a context dictionary based on the inputs and then proceeds to handle global fields if they exist in the input dictionary. However, the documentation provided is not complete; you should include method docstrings for all functions (like `get_node_params_serializer_class`) that do not have them yet.

PR #3121, opened 2025-05-21
fix: Verification code is not case sensitive

shaohuzhang1:
fix: Verification code is not case sensitive

### DIFF
--- a/apps/users/serializers/user_serializers.py
+++ b/apps/users/serializers/user_serializers.py
@@ -64,7 +64,7 @@ class CaptchaSerializer(ApiMixin, serializers.Serializer):
         image = ImageCaptcha()
         data = image.generate(chars)
         captcha = base64.b64encode(data.getbuffer())
-        captcha_cache.set(f"LOGIN:{chars}", chars, timeout=5 * 60)
+        captcha_cache.set(f"LOGIN:{chars.lower()}", chars, timeout=5 * 60)
         return 'data:image/png;base64,' + captcha.decode()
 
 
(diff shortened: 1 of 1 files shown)


apps/users/serializers/user_serializers.py
```diff
@@ -105,7 +105,7 @@ def is_valid(self, *, raise_exception=False):
         """
         super().is_valid(raise_exception=True)
         captcha = self.data.get('captcha')
-        captcha_value = captcha_cache.get(f"LOGIN:{captcha}")
+        captcha_value = captcha_cache.get(f"LOGIN:{captcha.lower()}")
         if captcha_value is None:
             raise AppApiException(1005, _("Captcha code error or expiration"))
         username = self.data.get("username")
```
shaohuzhang1, line 111: The provided code snippet has a few improvements and clarifications that can be made:

1. **Case Insensitivity Fix**: The `captcha_cache` keys should be consistent to avoid errors related to case sensitivity. Change both instances where keys are formed from `chars` (line 63) and `captcha_value` (line 108). Use lowercase for uniformity, as shown:
   ```python
   captcha_cache.set(f"LOGIN:{captcha.lower()}", chars, timeout=5 * 60)
   ```

2. **Code Readability**: Consider adding comments at the beginning of your functions to explain what they do briefly, especially if they are large.

3. **Error Handling**: Ensure you handle exceptions correctly within the `is_valid` method to provide meaningful feedback to the user when the captcha code is invalid or expires.

4. **Security Advice**: Although not directly relevant to the code itself, it's worth considering that using passwords or captchas in plain text storage could expose sensitive information. In practice, use environment variables or secure vaults to store such credentials rather than hardcoding them in your application.

With these changes, the code will behave more robustly and securely while maintaining its functionality.

PR #3124, opened 2025-05-21
fix: The default suffix for workflow file upload nodes is set to uppercase DOC and PPT

shaohuzhang1:
fix: The default suffix for workflow file upload nodes is set to uppercase DOC and PPT

### DIFF
--- a/ui/src/workflow/nodes/base-node/component/FileUploadSettingDialog.vue
+++ b/ui/src/workflow/nodes/base-node/component/FileUploadSettingDialog.vue
@@ -56,7 +56,7 @@
                 <p class="line-height-22 mt-4">
                   {{ $t('common.fileUpload.document') }}
                   <el-text class="color-secondary"
-                  >{{
+                    >{{
                       $t(
                         'views.applicationWorkflow.nodes.baseNode.FileUploadSetting.fileUploadType.documentText'
                       )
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/base-node/component/FileUploadSettingDialog.vue
```diff
@@ -141,7 +141,7 @@
                 <p class="line-height-22 mt-4">
                   {{ $t('common.fileUpload.other') }}
                   <el-text class="color-secondary"
-                  >{{
+                    >{{
                       $t(
                         'views.applicationWorkflow.nodes.baseNode.FileUploadSetting.fileUploadType.otherText'
                       )
```
shaohuzhang1, line 147: There seem to be two instances where there is an extra `>` character after `$t()` calls:
```html
<el-text class="color-secondary">{{
```
Ensure that these should instead be:
```html
<el-text class="color-secondary">{{ --}}
```

Additionally, it's a good practice to use single quotes around template literals to avoid syntax errors, especially when mixing inline and multi-line templates. Here’s how you can fix them:

Before:
```javascript
'{{ $t('someKey', 'arg1 arg2') }}'
```

After replacing double/single quotes for consistency:
```javascript
"$t('someKey', 'arg1 args2')"
```
Or using single quotes if prefered for clarity within the string itself:
```javascript
'$t(\'someKey\', \'arg1 args2\')'
```

PR #3125, opened 2025-05-21
fix: The default suffix for workflow file upload nodes is set to uppercase DOC and PPT

shaohuzhang1:
fix: The default suffix for workflow file upload nodes is set to uppercase DOC and PPT

### DIFF
--- a/ui/src/workflow/nodes/base-node/index.vue
+++ b/ui/src/workflow/nodes/base-node/index.vue
@@ -316,7 +316,7 @@ const switchFileUpload = () => {
     audio: false,
     video: false,
     other: false,
-    otherExtensions: ['ppt', 'doc']
+    otherExtensions: ['PPT', 'DOC']
   }
 
   if (form_data.value.file_upload_enable) {


ui/src/workflow/nodes/base-node/index.vue
```diff
@@ -316,7 +316,7 @@ const switchFileUpload = () => {
     audio: false,
     video: false,
     other: false,
-    otherExtensions: ['ppt', 'doc']
+    otherExtensions: ['PPT', 'DOC']
   }
 
   if (form_data.value.file_upload_enable) {
```
shaohuzhang1, line 322: The code you provided has one minor issue that needs to be addressed:

The `otherExtensions` array now contains uppercase extensions ('PPT', 'DOC') instead of lowercase ('ppt', 'doc'). This may cause mismatches between file types detected on the client side and those stored in your backend/database.

To make sure both client-side and server-side representations are consistent, it would be best to use the same case throughout. If this is acceptable based on how your system handles files, you can remove this change. Alternatively, ensure all other places where these extensions are matched conform to this new casing convention. 

Here's an updated version with no changes needed for your specific question but still providing context:

```javascript
const switchFileUpload = () => {
  // ... rest of the function remains unchanged ...
}

if (form_data.value.file_upload_enable) {
  // ... continue with the rest of the logic if necessary ...
}
```

No need to add or remove anything else unless you specifically want to handle filenames differently on either end. The rest of the code looks clean apart from the extension typo mentioned above.

PR #3129, opened 2025-05-22
fix: The dropdown data of subsequent nodes in the form cannot be displayed back

shaohuzhang1:
fix: The dropdown data of subsequent nodes in the form cannot be displayed back

### DIFF
--- a/ui/src/workflow/nodes/form-node/index.ts
+++ b/ui/src/workflow/nodes/form-node/index.ts
@@ -4,6 +4,24 @@ class FormNode extends AppNode {
   constructor(props: any) {
     super(props, FormNodeVue)
   }
+  get_node_field_list() {
+    const result = []
+    const fields = this.props.model.properties?.config?.fields || []
+    let otherFields = []
+    try {
+      otherFields = this.props.model.properties.node_data.form_field_list.map((item: any) => ({
+        label: typeof item.label == 'string' ? item.label : item.label.label,
+        value: item.field
+      }))
+    } catch (e) {}
+    result.push({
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/form-node/index.ts
```diff
@@ -4,6 +4,24 @@ class FormNode extends AppNode {
...
+      value: this.props.model.id,
+      label: this.props.model.properties.stepName,
+      type: this.props.model.type,
+      children: [...fields, ...otherFields]
+    })
+    return result
+  }
 }
 export default {
   type: 'form-node',
```
shaohuzhang1, line 27: The provided code has a couple of issues:

1. **Incorrect Constructor Call**: The `AppNodeVue` import is missing from the constructor call (`super()`). It should be `super(props, FormModel)` where `FormModel` is assumed to be an alternative component that replaces `FormNodeVue`.

2. **Type Casting Issue**: In JavaScript/TypeScript, you cannot directly assign a string property ('label') with another TypeScript object without explicit type casting. You can use parentheses around each assignment inside the map function.

3. **Return Type**: The function signature indicates it returns an array but actually assigns results to variables instead before returning them. Consider modifying the logic to accumulate results in memory and then return once after all operations are completed.

Optimization Suggestions:
- Ensure proper error handling for the field list mapping; consider logging errors or throwing exceptions at appropriate levels.
- If the form data might change frequently, consider memoizing the derived state to improve performance.
- Use more descriptive variable names and ensure consistent naming conventions within classes.

PR #3131, opened 2025-05-22
fix: The dropdown data of subsequent nodes in the form cannot be displayed back

shaohuzhang1:
fix: The dropdown data of subsequent nodes in the form cannot be displayed back

### DIFF
--- a/ui/src/workflow/nodes/form-node/index.ts
+++ b/ui/src/workflow/nodes/form-node/index.ts
@@ -7,18 +7,22 @@ class FormNode extends AppNode {
   get_node_field_list() {
     const result = []
     const fields = this.props.model.properties?.config?.fields || []
-    let otherFields = []
+
     try {
-      otherFields = this.props.model.properties.node_data.form_field_list.map((item: any) => ({
-        label: typeof item.label == 'string' ? item.label : item.label.label,
-        value: item.field
-      }))
+      this.props.model.properties.node_data.form_field_list.forEach((item: any) => {
+        if (!fields.some((f: any) => f.value === item.field)) {
+          fields.push({
+            value: item.field,
+            label: typeof item.label == 'string' ? item.label : item.label.label
+          })
+        }
+      })
     } catch (e) {}
     result.push({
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/form-node/index.ts
```diff
@@ -7,18 +7,22 @@ class FormNode extends AppNode {
...
     } catch (e) {}
     result.push({
       value: this.props.model.id,
       label: this.props.model.properties.stepName,
       type: this.props.model.type,
-      children: [...fields, ...otherFields]
+      children: fields
     })
     return result
   }
```
shaohuzhang1, line 28: Here is a review of the provided code:

### Issues:
1. **Redundancy**: The `otherFields` array is created but not used within the scope where it's defined, leading to potential memory wastage.
2. **Simplification**: Instead of mapping over `form_field_list`, you can use an arrow function directly. This reduces unnecessary steps.
3. **Type Checking**: The code assumes that all properties have a non-null `.properties` object with nested objects. Ensure this assumption holds true.

### Optimizations:
1. **Remove Unnecessary Variable Assignment**: Since `fields` will be populated based on `form_field_list`, there’s no need to create `otherFields`.

Here's the revised code:

```javascript
class FormNode extends AppNode {
  get_node_field_list() {
    const result = [];
    
    try {
      this.props.model.properties.node_data &&
      this.props.model.properties.node_data.form_field_list.length > 0 &&
      
      this.props.model.properties.node_data.form_field_list.forEach((item: any) => {
        // Check if the field already exists in fields
        if (!fields.includes(item.field)) {
          fields.push({
            value: item.field,
            label: typeof item.label == 'string' ? item.label : item.label.label
          });
        }
      });
      
    } catch (e) {}

    result.push({
      value: this.props.model.id,
      label: this.props.model.properties.stepName,
      type: this.props.model.type,
      children: fields || []
    });

    return result;
  }
}
```

This refactored version of your method should perform better and reduce memory usage without any functionality changes.

PR #3176, opened 2025-05-30
feat: application

shaohuzhang1:
feat: application


apps/chat/serializers/chat_embed_serializers.py
```diff
@@ -0,0 +1,103 @@
...
+                        input_field_list = node.get('properties', {}).get('api_input_field_list',
+                                                                          node.get('properties', {}).get(
+                                                                              'input_field_list', []))
+                        if input_field_list is not None:
+                            for field in input_field_list:
+                                if field['assignment_method'] == 'api_input' and field['variable'] in params:
+                                    query += f"&{field['variable']}={params[field['variable']]}"
+        if 'asker' in params:
+            query += f"&asker={params.get('asker')}"
+        return query
```
shaohuzhang1, line 103: The provided code snippet appears to be an implementation of a serializer for embedding functionality in a Django REST Framework project. It includes logic for validating a request, reading an embedded JavaScript file, fetching necessary data from other models, and rendering the HTML content as a JavaScript response.

Here's a summary of potential issues, improvements, and suggestions:

### Issues:
1. **Security**: The code reads files from disk using `os.path.join` and opens them immediately without ensuring they exist or checking permissions.
2. **Database Access Efficiency**: The `QuerySet` retrieval inside loops can lead to repeated queries if there are many nodes or fields.
3. **Template Injection Risk**: The rendered template string might have security implications due to the way variables are injected into it (`render(Context(...))`).
4. **Unused Variables**: Some variables (e.g., `xpack_license_is_valid`, `header_font_color`) may not be used in every scenario but are still set.

### Improvements and Suggestions:
1. **File Existence Check**:
   ```python
   import os

   index_path = os.path.join(PROJECT_DIR, 'apps', "chat", 'template', 'embed.js')

   if not os.path.isfile(index_path):
       raise FileNotFoundError(f"Embedding script does not exist at {index_path}")
   ```

2. **Efficient Database Retrieval**:
   Use caching or bulk lookups where possible to avoid redundant database queries. For example, if you frequently need application settings, consider caching them.
   ```python
   @staticmethod
   def get_application_settings(application, cache):
       key = f'app_settings_{application.id}'
       cached_settings = cache.get(key)
       if cached_settings:
           return cached_settings

       application_setting_model = DatabaseModelManage.get_model('application_setting')
       if application_setting_model is not None:
           try:
               application_setting = QuerySet(application_setting_model).filter(
                   application_id=application.id).first()
               if application_setting:
                   cache.set(key, application_setting)
                   return application_setting
           except Exception as e:
               # Log the error
               print(f"Error fetching application setting: {e}")

   # Example usage
   application = ...
   application_settings = ChatEmbedSerializer.get_application_settings(application, your_cache_instance)
   ```

3. **Sanitize Input**:
   Ensure that input parameters are sanitized before being used in SQL queries or directly in context templates. Consider using parameterized queries for database operations.
   ```python
   def safe_get(value, default=None):
       return value.strip() if isinstance(value, str) else default

   float_icon = safe_get(float_icon, "")
   white_list = [safe_get(item.strip(), "") for item in application_access_token.white_list]
   ```

4. **Avoid Repeated Lookups**:
   If certain objects like `ApplicationAccessToken` and related models are fetched multiple times, consider combining the calls into one or use appropriate caching mechanisms.
   ```python
   def fetch_required_objects(token_str, cache):
       # Fetch once, store in a dictionary to reuse later
       key = f'required_object_{token_str}'
       stored_objects = cache.get(key)

       if not stored_objects:
           tokens = Token.objects.filter(access_token__icontains=token_str)
           applications = ApplicationToken.objects.filter(application__work_flow=tokens)
           stored_objects = {
               tk.access_token: tk.application for tk in tokens}
        
           cache.set(key, stored_objects)

       required_objects = stored_objects[token_str] if token_str in stored_objects else None
       
       # Return individual objects lazily as needed
       def get_object_by_name(name):
           obj, created = Object.objects.update_or_create(
               name=name, defaults={'description': f'Description for {name}'})
           return obj

       # Usage example
       required_applications = [
           {'id': app.id, 'name': get_object_by_name(app.name)}
           for app in required_objects.values()
           ]

       return required_applications

   result = fetch_required_objects("yourTokenStr", your_cache_instance)
   ```

5. **Code Documentation**:
   Enhance docstrings throughout the code to clearly describe what each part does, especially critical functions and methods like `get_query_api_input` and `fetch_required_objects`.
   ```python
   class ChatEmbedSerializer(serializers.Serializer):
       """
       A Serializer for embedding functionality.
      
       Attributes:
           host (str): The host URL.
           protocol (str): The protocol (http or https).
           token (str): The access token.
   
       Methods:
           get_embed(): Generates the embed JavaScript code based on input parameters.
           get_query_api_input(): Retrieves API input parameters defined in the application workflow.
       """

       # ... rest of the class definition ...

       @staticmethod
       def get_request_data(params):
           """
           Sanitizes and retrieves input data from the request.
           
           Args:
                   params (dict): Request parameters containing inputs required for generation.
               
           Returns:
                   dict: Validated input data.
           """
           sanitizedNameValueMap = {}
           for key, value in params.items():
               santitizedKey = sanitize_key(key)
               sanitizedValue = sanitize_value(sanitizeNameValuePair(sanitizedKey, value))

               # Add only valid keys to final map
               if validatedKeys.get(sanitizedKey):
                   sanitizedNameValueMap[sanitizedKey] = sanitizedValue

           return sanitizedNameValueMap

       @staticmethod
       def sanitize_key(key):
           """Removes special characters (except underscores), converts strings to lowercase."""
           return re.sub(r'[^\w_]', '', key.lower())

       @staticmethod
       def sanitize_value(value):
           '''Removes all spaces and returns empty string'''
           return '' if type(value).__bases__[0].__name__ == 'basestring' and value.replace(' ', '') == '' else value  
       ```

By addressing these points, you improve the robustness, efficiency, and maintainability of the code while enhancing its security features.

apps/common/utils/function_code.py
```diff
@@ -0,0 +1,99 @@
...
+        subprocess_result = subprocess.run(
+            ['su', '-s', python_directory, '-c', "exec(open('" + exec_python_file + "').read())", self.user],
+            text=True,
+            capture_output=True, **kwargs)
+        os.remove(exec_python_file)
+        return subprocess_result
+
+    @staticmethod
+    def _exec(_code):
+        return subprocess.run([python_directory, '-c', _code], text=True, capture_output=True)
```
shaohuzhang1, line 99: Some potential issues with the code:

  * The `_exec_sandbox` method has a typo where it tries to access `env['os']` instead of `os.environ`.
  * There is no error handling for cases where the `_exec_sandbox` function encounters an unexpected output from the Python executable.
  * It would be useful to add more detailed logging around critical sections of the code for easier debugging.
  * Some variables like `PYTHONPATH`, `LD_LIBRARY_PATH`, etc. may need to be added to the sandbox environment in order to ensure the script can execute properly.

Optimization suggestions:

  * Consider using multiprocessing instead of running each execution request in its own process, as this could potentially reduce overhead when running many requests sequentially.
  * If possible, use a precompiled version of the python interpreter within the sandbox rather than invoking the shell.
  * Use namedtuples or other immutable data structures instead of dictionaries for `locals_v` and `globals_v` because they provide better performance.
  * Consider adding support for additional language features such as async/await syntax if necessary for certain types of execution requests.

PR #3183, opened 2025-06-04
feat: Resources authorization

shaohuzhang1:
feat: Resources authorization


apps/application/serializers/application_version.py
```diff
@@ -0,0 +1,89 @@
...
+            work_flow_version = QuerySet(WorkFlowVersion).filter(application_id=self.data.get('application_id'),
+                                                                 id=self.data.get('work_flow_version_id')).first()
+            if work_flow_version is not None:
+                name = instance.get('name', None)
+                if name is not None and len(name) > 0:
+                    work_flow_version.name = name
+                work_flow_version.save()
+                return ApplicationVersionModelSerializer(work_flow_version).data
+            else:
+                raise AppApiException(500, _('Workflow version does not exist'))
```
shaohuzhang1, line 89: There aren't any major issues or performance optimizations needed in this code at present. Here's a brief review:

### Code Review

1. **Imports**: The `page_search` function from the `common.db.search` module is imported but isn't used anywhere else. It might be redundant.

2. **Serializer Class Documentation**: The docstring for each serializer class and method starts with `"##"`, which doesn't seem appropriate for Python docstrings (`""" """`).

3. **Method Signatures**: 
   - The `list` method takes two optional parameters (`with_valid`), which could potentially confuse users about their purpose.
   - The `page` method has a similar issue with its parameter description.

4. **Exception Handling**:
   - All exceptions are handled within the respective methods using `raise AppApiException`.
   - This suggests that all invalid operations should lead to an API exception, which is reasonable given the business logic involved.

Overall, the code looks well-structured and implements the required functionality for querying and editing workflow versions associated with applications. However, minor improvements can be made in documentation and handling signatures.

### Possible Improvements

1. **Docstring Formatting and Clarity**:
   - Ensure consistent docstring formatting using triple quotes (`''' ... '''`) instead of double hash symbols.

2. **Parameter Descriptions**: 
   - Clarify the use of `with_valid` parameters in both the `list` and `page` methods. Perhaps rename them to something like `validate_input`.

3. **Code Readability**:
   - Consider breaking down larger methods into smaller functions when they become too complex.

Here's a revised version of the first few lines based on these suggestions:

```python
#coding=utf-8
"""
project: MaxKB
Author：虎虎
file： application_version.py
date：2025/6/3 16:25
desc:
"""

from typing import Dict

from application.models import WorkFlowVersion
from common.db.search import page_search
from common.exception.app_exception import AppApiException
```

This adjustment makes it easier for anyone reading the code to understand what each part represents.

ui/src/views/resource-authorization/component/PermissionSetting.vue
```diff
@@ -168,32 +168,32 @@ const filterData = computed(() =>
...
       if (Name === AuthorizationEnum.MANAGE && e) {
-        item.operate[AuthorizationEnum.USE] = true
-      } else if (Name === AuthorizationEnum.USE && !e) {
-        item.operate[AuthorizationEnum.MANAGE] = false
+        item.permission[AuthorizationEnum.VIEW] = true
+      } else if (Name === AuthorizationEnum.VIEW && !e) {
+        item.permission[AuthorizationEnum.MANAGE] = false
       }
     }
   })
```
shaohuzhang1, line 199: The provided code appears to be related to managing permissions for different types of resources in an application, such as datasets or knowledge. Here are some points and potential improvements:

### Code Points Review

1. **Type Checking**:
   ```javascript
   const isDataset = computed(() => props.type === AuthorizationEnum.DATASET)
   ```
   Ensure that `AuthorizationEnum` is properly defined elsewhere in your project.

2. **Variable Naming Consistency**:
   - In template conditions (`v-if`, etc.), you use `icon` but store it in `type`. Make sure these correspond correctly.
   
3. **Permission Array Structure**:
   ```javascript
   type: String, // Assuming this should be Number for enum values
   operate: {
     type: Object,
     default() {
       return {};
     },
   },
   ```

4. **Template Updates**:
   - Corrected variable names from `operate` to `permission`.
   - Adjusted logic to handle role checks based on permission properties rather than operation properties.

5. **State Management**:
   - Ensure `manage` prop is properly passed down to control checkbox behavior when editing mode is off.

6. **Logic Flow Corrections**:
   - Fixed logic involving `checkedOperateChange` to ensure correct handling of user roles based on `view` and `manage` permissions.

7. **Validation Check**:
   Verify if there's a validation component being used somewhere since you're referring to `@/components/auto-tooltip`.

### Optimization Suggestion

- **Use More Descriptive Variable Names**:
  Instead of using generic variables like `all_indeterminate` and `all_checked`, consider more descriptive names that clearly indicate their purpose (e.g., `datasetAllChecked`, `knowledgeAllChecked`).

### Potential Improvements

1. **Refactored Logic for Role Checks**:
   The current logic around managing and usage toggling can be simplified:
   - If `Manage` is enabled, always enable `Use`.
   - Only enable `Manage` if both `View` and `Use` are checked.

2. **Code DRYing**:
   Extract common operations into helper methods to reduce duplicates.

3. **Validation Enhancements**:
   Ensure that any input validations (like ensuring that each resource has at least one permission checked) are implemented where applicable.

Overall, the code structure looks mostly consistent with expected conventions for Vue.js. However, reviewing specific contexts may uncover additional areas needing attention depending on your full implementation details.

PR #3193, opened 2025-06-04
feat: system profile

shaohuzhang1:
feat: system profile

### DIFF
--- a/apps/system_manage/serializers/system.py
+++ b/apps/system_manage/serializers/system.py
@@ -1,0 +1,41 @@
+# coding=utf-8
+"""
+    @project: MaxKB
+    @Author：虎虎
+    @file： system.py
+    @date：2025/6/4 16:01
+    @desc:
+"""
+import os
+
+from django.db import models
+from rest_framework import serializers
+from django.core.cache import cache
+
+from common.constants.cache_version import Cache_Version
+from maxkb import settings
+
+
+class SettingType(models.CharField):
+    # Community Edition
+    CE = "CE", "社区"
+    # Enterprise Edition
+    PE = "PE", "专业版"
+    # Professional Edition
+    EE = "EE", '企业版'
+
+
+class SystemProfileResponseSerializer(serializers.Serializer):
+    version = serializers.CharField(required=True, label="version")
+    edition = serializers.CharField(required=True, label="edition")
+    license_is_valid = serializers.BooleanField(required=True, label="License is valid")
+
+
+class SystemProfileSerializer(serializers.Serializer):
+    @staticmethod
+    def profile():
+        version = os.environ.get('MAXKB_VERSION')
+        license_is_valid = cache.get(Cache_Version.SYSTEM.get_key(key='license_is_valid'),
+                                     version=Cache_Version.SYSTEM.get_version())
+        return {'version': version, 'edition': settings.edition,
+                'license_is_valid': license_is_valid if license_is_valid is not None else False}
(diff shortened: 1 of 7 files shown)


apps/system_manage/serializers/system.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class SystemProfileSerializer(serializers.Serializer):
+    @staticmethod
+    def profile():
+        version = os.environ.get('MAXKB_VERSION')
+        license_is_valid = cache.get(Cache_Version.SYSTEM.get_key(key='license_is_valid'),
+                                     version=Cache_Version.SYSTEM.get_version())
+        return {'version': version, 'edition': settings.edition,
+                'license_is_valid': license_is_valid if license_is_valid is not None else False}
```
shaohuzhang1, line 41: The code looks well-formed and appears to be correctly structured for use with Django Rest Framework in a MaxKB project. There are no critical issues detected, but there are some optimizations that could be considered:

1. **Environment Variable Check**: The `os.environ.get` call should include a default value to avoid raising an error if the environment variable is not set.

2. **License Validation Cache Fetching**: The cached lookup of `license_is_valid` can optimize the response process by retrieving it once at startup rather than fetching each time from the cache during subsequent requests.

3. **Documentation Comments**: While the comments are adequate, they could be slightly more descriptive without redundancy or repetition where possible.

Here's the optimized version of the provided function and class definitions with suggested improvements:

```python
import os

from django.db import models
from django.core.cache import cache

from common.constants.cache_version import Cache_Version
from maxkb import settings


class SettingType(models.CharField):
    # Community Edition
    CE = "CE", "社区"
    # Enterprise Edition
    PE = "PE", "专业版"
    # Professional Edition
    EE = "EE", '企业版'


CACHED_LICENSE_VALID_KEY = f"system_license_validation_{settings.CACHE_PREFIX}"

def system_profile():
    """Retrieve and prepare the System Profile."""
    version = os.getenv('MAXKB_VERSION', '')  # Default empty string if MAXKB_VERSION isn't set
    
    license_is_valid_cached = cache.get(CACHED_LICENSE_VALID_KEY)
    
    if license_is_valid_cached is None:
        licensee_key = Cache_Version.SYSTEM.get_key('license_is_valid')
        licensed_value = settings.get(licensee_key).get("default")  # Assuming settings have a getter method
        cache.set(CACHED_LICENSE_VALID_KEY, licensed_value, version=Cache_Version.SYSTEM.get_version())  
    
    return {
        'version': version,
        'edition': settings.edition,
        'license_is_valid': True if (license_is_valid_cached) else False
    }

class SystemProfileSerializer(serializers.Serializer):
    versionserializer = serializers.CharField(label="version")
    editionserializer = serializers.CharField(label="edition")
    licenseserializer = serializers.BooleanField(label="License is valid")

    @staticmethod
    def profile():
        """
        Retrieve and format the System Profile data.
        
        This method uses caching to minimize repeated database queries
        and improves performance.
        """
        return system_profile()
```

### Key Changes Made:
- **Default Value for Environment Variables**: Added a default value (`'')`) to the `os.getenv` call for `MAXKB_VERSION`.
- **Caching Key for License Validity**: A unique key was generated for storing the license validity check result to prevent conflicts even if multiple instances run concurrently.
- **Optimized Return Structure in Serializer Method**:
  - The `profile()` method within `SystemProfileSerializer` now directly calls the `system_profile()` function to get the complete profile information and returns it in a straightforward format, leveraging the `versionserializer`, `editionserializer`, and `licenseserializer`.

These changes make the code both efficient and maintainable while adhering to best practices for Python development and using Django and REST framework effectively.

PR #3215, opened 2025-06-09
fix: anonymous authentication

shaohuzhang1:
fix: anonymous authentication

### DIFF
--- a/apps/common/auth/common.py
+++ b/apps/common/auth/common.py
@@ -14,13 +14,11 @@ from common.utils.rsa_util import encrypt, decrypt
 
 
 class ChatAuthentication:
-    def __init__(self, auth_type: str | None, is_auth: bool, auth_passed: bool):
-        self.is_auth = is_auth
-        self.auth_passed = auth_passed
+    def __init__(self, auth_type: str | None):
         self.auth_type = auth_type
 
     def to_dict(self):
-        return {'is_auth': self.is_auth, 'auth_passed': self.auth_passed, 'auth_type': self.auth_type}
+        return {'auth_type': self.auth_type}
 
     def to_string(self):
         return encrypt(json.dumps(self.to_dict()))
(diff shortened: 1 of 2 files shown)


apps/common/auth/common.py
```diff
@@ -14,21 +14,19 @@
...
         return encrypt(json.dumps(self.to_dict()))
 
     @staticmethod
     def new_instance(authentication: str):
         auth = json.loads(decrypt(authentication))
-        return ChatAuthentication(auth.get('auth_type'), auth.get('is_auth'), auth.get('auth_passed'))
+        return ChatAuthentication(auth.get('auth_type'))
 
 
 class ChatUserToken:
```
shaohuzhang1, line 32: The code is generally clean and efficient, but there are a few suggestions for improvement:

1. **Remove Unnecessary Variables**: The `is_auth` and `auth_passed` properties can be removed since they are not used after initialization.

   ```python
-    def __init__(self, auth_type: str | None, is_auth: bool, auth_passed: bool):
-        self.is_auth = is_auth
-        self.auth_passed = auth_passed
+    def __init__(self, auth_type: str | None):
         self.auth_type = auth_type
   ```

2. **Simplify Initialization Logic**: Since `authentication` is only used once and then processed again using `json.loads`, you can directly set the attributes during initialization.

   ```python
     @staticmethod
     def new_instance(authentication: str):
         auth = json.loads(decrypt(authentication))
-        return ChatAuthentication(auth.get('auth_type'), auth.get('is_auth'), auth.get('auth_passed'))
+        return ChatAuthentication(auth['auth_type'])
   ```

3. **Consider Adding Type Annotations**: While Python type hints can help with readability and maintainability, they do not enforce at runtime. If performance is critical or correctness matters significantly, consider adding explicit checks or constraints.

These changes simplify the code while maintaining its functionality.

PR #3216, opened 2025-06-09
fix: anonymous authentication

shaohuzhang1:
fix: anonymous authentication

### DIFF
--- a/apps/common/auth/common.py
+++ b/apps/common/auth/common.py
@@ -14,13 +14,11 @@ from common.utils.rsa_util import encrypt, decrypt
 
 
 class ChatAuthentication:
-    def __init__(self, auth_type: str | None, is_auth: bool, auth_passed: bool):
-        self.is_auth = is_auth
-        self.auth_passed = auth_passed
+    def __init__(self, auth_type: str | None):
         self.auth_type = auth_type
 
     def to_dict(self):
-        return {'is_auth': self.is_auth, 'auth_passed': self.auth_passed, 'auth_type': self.auth_type}
+        return {'auth_type': self.auth_type}
 
     def to_string(self):
         return encrypt(json.dumps(self.to_dict()))
@@ -28,7 +26,7 @@ class ChatAuthentication:
     @staticmethod
     def new_instance(authentication: str):
         auth = json.loads(decrypt(authentication))
-        return ChatAuthentication(auth.get('auth_type'), auth.get('is_auth'), auth.get('auth_passed'))
+        return ChatAuthentication(auth.get('auth_type'))
 
 
 class ChatUserToken:
--- a/apps/common/auth/handle/impl/chat_anonymous_user_token.py
+++ b/apps/common/auth/handle/impl/chat_anonymous_user_token.py
@@ -45,7 +45,8 @@ class ChatAnonymousUserToken(AuthBaseHandle):
         if application_setting_model is not None:
             application_setting = QuerySet(application_setting_model).filter(application_id=application_id).first()
             if application_setting.authentication:
-                raise AppAuthenticationFailed(1002, _('Authentication information is incorrect'))
+                if 'password' != chat_user_token.authentication.auth_type:
+                    raise AppAuthenticationFailed(1002, _('Authentication information is incorrect'))
         return None, ChatAuth(
             current_role_list=[RoleConstants.CHAT_ANONYMOUS_USER],
             permission_list=[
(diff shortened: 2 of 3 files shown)


apps/common/auth/common.py
```diff
@@ -14,21 +14,19 @@
...
         return encrypt(json.dumps(self.to_dict()))
 
     @staticmethod
     def new_instance(authentication: str):
         auth = json.loads(decrypt(authentication))
-        return ChatAuthentication(auth.get('auth_type'), auth.get('is_auth'), auth.get('auth_passed'))
+        return ChatAuthentication(auth.get('auth_type'))
 
 
 class ChatUserToken:
```
shaohuzhang1, line 32: There are several areas where the code can be improved:

1. **Consistent Initialization**: In `ChatAuthentication`, you have different initializers with varying parameters. Only accept one or two of these in future implementations. 

2. **Use Constants for Boolean Values**: For `is_auth` and `auth_passed`, define constants (e.g., `AUTH_IS_ACTIVE`) to avoid magic numbers.

3. **Avoid Implicit Decryption in Constructor**: Direct decryption within the constructor might expose sensitive data if not handled securely.

4. **Remove Redundant Parameters**: Since `is_auth` becomes redundant with `authorization_status`, remove it from the constructor and update methods accordingly.

Here’s a proposed improvement on how the class could look like after addressing these points. Note that actual encryption/decryption logic is omitted as it depends on external libraries such as AES-CBC, but this should give an idea of restructuring.

```python
@@ -14,7 +14,8 @@
class ChatAuthentication:
    AUTH_TYPE_NOT_SET = "Not Set"
    NOT_AUTHENTICATED = False
    AUTHENTICATION_SUCCESSFUL = True

    def __init__(self, auth_type: str):
        """Initialize the authentication type."""
        self.auth_type = auth_type or self.AUTH_TYPE_NOT_SET

    def to_dict(self):
        # This method doesn't need to include all fields anymore due to simplification
        return {'auth_type': self.auth_type}

    def to_string(self):
        encrypted_data = encrypt(json.dumps(self.to_dict()))
        # Encrypted result will be stringified before returning
        return encrypted_result

    @staticmethod
    def new_instance(authentication: str):
        decrypted_data = decrypt(authentication)
        auth_obj = {k.lower(): v for k,v in json.loads(decrypted_data).items()}
        return ChatAuthentication(
            auth_type=auth_obj.get('auth_type', ChatAuthentication.AUTH_TYPE_NOT_SET),
            authorization_status=auth_obj.get('authorization_status', ChatAuthentication.NOT_AUTHENTICATED)  # Assuming 'authorization_status' replaces 'is_auth'
        )
```

**Note:** Replace `encrypt` and `decrypt` functions with proper cryptographic library function calls, ensuring they manage secrets appropriately.

apps/common/auth/handle/impl/chat_anonymous_user_token.py
```diff
@@ -45,7 +45,8 @@ def handle(self, request, token: str, get_token_details):
         if application_setting_model is not None:
             application_setting = QuerySet(application_setting_model).filter(application_id=application_id).first()
             if application_setting.authentication:
-                raise AppAuthenticationFailed(1002, _('Authentication information is incorrect'))
+                if 'password' != chat_user_token.authentication.auth_type:
+                    raise AppAuthenticationFailed(1002, _('Authentication information is incorrect'))
         return None, ChatAuth(
             current_role_list=[RoleConstants.CHAT_ANONYMOUS_USER],
             permission_list=[
```
shaohuzhang1, line 52: There's an inconsistency in checking the `auth_type`. If the authentication type is neither `'password'` nor anything else specified, it should be explicitly checked within the `AuthenticationFailed` exception message to prevent incorrect error messages. Here’s an updated version with proper logic:

```python
@@ -45,7 +45,8 @@ def handle(self, request, token: str, get_token_details):
         if application_setting_model is not None:
             application_setting = QuerySet(application_setting_model).filter(application_id=application_id).first()
             if application_setting.authentication:
-                raise AppAuthenticationFailed(1002, _('Authentication information is incorrect'))
+                auth_types_allowed = ['custom', 'oauth']  # Add the allowed types here
+                if chat_user_token.authentication.auth_type not in auth_types_allowed:
+                    raise AppAuthenticationFailed(1002,
+                                                 _("Unknown or unsupported authentication method"))
```

This ensures that only known and supported authentication methods trigger the exception, instead of being overly generic (e.g., "Authentication information is incorrect").

PR #3225, opened 2025-06-09
feat: application stats

shaohuzhang1:
feat: application stats

### DIFF
--- a/apps/application/serializers/application_stats.py
+++ b/apps/application/serializers/application_stats.py
@@ -1,0 +1,106 @@
+# coding=utf-8
+"""
+    @project: MaxKB
+    @Author：虎虎
+    @file： application_stats.py
+    @date：2025/6/9 20:34
+    @desc:
+"""
+import datetime
+import os
+from typing import Dict, List
+
+from django.db import models
+from django.db.models import QuerySet
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
+
+from application.models import ApplicationChatUserStats
+from common.db.search import native_search, get_dynamics_model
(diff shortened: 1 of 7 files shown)


apps/application/serializers/application_stats.py
```diff
@@ -0,0 +1,106 @@
...
+    @staticmethod
+    def get_days_between_dates(start_date, end_date):
+        start_date = datetime.datetime.strptime(start_date, '%Y-%m-%d')
+        end_date = datetime.datetime.strptime(end_date, '%Y-%m-%d')
+        days = []
+        current_date = start_date
+        while current_date <= end_date:
+            days.append(current_date.strftime('%Y-%m-%d'))
+            current_date += datetime.timedelta(days=1)
+        return days
```
shaohuzhang1, line 106: 1. The file's encoding is not set correctly (`# coding=utf-8`) but UTF-8 is recommended.

2. There are some redundant imports that can be removed.

3. `get_end_time` and `get_start_time` methods use the same logic, which seems like they should differ only in formatting. Consider using one method instead.

4. In `merge_customer_chat_record`, `find` function is called twice to retrieve data for each metric (star_num, trample_num, etc.). This can be optimized by calling it once to fetch all metrics at once for better performance.

5. Ensure that all required fields in both Serializers are present and validate.

6. Use more descriptive variable names within functions like `get_days_between_dates`.

7. Make sure that Django templates or other parts of the system handle UUID strings properly when passed from this view layer.

These are general improvements rather than specific fixes to errors; if you encounter actual bugs while implementing these changes, please let me know so I can assist further!

PR #3749, opened 2025-07-25
fix: Model swagger

shaohuzhang1:
fix: Model swagger

### DIFF
--- a/apps/common/utils/common.py
+++ b/apps/common/utils/common.py
@@ -334,3 +334,6 @@ def parse_image(content: str):
 
 def generate_uuid(tag: str):
     return str(uuid.uuid5(uuid.NAMESPACE_DNS, tag))
+
+def filter_workspace(query_list):
+    return [q for q in query_list if q.name!="workspace_id"]
--- a/apps/models_provider/api/model.py
+++ b/apps/models_provider/api/model.py
@@ -108,6 +108,9 @@ class GetModelApi(APIMixin):
             required=True,
         )
         ]
+    @staticmethod
+    def get_request():
+        return []
 
     @staticmethod
     def get_response():


apps/common/utils/common.py
```diff
@@ -334,3 +334,6 @@ def parse_image(content: str):
 
 def generate_uuid(tag: str):
     return str(uuid.uuid5(uuid.NAMESPACE_DNS, tag))
+
+def filter_workspace(query_list):
+    return [q for q in query_list if q.name!="workspace_id"]
\ No newline at end of file
```
shaohuzhang1, line 339: 1) The current code is complete and free from significant irregularities or issues as written.

2) There are no immediate areas needing optimization in this function. 

3) If more specific optimizations were needed (e.g., using list comprehensions for efficiency), the `filter_workspace` function could be rewritten to:

   ```python
   def filter_workspace(query_list): 
       return [q for q in query_list if q.name != "workspace_id"]
   ```
   
This version uses Python's built-in capabilities efficiently. Note that you may want to add type hints (`-> List[YourObjectType]`) depending on how 'query_list' should be used elsewhere in your program.

apps/models_provider/api/model.py
```diff
@@ -108,6 +108,9 @@ def get_query_params_api():
             required=True,
         )
         ]
+    @staticmethod
+    def get_request():
+        return []
 
     @staticmethod
     def get_response():
```
shaohuzhang1, line 116: The provided code snippet has a few issues and can be optimized:

1. **Static Method Definition**: The `get_request` method is defined with `@staticmethod`, but it appears that you intended to define regular methods (`def`) instead, as static methods do not have access to instance variables unless they are passed explicitly.

2. **Response Method Missing**: There is no `get_response` method at all, which contradicts the comments suggesting an empty list for this method. Assuming you want to implement a response generation logic, consider adding one.

3. **Missing Import Statements**: It's common to see import statements near the top of Python files when using classes. Ensure these are included if necessary.

4. **Docstring Consistency**: While there are docstrings explaining parameters, some functions lack them entirely or contain incorrect information. Add docstrings for clarity and maintainability.

5. **Parameter Documentation in Comments**: In the `get_query_params_api` function, the parameter descriptions are embedded directly within the comment block rather than being part of the function signature. Consider keeping everything consistent.

Here’s an optimized version with improvements suggested:

```python
class MyAPI:
    """
    Base class for API configurations.
    """

    @classmethod
    def get_query_params_api(cls):
        """
        Fetches query parameters specific to the API.

        :return: List of dictionaries containing query parameter names and their descriptions.
                 Example:
                       [
                           {'name': 'param1', 'description': 'Description of param1'},
                           ...
                       ]
        """
        return [
            {
                'name': 'param',
                'type': str,
                'required': True,
                "default": ""
            }
        ]

    @staticmethod
    def get_request(request_data=None):
        """
        Generates request JSON given input data.

        :param request_data: Dictionary containing relevant fields from the form submission.
        :return: JSON string representing the request payload.
        """
        # Placeholder implementation; replace with actual request logic
        if request_data:
            return {
                key: value
                for key, value in request_data.items()
                if key in ('field_name')
            }

        return {}

    @staticmethod
    def get_response():
        """
        Generates a JSON structure with sample results based on user inputs.

        :param response_data: Dictionary containing relevant results from the database or calculation.
        :return: JSON string representing the response payload.
        """
        # Placeholder implementation; replace with actual response logic
        if isinstance(response_data, dict):
            return {
                'status': 'success',
                'data': response_data.get('result_key', {})
            }
        
        return {}
```

This version maintains good consistency across the class, includes clear documentation, adds missing functionality like generating requests and responses, and ensures proper use of static vs. non-static methods and docstrings.

PR #3770, opened 2025-07-29
fix: The conversation page cannot be opened

shaohuzhang1:
fix: The conversation page cannot be opened

### DIFF
--- a/ui/src/stores/modules/application.ts
+++ b/ui/src/stores/modules/application.ts
@@ -3,7 +3,7 @@ import applicationApi from '@/api/application/application'
 import { type Ref } from 'vue'
 const useApplicationStore = defineStore('application', {
   state: () => ({
-    location: `${window.location.origin}${window.MaxKB.chatPrefix}/`,
+    location: `${window.location.origin}${window.MaxKB.chatPrefix ? window.MaxKB.chatPrefix : window.MaxKB.prefix}/`,
   }),
   actions: {
     async asyncGetApplicationDetail(id: string, loading?: Ref<boolean>) {


ui/src/stores/modules/application.ts
```diff
@@ -3,7 +3,7 @@ import applicationApi from '@/api/application/application'
 import { type Ref } from 'vue'
 const useApplicationStore = defineStore('application', {
   state: () => ({
-    location: `${window.location.origin}${window.MaxKB.chatPrefix}/`,
+    location: `${window.location.origin}${window.MaxKB.chatPrefix ? window.MaxKB.chatPrefix : window.MaxKB.prefix}/`,
   }),
   actions: {
     async asyncGetApplicationDetail(id: string, loading?: Ref<boolean>) {
```
shaohuzhang1, line 9: The provided code appears to have some minor adjustments:

1. **Condition check in URL construction**: The original code uses `window.MaxKB.chatPrefix` directly without checking if it exists before concatenating with the origin and prefix. This can lead to an error if `chatPrefix` is not defined.

2. **Parameter naming consistency**: In the action method, you mentioned using a variable named `id`, but then used `loading? Ref<boolean>` which seems like TypeScript syntax for optionals, possibly intended to indicate whether a loading state is present.

### Optimization Suggestion:
To improve robustness and maintainability, consider adding a default value to ensure that both `chatPrefix` and `prefix` are used if `chatPrefix` is indeed undefined. You could also extract this logic into a utility function or property if necessary.

Here's how you might structure the changes:

```typescript
import appApi from '@/api/application/application'
import type { Ref } from 'vue'
const useApplicationStore = defineStore('application', {
   state: () => ({
     location: `${window.location.origin && window.MaxKB.chatPrefix}`
       ? `${window.location.origin}${window.MaxKB.chatPrefix}/`
       : `${window.location.origin}${window.MaxKB.prefix || ''}/`,
   }),
   actions: {
     async asyncGetApplicationDetail(id: string, loading?: Ref<boolean>) {
     // Your existing code here
     }
   },
})
```

By ensuring these checks and consolidations, your component should be more resilient and easier to understand.

Issue #3671, opened 2025-07-18
[Bug] [Model] The model parameter setting is, open the model parameters. If there are multiple parameters, open the parameter editing and the content inside will become the first parameter.


cjc1108:
### MaxKB Version

v2.0.0

### Problem Description

打开两个参数，都是第一个温度的相关信息
<img width="1035" height="453" alt="Image" src="https://github.com/user-attachments/assets/bcb6d222-d3c1-4e74-9805-c1effcd83d67" />

<img width="1035" height="634" alt="Image" src="https://github.com/user-attachments/assets/c81b4cfc-30e8-470e-a09a-8b2e93bbc118" />

<img width="1035" height="447" alt="Image" src="https://github.com/user-attachments/assets/ee7b4f0f-2120-4e05-ab29-f3c9d5253177" />

<img width="1035" height="634" alt="Image" src="https://github.com/user-attachments/assets/c0b90437-1796-4e2f-8fac-7061aee71cc0" />
打开三个参数，都是第一个size的信息
<img width="2229" height="1212" alt="Image" src="https://github.com/user-attachments/assets/cd993484-577f-4f11-aef9-3bc3b78e8864" />

### Steps to Reproduce

创建模型——查看参数设置

### Related log output

```shell

```

youliyuan-fit2cloud:
Known issue，will be fixed next version.

Shenguobin0102:
已修复，请关注下个版本。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Fixed, please follow the next version.

PR #3779, opened 2025-07-30
fix: The model parameter setting is, open the model parameters. If there are multiple parameters, open the parameter editing and the content inside will become the first parameter. #3671

shaohuzhang1:
fix: The model parameter setting is, open the model parameters. If there are multiple parameters, open the parameter editing and the content inside will become the first parameter. #3671

### DIFF
--- a/ui/src/views/model/component/AddParamDrawer.vue
+++ b/ui/src/views/model/component/AddParamDrawer.vue
@@ -7,6 +7,7 @@
     :close-on-click-modal="false"
     :close-on-press-escape="false"
     :before-close="cancelClick"
+    destroy-on-close
   >
     <template #header>
       <h4>


ui/src/views/model/component/AddParamDrawer.vue
```diff
@@ -7,6 +7,7 @@
     :close-on-click-modal="false"
     :close-on-press-escape="false"
     :before-close="cancelClick"
+    destroy-on-close
   >
     <template #header>
       <h4>
```
shaohuzhang1, line 13: There is an unnecessary semicolon at the end of the line `>`. Additionally, there should be spaces between elements within inline attributes:

```html
<el-dialog title="..." append-to-body 
          close-on-click-modal="false" 
          close-on-press-escape="false"
          before-close="cancelClick"
          :destroy-on-close=true>
```

This makes the HTML more readable and corrects the syntax error.

PR #3788, opened 2025-07-31
fix: Permission constants

shaohuzhang1:
fix: Permission constants

### DIFF
--- a/apps/common/constants/permission_constants.py
+++ b/apps/common/constants/permission_constants.py
@@ -1212,10 +1212,6 @@ class PermissionConstants(Enum):
         group=Group.SYSTEM_RES_APPLICATION, operate=Operate.DEBUG, role_list=[RoleConstants.ADMIN],
         parent_group=[SystemGroup.RESOURCE_APPLICATION]
     )
-    RESOURCE_APPLICATION_IMPORT = Permission(
-        group=Group.SYSTEM_RES_APPLICATION, operate=Operate.IMPORT, role_list=[RoleConstants.ADMIN],
-        parent_group=[SystemGroup.RESOURCE_APPLICATION]
-    )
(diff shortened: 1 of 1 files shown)


apps/common/constants/permission_constants.py
```diff
@@ -1212,10 +1212,6 @@ class PermissionConstants(Enum):
         group=Group.SYSTEM_RES_APPLICATION, operate=Operate.DEBUG, role_list=[RoleConstants.ADMIN],
         parent_group=[SystemGroup.RESOURCE_APPLICATION]
     )
-    RESOURCE_APPLICATION_IMPORT = Permission(
-        group=Group.SYSTEM_RES_APPLICATION, operate=Operate.IMPORT, role_list=[RoleConstants.ADMIN],
-        parent_group=[SystemGroup.RESOURCE_APPLICATION]
-    )
     RESOURCE_APPLICATION_EXPORT = Permission(
         group=Group.SYSTEM_RES_APPLICATION, operate=Operate.EXPORT, role_list=[RoleConstants.ADMIN],
         parent_group=[SystemGroup.RESOURCE_APPLICATION]
```
shaohuzhang1, line 1217: There seems to be an issue with the duplication of the `GROUP` field between "Import" and "Export" permissions in the same permission class. The current implementation includes both fields redundantly for each entry (e.g., `group`, `operate`, etc.). To fix this, only include unique groups once:

```python
@@ -1213,6 +1212,6 @@ class PermissionConstants(Enum):
     SYS_APP_DEBUG = Permission(
         operate=Operate.DEBUG, role_list=[RoleAdmin],
         parent_group=[SystemGroup.RESOURCE_APPLICATION]
-    )
```

This will clean up redundancy while maintaining consistent data structures within your codebase.

PR #3791, opened 2025-08-01
fix: Problem related permission

shaohuzhang1:
fix: Problem related permission

### DIFF
--- a/ui/src/views/paragraph/component/ProblemComponent.vue
+++ b/ui/src/views/paragraph/component/ProblemComponent.vue
@@ -3,7 +3,9 @@
     <span class="flex align-center">
       <span>{{ $t('views.paragraph.relatedProblem.title') }}</span>
       <el-divider direction="vertical" class="mr-4" />
-      <el-button text @click="addProblem">
+      <el-button text @click="addProblem"
+        v-if="permissionPrecise.problem_relate(id)"
+      >
         <el-icon><Plus /></el-icon>
       </el-button>
     </span>
@@ -43,7 +45,7 @@
             class="question-tag"
             type="info"
             effect="plain"
-            closable
+            v-bind="permissionPrecise.problem_relate(id) ? {closable:true} : {} "
           >
             <auto-tooltip :content="item.content">
               {{ item.content }}
@@ -55,9 +57,11 @@
   </div>
 </template>
 <script setup lang="ts">
-import { ref, nextTick, onMounted, onUnmounted, watch } from 'vue'
+import { ref, nextTick, onMounted, onUnmounted, watch, computed } from 'vue'
 import { useRoute } from 'vue-router'
 import { loadSharedApi } from '@/utils/dynamics-api/shared-api'
+import permissionMap from '@/permission'
+
 
 const props = defineProps<{
   paragraphId: String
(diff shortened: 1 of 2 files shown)


ui/src/views/paragraph/component/ProblemComponent.vue
```diff
@@ -71,6 +75,10 @@ const {
   params: { id, documentId }, // id为knowledgeId
 } = route as any
 
+const permissionPrecise = computed(() => {
+  return permissionMap['knowledge'][props.apiType]
+})
+
 const inputRef = ref()
 const loading = ref(false)
 const isAddProblem = ref(false)
```
shaohuzhang1, line 84: The provided TypeScript code looks generally well-structured and follows Vue's syntax conventions. However, there are a couple of potential improvements that could be made:

1. **Use `computed` Properly**: The current implementation of `permissionPrecise` uses computed properties to determine if the button should be disabled based on permissions. This approach seems correct when used with other computed properties or directives. Ensure it does not cause unnecessary recomputations.

2. **Ensure Proper Permissions Check**: It's important to verify that the `permissionMap['knowledge'][props.apiType]` is initialized correctly and contains all necessary keys like `'problem_relate'`. If these permissions might change at any time during runtime, consider moving this logic into a method or event handler instead of directly within a template expression.

3. **Optimize Template Interpolation**: In your loop (`v-for`), you're using `v-bind="{ closable: true }"` for conditional classes. Consider using more concise expressions wherever possible. For example:
   ```html
   <template #icon>
     <el-icon :class="[permissionPrecise.problem_relate(id) ? 'close-btn-active' : '']"><Plus /></el-icon>
   </template>
   ```

4. **Avoid Inline Styles in Templates**: Although inline styles can enhance performance and readability in small snippets, they should be minimized due to their verbosity and lack of reusability. Instead, style them through CSS classes.

### Example of Refactor Suggestions

```typescript
// Add a method to compute closability conditionally
const getIsCloseable = () => permissionPrecise.value && permissionPrecise.value.problem_relate(props.id);

<template #icon>
  <el-icon :class="[getIsCloseable() ? 'close-btn-active' : '']"><Plus /></el-icon>
</template>

<style scoped>
.close-btn-active {
  /* Styling for active close icon */
}
</style>
```

These changes aim to improve maintainability and potentially optimize the code performance by reducing inline logic and promoting clean separation of concerns between JavaScript and templates.

ui/src/views/problem/component/DetailProblemDrawer.vue
```diff
@@ -34,7 +34,7 @@
                         type="primary"
                         text
                         @click.stop="disassociation(item)"
-                        v-if="permissionPrecise.doc_edit(id as string)"
+                        v-if="permissionPrecise.problem_relate(id as string)"
                       >
                         <AppIcon iconName="app-quxiaoguanlian"></AppIcon>
                       </el-button>
```
shaohuzhang1, line 40: The line `v-if="permissionPrecise.doc_edit(id as string)"` should be changed to `v-if="permissionPrecise.problem_relate(id as string)"`
This change will ensure that the button is only displayed when an association can be created for "problem" rather than "document".

PR #3800, opened 2025-08-04
fix: After the application name change - the application name on the demo page has not been updated

shaohuzhang1:
fix: After the application name change - the application name on the demo page has not been updated

### DIFF
--- a/apps/application/serializers/application.py
+++ b/apps/application/serializers/application.py
@@ -25,11 +25,13 @@ from django.utils.translation import gettext_lazy as _
 from langchain_mcp_adapters.client import MultiServerMCPClient
 from rest_framework import serializers, status
 from rest_framework.utils.formatting import lazy_format
+
 from application.flow.common import Workflow
 from application.models.application import Application, ApplicationTypeChoices, ApplicationKnowledgeMapping, \
     ApplicationFolder, ApplicationVersion
 from application.models.application_access_token import ApplicationAccessToken
 from common import result
+from common.cache_data.application_access_token_cache import del_application_access_token
 from common.database_model_manage.database_model_manage import DatabaseModelManage
 from common.db.search import native_search, native_page_search
 from common.exception.app_exception import AppApiException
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application.py
```diff
@@ -734,6 +736,18 @@ def publish(self, instance, with_valid=True):
...
+        if application_access_token is None:
+            application_access_token = ApplicationAccessToken(application_id=application.id,
+                                                              access_token=access_token, is_active=True)
+            application_access_token.save()
+        else:
+            access_token = application_access_token.access_token
+        del_application_access_token(access_token)
         return self.one(with_valid=False)
 
     @staticmethod
```
shaohuzhang1, line 753: There are a few areas to consider for improvement:

1. **Code Duplication**:
   ```python
   from common.cache_data.application_access_token_cache import del_application_access_token
   ```
   This line can be moved out of the function, so it's only called once.

2. **Redundant Code**: The following lines:
   ```python
   access_token = hashlib.md5(str(uuid.uuid7()).encode()).hexdigest()[8:24]
   application_access_token = QuerySet(ApplicationAccessToken).filter(application_id=application.id).first()
   if application_access_token is None:
       application_access_token = ApplicationAccessToken(application_id=application.id,
                                                         access_token=access_token, is_active=True)
       application_access_token.save()
   else:
       access_token = application_access_token.access_token
   ```

   could be combined into one more concise statement using `get_or_create` method:
   ```python
   application_access_token, _ = ApplicationAccessToken.objects.get_or_create(
       application_id=application.id,
       defaults={'access_token': hashlib.md5(str(uuid.uuid7()).encode()).hexdigest()[8:24], 'is_active': True}
   )
   ```

3. **Avoid Unused Imports**:
   Ensure that imports you're not actually using should be removed from your file.

4. **Documentation Comments**:
   Add comments where necessary to describe complex blocks of code or logic flows, which will make maintenance easier down the road.

Following these recommendations leads to cleaner and potentially more efficient code like this:

```python
from langchain_mcp_adapters.client import MultiServerMCPClient
from rest_framework import serializers, status
from rest_framework.utils.formatting import lazy_format

from application.flow.common import Workflow
from application.models.application import (
    Application,
    ApplicationTypeChoices,
    ApplicationKnowledgeMapping,
    ApplicationFolder,
    ApplicationVersion
)
from application.models.application_access_token import ApplicationAccessToken
from common import result
from common.database_model_manage import DatabaseModelManage
from common.db.search import native_search, native_page_search
from common.exception.app_exception import AppApiException

class YourClass:
    # Function implementation here...
    
    def publish(self, instance, with_valid=True):
        # Function body remains largely the same...

    @staticmethod
    def get_or_update_app_access_token(application):
        return ApplicationAccessToken.objects.get_or_create(
            application_id=application.id,
            defaults={
                'access_token': hashlib.md5(str(uuid.uuid7()).encode()).hexdigest()[8:24],
                'is_active': True
            }
        )

# Call within your method as needed:
token, created = get_or_update_app_access_token(instance.application)
```

By applying some of these suggestions, your codebase becomes leaner while maintaining readability and functionality integrity.

Issue #3754, opened 2025-07-28
[Bug] [Advanced Orchestration] The content in the specified reply was repeated several times when the Q&A was output.


2lipan:
### MaxKB Version

v1.10.9-lts

### Problem Description

【高级编排】指定回复里的内容，问答输出时重复了几遍。

### Steps to Reproduce

【高级编排】指定回复里的内容，问答输出时重复了几遍。

<img width="1913" height="956" alt="Image" src="https://github.com/user-attachments/assets/f05dad87-218e-4b7d-a96c-23f8b89f701f" />

<img width="1639" height="815" alt="Image" src="https://github.com/user-attachments/assets/b7818933-47a5-4699-bf34-ca379777654e" />

高级编排的场景如下:
[财务共享V1.0多模态模型 (1).zip](https://github.com/user-attachments/files/21459386/V1.0.1.zip)

### Related log output

```shell

```

Shenguobin0102:
你好，出问题的这个节点和前面的判断器连接了三次，请删除三个重复连线后，重新连线，会恢复正常。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Hello, the node where the problem is connected to the previous judge three times. Please delete the three repeated connections and reconnect them, and it will return to normal.

Shenguobin0102:
问题已修复，请关注新版本。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
The issue has been fixed, please pay attention to the new version.

PR #3801, opened 2025-08-04
fix: [Advanced Orchestration] The content in the specified reply was repeated several times when the Q&A was output. #3754

shaohuzhang1:
fix: [Advanced Orchestration] The content in the specified reply was repeated several times when the Q&A was output. #3754

### DIFF
--- a/ui/src/workflow/common/app-node.ts
+++ b/ui/src/workflow/common/app-node.ts
@@ -372,7 +372,9 @@ class AppNodeModel extends HtmlResize.model {
     this.sourceRules.push({
       message: t('views.applicationWorkflow.tip.notRecyclable'),
       validate: (sourceNode: any, targetNode: any, sourceAnchor: any, targetAnchor: any) => {
-        return !isLoop(sourceNode.id, targetNode.id)
+        const up_node_list = this.graphModel.getNodeIncomingNode(targetNode.id)
+        const is_c = up_node_list.find((up_node) => up_node.id == sourceNode.id)
+        return !is_c && !isLoop(sourceNode.id, targetNode.id)
       },
     })
 


ui/src/workflow/common/app-node.ts
```diff
@@ -372,7 +372,9 @@ class AppNodeModel extends HtmlResize.model {
     this.sourceRules.push({
       message: t('views.applicationWorkflow.tip.notRecyclable'),
       validate: (sourceNode: any, targetNode: any, sourceAnchor: any, targetAnchor: any) => {
-        return !isLoop(sourceNode.id, targetNode.id)
+        const up_node_list = this.graphModel.getNodeIncomingNode(targetNode.id)
+        const is_c = up_node_list.find((up_node) => up_node.id == sourceNode.id)
+        return !is_c && !isLoop(sourceNode.id, targetNode.id)
       },
     })
```
shaohuzhang1, line 380: The provided code has a minor issue where it checks if there's a loop between nodes before determining if one node can be placed in another without causing recursion. This seems to unnecessarily double-check the non-loop condition. 

Here is an improved version of the function:

```javascript
validate: (
  sourceNode: any,
  targetNode: any,
  sourceAnchor: any,
  targetAnchor: any
) => {
  const up_nodes = this.graphModel.getNodeIncomingNodes(targetNode.id);
  const is_c_or_loop = up_nodes.some(up_node => up_node.id === sourceNode.id || isLoop(sourceNode.id, targetNode.id));
  return !is_c_or_loop;
}
```

By combining the two conditions into a single `some` call using logical OR (`||`), we avoid redundant checking. If either condition (`up_node.id === sourceNode.id` or `isLoop`) evaluates to true, the function returns false, meaning the placement is invalid due to recursion. Otherwise, it returns true.

PR #3813, opened 2025-08-05
fix: refactor delete method in DocumentSerializers for improved clarity

shaohuzhang1:
fix: refactor delete method in DocumentSerializers for improved clarity  --bug=1060005 --user=刘瑞斌 【资源管理】知识库-删除文档报错 https://www.tapd.cn/62980211/s/1749123

### DIFF
--- a/apps/knowledge/serializers/document.py
+++ b/apps/knowledge/serializers/document.py
@@ -662,6 +662,7 @@ class DocumentSerializers(serializers.Serializer):
 
         @transaction.atomic
         def delete(self):
+            self.is_valid(raise_exception=True)
             document_id = self.data.get("document_id")
             source_file_ids = [
                 doc['meta'].get(


apps/knowledge/serializers/document.py
```diff
@@ -662,6 +662,7 @@ def cancel(self, instance, with_valid=True):
 
         @transaction.atomic
         def delete(self):
+            self.is_valid(raise_exception=True)
             document_id = self.data.get("document_id")
             source_file_ids = [
                 doc['meta'].get(
```
shaohuzhang1, line 668: Your code checks if `self.is_valid` raises an exception during the deletion process. However, this might not be ideal because raising exceptions could be cumbersome and potentially interrupt operations elsewhere.

### Potential Issues:
1. **Exception Handling**: Raising exceptions in deletion can lead to unpredictable behavior if it disrupts other business logic.
2. **Resource Usage**: If you need validation without interruptions, consider using a separate method or returning status codes instead of throwing exceptions.
3. **Logging**: Ensure that any errors are logged clearly, helping with debugging and system monitoring.

### Optimization Suggestions:
1. **Simplified Validation**: You could simplify the call to `self.is_valid(raise_exception=True)` by directly checking the result. This is less error-prone but requires careful handling of valid states.
   
   ```python
   from django.core.exceptions import ValidationError

   def cancel(self, instance, with_valid=True):

       @transaction.atomic
       def delete(self):
           try:
               # Validate without interrupting normal flow
               self.is_valid()
           except ValidationError as e:
               raise Exception("Deletion failed due to: {}".format(e)) from None
      
           document_id = self.data.get("document_id")
           source_file_ids = [
                   doc['meta'].get(
                       'source_file_id',
                       default=[],
                   )
                   for doc in (json.loads(document_content) or [])
               ]
           ...  # Rest of the function implementation
   ```

2. **Separate Method**: Consider creating a dedicated method that handles validation and returns an appropriate response. While this won't prevent exceptions being raised, it can improve readability and separation of concerns.

   ```python
   def cancel(self, instance, with_valid=True):
       try:
           if with_valid:
               self.validate()
           ...
       except Exception as e:
           return {"error": str(e)}
       
       @transaction.atomic
       def delete():
           # Directly handle the data retrieval and processing here
           ...

   def validate(self):
       self.is_valid()

   def process_data(self):
       document_content = request.POST.get('content')
       ...  # Process the JSON content
   ```

Overall, while your current approach works, integrating more robust validation strategies would make your code cleaner, more maintainable, and less prone to runtime issues.

PR #3816, opened 2025-08-05
fix: When the workflow session variable is not added, the dropdown option does not display the session variable attribute

shaohuzhang1:
fix: When the workflow session variable is not added, the dropdown option does not display the session variable attribute

### DIFF
--- a/ui/src/workflow/common/NodeCascader.vue
+++ b/ui/src/workflow/common/NodeCascader.vue
@@ -53,8 +53,16 @@ function visibleChange(bool: boolean) {
     options.value = props.global
       ? props.nodeModel
           .get_up_node_field_list(false, true)
-          .filter((v: any) => ['global', 'chat'].includes(v.value))
-      : props.nodeModel.get_up_node_field_list(false, true)
+          .map((v: any) => {
+            console.log(v)
+            return v
+          })
+          .filter(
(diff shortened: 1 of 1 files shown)


ui/src/workflow/common/NodeCascader.vue
```diff
@@ -53,8 +53,16 @@ function visibleChange(bool: boolean) {
...
+            return v
+          })
+          .filter(
+            (v: any) => ['global', 'chat'].includes(v.value) && v.children && v.children.length > 0,
+          )
+      : props.nodeModel
+          .get_up_node_field_list(false, true)
+          .filter((v: any) => v.children && v.children.length > 0)
   }
 }
```
shaohuzhang1, line 68: There are two main issues:

  1. The `if` statement condition is not properly formatted, making it difficult to read.
  2. The filter condition inside the second `map()` call includes an unnecessary condition of checking if `v.children` exists.

Here's the corrected code with these issues fixed:
```
function visibleChange(bool: boolean) {
  const options = props.global
    ? props.nodeModel.get_up_node_field_list(false, true)
        .filter((v: any) =>
          ['global', 'chat'].includes(v.value) &&
          v.children &&
          v.children.length > 0
        )
    : props.nodeModel.get_up_node_field_list(false, true);
}
```

PR #3818, opened 2025-08-05
fix: Node copied in orchestration, default name incorrect, should be in numerical calculation form of+1, not string form of+1

shaohuzhang1:
fix: Node copied in orchestration, default name incorrect, should be in numerical calculation form of+1, not string form of+1

### DIFF
--- a/ui/src/workflow/common/NodeContainer.vue
+++ b/ui/src/workflow/common/NodeContainer.vue
@@ -244,9 +244,9 @@ const editName = async (formEl: FormInstance | undefined) => {
   await formEl.validate((valid) => {
     if (valid) {
       if (
-        !props.nodeModel.graphModel.nodes?.some(
-          (node: any) => node.properties.stepName === form.value.title,
-        )
+        !props.nodeModel.graphModel.nodes
+          .filter((node: any) => node.id !== props.nodeModel.id)
+          ?.some((node: any) => node.properties.stepName === form.value.title)
       ) {
         set(props.nodeModel.properties, 'stepName', form.value.title)
         nodeNameDialogVisible.value = false
--- a/ui/src/workflow/common/app-node.ts
+++ b/ui/src/workflow/common/app-node.ts
@@ -12,6 +12,20 @@ import { nodeDict } from '@/workflow/common/data'
 import { isActive, connect, disconnect } from './teleport'
 import { t } from '@/locales'
 import { type Dict } from '@/api/type/common'
+const getNodeName = (nodes: Array<any>, baseName: string) => {
+  let index = 0
+  let name = baseName
+  while (true) {
+    if (index > 0) {
+      name = baseName + index
+      console.log(name)
+    }
+    if (!nodes.some((node: any) => node.properties.stepName === name.trim())) {
+      return name
+    }
+    index++
+  }
+}
 class AppNode extends HtmlResize.view {
   isMounted
   r?: any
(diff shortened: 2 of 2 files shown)


ui/src/workflow/common/app-node.ts
```diff
@@ -32,23 +46,14 @@ class AppNode extends HtmlResize.view {
...
-      if (!props.graphModel.nodes?.some((node: any) => node.properties.stepName === name.trim())) {
-        props.model.properties.stepName = name
-      } else {
-        getNodesName(number + 1)
-      }
     }
+
     props.model.properties.config = nodeDict[props.model.type].properties.config
     if (props.model.properties.height) {
       props.model.height = props.model.properties.height
```
shaohuzhang1, line 59: ### Code Review and Suggestions:

```typescript
// Import statements
import { nodeDict } from '@/workflow/common/data';
import { isActive, connect, disconnect } from './teleport';
import { type Dict } from '@/api/type/common';

// Function to generate unique step names
const getNodeName = (nodes: Array<any>, baseName: string): string => {
  let index = 0;
  let newName = baseName;

  while (true) {
    if (newName.trim() in nodes.map(node => node.properties.stepName)) {
      index++;
      newName = `${baseName}${index}`;
    } else {
      return newName;
    }
  }
};

  // Existing fields and methods

  // Update method to handle step name uniqueness
  update(props: any) {
    super.update(props);

    if (props.model.properties.noRender) {
    } else {
      const filterNodes = props.graphModel.nodes.filter((node: any) => node.type === props.model.type);
      
      // Generate a unique step name if it already exists
      const currentStepName = props.model.properties.stepName || '';
      if (filterNodes.some(node => node.properties.stepName.toLowerCase() === currentStepName.trim().toLowerCase())) {
        const uniqueStepName = getNodeName(
          props.graphModel.nodes.filter(n => n.id !== props.model.id),
          currentStepName
        );
        if (uniqueStepName) {
          props.model.properties.stepName = uniqueStepName;
        }
      }

      props.model.properties.config = nodeDict[props.model.type].properties.config;
      if (props.model.properties.height && !isNaN(Number(props.model.properties.height))) {
        props.model.height = Number.parseInt(props.model.properties.height, 10);
      }
    }
  }
}

export default AppNode;
```

### Potential Improvements:

1. **Avoid Using `console.log`**: While `console.log` can be helpful during development, it should be used judiciously and removed before production. Ensure that logs are not present in final output.

2. **Use `.includes` Instead of `map`.filter`: The use of `map(nodes).filter(...)` creates an intermediate array when filtering nodes with property matches. Use `.includes()` directly on the array returned by `filter(...)` which is more efficient and readable.

3. **Consistency in Variable Names**: Use consistent naming conventions throughout the codebase and avoid inconsistent case usage within the same variable.

4. **Edge Case Handling**: Consider adding additional checks for edge cases, such as empty arrays or properties like null values affecting the step name generation logic.

5. **Optimize Regular Expression Usage**: If necessary, consider using regular expressions sparingly to improve performance, especially when dealing with large datasets. In this context, it seems unnecessary given the approach taken.

These improvements aim to make the code cleaner, more performant, and easier to maintain.

ui/src/workflow/common/NodeContainer.vue
```diff
@@ -244,9 +244,9 @@ const editName = async (formEl: FormInstance | undefined) => {
...
       if (
-        !props.nodeModel.graphModel.nodes?.some(
-          (node: any) => node.properties.stepName === form.value.title,
-        )
+        !props.nodeModel.graphModel.nodes
+          .filter((node: any) => node.id !== props.nodeModel.id)
+          ?.some((node: any) => node.properties.stepName === form.value.title)
       ) {
         set(props.nodeModel.properties, 'stepName', form.value.title)
         nodeNameDialogVisible.value = false
```
shaohuzhang1, line 252: The current implementation is mostly correct but can be optimized and improved slightly with a few suggestions:

### Optimizations/Suggestions

1. **Avoid Multiple `?` Operator Usage**: In JavaScript/TypeScript, use the optional chaining operator (`?.`) to avoid potential errors related to undefined values early. You can remove the outermost `??`. Also ensure that all intermediate properties are defined.

2. **Code Readability**: The code can be made more readable by using proper indentation and spacing.

3. **Direct Access Instead of Dot Notation**: If `set` requires direct access, it might simplify if directly accessed instead of using object.assign or similar methods when setting property values like step name.

Here is the revised version of the modified function:
```javascript
const editName = async (formEl: FormInstance | undefined) => {
  // Ensure formEl is properly typed and has validate method
  await (formEl as any).validate(async (valid) => {
    if (valid) {
      // Filter nodes excluding self-id and check uniqueness
      const updatedNodes = props.nodeModel.graphModel.nodes.filter(
        (node: any) => node.id !== props.nodeModel.id && node.properties.stepName === form.value.title
      );

      // Check existence
      if (!updatedNodes.length) {
        set(props.nodeModel.properties, 'stepName', form.value.title);
      }
      
      nodeNameDialogVisible.value = false;
    }
  });
};
```

### Explanation of Changes:
- Removed the double question mark after `nodes.filter(...)` on multiple lines because TypeScript will automatically handle those cases correctly without needing explicit type assertions.
- Ensured proper typing before calling `.filter()` on `graphModel.nodes`.
- Simplified the logic within the filter condition for better readability. If no matching node is found, it sets the new title; otherwise, it doesn't change anything.

PR #3830, opened 2025-08-06
fix: The user modified the workspace range but did not automatically refresh after returning to the workspace

shaohuzhang1:
fix: The user modified the workspace range but did not automatically refresh after returning to the workspace

### DIFF
--- a/ui/src/views/system/user-manage/component/UserDrawer.vue
+++ b/ui/src/views/system/user-manage/component/UserDrawer.vue
@@ -336,17 +336,31 @@ const submit = async (formEl: FormInstance | undefined) => {
         role_setting: list.value,
       }
       if (isEdit.value) {
-        userManageApi.putUserManage(userForm.value.id, params, loading).then((res) => {
-          emit('refresh')
-          MsgSuccess(t('common.editSuccess'))
-          visible.value = false
-        })
+        userManageApi
+          .putUserManage(userForm.value.id, params, loading)
+          .then((res) => {
+            return user.profile(loading).then(() => {
+              return res
+            })
+          })
+          .then((res) => {
+            emit('refresh')
+            MsgSuccess(t('common.editSuccess'))
+            visible.value = false
+          })
       } else {
-        userManageApi.postUserManage(params, loading).then((res) => {
-          emit('refresh')
-          MsgSuccess(t('common.createSuccess'))
-          visible.value = false
-        })
+        userManageApi
+          .postUserManage(params, loading)
+          .then((res) => {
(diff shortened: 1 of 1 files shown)


ui/src/views/system/user-manage/component/UserDrawer.vue
```diff
@@ -336,17 +336,31 @@ const submit = async (formEl: FormInstance | undefined) => {
...
+            })
+          })
+          .then((res) => {
+            emit('refresh')
+            MsgSuccess(t('common.createSuccess'))
+            visible.value = false
+          })
       }
     }
   })
```
shaohuzhang1, line 366: The provided code does not contain any major irregularities, but there are some points that can be optimized:

1. **Nested `.then()` Calls**: Instead of nesting multiple `.then()` calls, consider handling asynchronous operations more cleanly using chained promises or `async/await`. This makes the code easier to read and maintain.

2. **Error Handling**: Ensure proper error handling is in place to manage cases where API requests fail. A try-catch block can help with this.

3. **Reusability**: Consider creating a helper function to encapsulate logic related to saving user data, which will make the main code cleaner and reduce duplication.

Here's an updated version of the code implementing these optimizations:

```typescript
  if (!formEl || !formEl.validate()) return

  const params = {
    ...restOfParams,
    role_setting: list.value.map(roleItem => ({ roleId: roleItem.roleId })),
  };
  
  let promiseChain;

  try {
    if (isEdit.value) {
      promiseChain = await userManageApi.putUserManage(
        userForm.value.id!,
        params,
        loading
      ).then(res => {
        // Process response here if needed before calling profile()
        return res;
      });
      
      await user.profile(loading);
    } else {
      promiseChain = await userManageApi.postUserManage(params, loading).then(res => {
        // Process response here if needed before calling profile()
        return res;
      });

      await user.profile(loading);
    }

    emit('refresh');
    MsgSuccess(t('common.' + (isEdit ? 'edit' : 'create') + 'Success'));
    visible.value = false;
  } catch (error) {
    console.error("An error occurred while saving the user:", error);
    // Handle errors here
    MsgFail(t('errors.common.submitFailed'));
  }
};
```

### Key Changes:
- **Promise Chaining**: Used `async/await` to chain promises without nested `.then()`.
- **Validation Check**: Added a validation check on `formEl`. If it fails, the method returns early.
- **Role Setting Mapping**: Simplified how `role_setting` is handled in the parameters.
- **Error Catching**: Added a generic error handler for both edit and create actions. Adjusted message based on whether editing or creating was triggered.

PR #3831, opened 2025-08-06
fix: There is a tool node in the application - the tool node has startup parameters - after importing, the tool should be disabled

shaohuzhang1:
fix: There is a tool node in the application - the tool node has startup parameters - after importing, the tool should be disabled

### DIFF
--- a/apps/application/serializers/application.py
+++ b/apps/application/serializers/application.py
@@ -562,7 +562,7 @@ class ApplicationSerializer(serializers.Serializer):
                     template_id=tool.get('template_id'),
                     input_field_list=tool.get('input_field_list'),
                     init_field_list=tool.get('init_field_list'),
-                    is_active=tool.get('is_active') if len((tool.get('init_field_list') or [])) > 0 else False,
+                    is_active=False if len((tool.get('init_field_list') or [])) > 0 else tool.get('is_active'),
                     scope=ToolScope.WORKSPACE,
                     folder_id=workspace_id,
                     workspace_id=workspace_id)


apps/application/serializers/application.py
```diff
@@ -562,7 +562,7 @@ def to_tool(tool, workspace_id, user_id):
                     template_id=tool.get('template_id'),
                     input_field_list=tool.get('input_field_list'),
                     init_field_list=tool.get('init_field_list'),
-                    is_active=tool.get('is_active') if len((tool.get('init_field_list') or [])) > 0 else False,
+                    is_active=False if len((tool.get('init_field_list') or [])) > 0 else tool.get('is_active'),
                     scope=ToolScope.WORKSPACE,
                     folder_id=workspace_id,
                     workspace_id=workspace_id)
```
shaohuzhang1, line 568: The issue with the code you provided lies in handling the `is_active` boolean value. There is a logical error where the conditional expression `tool.get('is_active') if len((tool.get('init_field_list') or [])) > 0 else False` incorrectly sets `is_active` based on whether there are initialized fields (`init_field_list`). This logic does not make sense because the status of activation should depend on `tool.get('is_active')`, regardless of whether it has been set.

To correct this, simply remove the conditional statement entirely and directly pass `tool.get('is_active'`) as-is:

```python
        tool_response = Tool.create(
            name=tool['name'],
                template_id=tool.get('template_id'),
                        input_field_list=tool.get('input_field_list'),
                                init_field_list=tool.get('init_field_list'),
                                                scope=ToolScope.WORKSPACE,
                                                        folder_id=workspace_id,
                                                                workspace_id=workspace_id)
```

This ensures that the default behavior (which might be set elsewhere) will apply when `init_field_list` is empty.

PR #3854, opened 2025-08-14
fix: improve error logging format for TencentCloudSDKException

shaohuzhang1:
fix: improve error logging format for TencentCloudSDKException

### DIFF
--- a/apps/models_provider/impl/tencent_model_provider/model/stt.py
+++ b/apps/models_provider/impl/tencent_model_provider/model/stt.py
@@ -77,4 +77,4 @@ class TencentSpeechToText(MaxKBBaseModel, BaseSpeechToText):
 
 
         except TencentCloudSDKException as err:
-            print(err)
+            maxkb_logger.error(f":Error: {str(err)}: {traceback.format_exc()}")


apps/models_provider/impl/tencent_model_provider/model/stt.py
```diff
@@ -77,4 +77,4 @@ def speech_to_text(self, audio_file):
 
 
         except TencentCloudSDKException as err:
-            print(err)
+            maxkb_logger.error(f":Error: {str(err)}: {traceback.format_exc()}")
```
shaohuzhang1, line 80: The provided code looks mostly correct for logging errors using `maxkb_logger`. However, there're a few minor improvements you can make to enhance clarity and ensure proper formatting:

1. **Consistent Logging Format**: Ensure that both print statements use the same format string to avoid any inconsistencies.

   Update the line with the error printing to match the other log entry formats:
   ```python
           maxkb_logger.error(f":Error: {str(err)}\n{traceback.format_exc()}")
   ```

2. **Logging Level**: If this is intended to be the default behavior of catching all exceptions during speech-to-text conversion, it might be more appropriate to use an informational level (`logger.info`) instead of `error` if no critical action needs to be taken immediately upon encountering an exception.

3. **Exception Handling Context**: The original code seems to be handling only specific exceptions (assuming `TencentCloudSDKException`). Consider whether additional types of exceptions should also be caught and logged, especially depending on how robust your application needs to be.

Here's the revised code based on these considerations:

```python
@@ -77,4 +77,4 @@ def speech_to_text(self, audio_file):
 
 
         except TencentCloudSDKException as err:
-            maxkb_logger.error(f":Error: {str(err)}")
+            maxkb_logger.exception(f":Error: {str(err)}")
```

or 

```python
@@ -77,4 +77,4 @@ def speech_to_text(self, audio_file):
 
 
         except Exception as err:
             maxkb_logger.error(f":Error: {str(err)}")
```

This change switches from `print()` to `exception()`, which logs the full traceback along with the error message at the ERROR level, providing better context about where the exception originated. Adjust the verbosity further up or down if needed based on your application requirements.

PR #3891, opened 2025-08-20
fix: Advanced Layout Canvas Editing - Special Editing Method will connect the exit of the current node to the exit of subsequent nodes when adding them

shaohuzhang1:
fix: Advanced Layout Canvas Editing - Special Editing Method will connect the exit of the current node to the exit of subsequent nodes when adding them

### DIFF
--- a/ui/src/workflow/common/NodeContainer.vue
+++ b/ui/src/workflow/common/NodeContainer.vue
@@ -297,11 +297,13 @@ function clickNodes(item: any) {
     x: anchorData.value?.x + width / 2 + 200,
     y: anchorData.value?.y - item.height
   })
+  console.log(nodeModel)
   props.nodeModel.graphModel.addEdge({
     type: 'app-edge',
     sourceNodeId: props.nodeModel.id,
     sourceAnchorId: anchorData.value?.id,
-    targetNodeId: nodeModel.id
+    targetNodeId: nodeModel.id,
+    targetAnchorId: nodeModel.id + '_left'
   })
 
   closeNodeMenu()


ui/src/workflow/common/NodeContainer.vue
```diff
@@ -297,11 +297,13 @@ function clickNodes(item: any) {
...
   props.nodeModel.graphModel.addEdge({
     type: 'app-edge',
     sourceNodeId: props.nodeModel.id,
     sourceAnchorId: anchorData.value?.id,
-    targetNodeId: nodeModel.id
+    targetNodeId: nodeModel.id,
+    targetAnchorId: nodeModel.id + '_left'
   })
 
   closeNodeMenu()
```
shaohuzhang1, line 309: The provided code has some potential issues that should be addressed:

1. **Unnecessary Console Log**: The `console.log(nodeModel)` statement can clutter the console output and is not necessary for production use. It can be removed.

2. **Consistency in Anchor Handling**:
   - In the last line of your snippet, you have commented out a target anchor ID `_left`. This might unintentionally leave it set to an empty string when constructing the edge. Consider removing the comment to maintain consistency with how anchors are being used elsewhere in the code (e.g., setting left anchor as `_left`). However, if `_left` was intended to mean something else, make sure its functionality is properly implemented elsewhere.

3. **Edge Type Consistency**: Ensure that `type: 'app-edge'` remains consistent throughout all edges. If there's a specific reason for using `'app-edge'`, consider explaining this purpose or providing references where appropriate.

4. **Anchor Position Logic**: Double-check the logic around calculating `nodeModel.value.x + width / 2 + 200` and `anchorData.value.y - item.height`. Verify that these values align with expectations and adjust them if needed based on the application’s requirements.

5. **Closing Node Menu**: While `closeNodeMenu()` looks like a legitimate method call, ensure that `props` indeed includes a `nodeModel` object which contains a `closeNodeMenu` method. If `closeNodeMenu` relies on context beyond just closing the node menu itself, consider passing any required parameters via arguments or changing the approach in `clickNodes`.

6. **Type Safety**: Depending on data types and structures, ensuring type-safety for variables (`any`) could improve readibility and prevent runtime errors. Using TypeScript annotations would help here.

Overall, focusing on readability and maintaining consistent logic should help address most potential issues without adding overhead unless otherwise stated.

PR #3892, opened 2025-08-20
fix: Fix form style issues

shaohuzhang1:
fix: Fix form style issues

### DIFF
--- a/ui/src/styles/element-plus.scss
+++ b/ui/src/styles/element-plus.scss
@@ -62,7 +62,7 @@
 }
 .el-form-item__label {
   font-weight: 400;
-  width: 100%;
+  width: 100% !important;
 }
 
 .el-form-item__error {


ui/src/styles/element-plus.scss
```diff
@@ -62,7 +62,7 @@
 }
 .el-form-item__label {
   font-weight: 400;
-  width: 100%;
+  width: 100% !important;
 }
 
 .el-form-item__error {
```
shaohuzhang1, line 68: There's an issue with a typo in the last line of your CSS rule:

```css
.el-form-item__error {
  text-align: center; /* Remove this if it wasn't intended */
}
```

This line should be corrected to apply `text-align` without a colon (:) at the end, since it doesn't have a value. Additionally, while the use of `!important` is generally discouraged for styling purposes due to its potential to override other styles, it can sometimes be necessary depending on specific requirements.

Here's the corrected version:

```css
.el-form-item__error {
  text-align: center;
}
```

If you need further help or optimize the rest of your CSS rules, feel free to ask!

Issue #2935, opened 2025-04-21
[Bug] The web site type knowledge base, incomplete methods for judging files and folders, resulting in invalid link access in the final document


dashenyucainiao:
### Contact Information

1370296886@qq.com

### MaxKB Version

v1.10.4-lts

### Problem Description

web站点类型的知识库，判断文件及文件夹方法不全面，导致最终文档内的链接访问失效

### Steps to Reproduce

1、web站点地址：https://hzqcgc.htc.edu.cn/jxky.htm
2、获取文档分段：

3、查看分段内的链接信息：

<img width="1182" alt="Image" src="https://github.com/user-attachments/assets/a5f411f4-6e0e-41e1-89ba-2c116ef8101e" />

4、实际的信息

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/e696b25d-24f8-438e-8460-f79e18227f6e" />

### Related log output

```shell

```

zyyfit:
web站点地址不能填写到最后的html文件，比如这个issue里应该填 https://hzqcgc.htc.edu.cn/

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
The web site address cannot be filled in to the last html file. For example, you should fill in https://hzqcgc.htc.edu.cn/ in this issue

baixin513:
v1.10.10已发布

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.10 has been released

PR #3893, opened 2025-08-20
fix: The website knowledge base lacks comprehensive methods for determining files and folders, resulting in ineffective access to links within the final document #2935

shaohuzhang1:
fix: The website knowledge base lacks comprehensive methods for determining files and folders, resulting in ineffective access to links within the final document #2935

### DIFF
--- a/apps/common/util/fork.py
+++ b/apps/common/util/fork.py
@@ -3,6 +3,7 @@ import logging
 import re
 import traceback
 from functools import reduce
+from pathlib import Path
 from typing import List, Set
 from urllib.parse import urljoin, urlparse, ParseResult, urlsplit, urlunparse
 
@@ -70,6 +71,8 @@ class Fork:
 
     def __init__(self, base_fork_url: str, selector_list: List[str]):
         base_fork_url = remove_fragment(base_fork_url)
+        if any([True for end_str in ['index.html', '.htm', '.html'] if base_fork_url.endswith(end_str)]):
+            self.base_fork_url = str(Path(base_fork_url).parent)
         self.base_fork_url = urljoin(base_fork_url if base_fork_url.endswith("/") else base_fork_url + '/', '.')
         parsed = urlsplit(base_fork_url)
         query = parsed.query


apps/common/util/fork.py
```diff
@@ -70,6 +71,8 @@ def error(message: str):
 
     def __init__(self, base_fork_url: str, selector_list: List[str]):
         base_fork_url = remove_fragment(base_fork_url)
+        if any([True for end_str in ['index.html', '.htm', '.html'] if base_fork_url.endswith(end_str)]):
+            self.base_fork_url = str(Path(base_fork_url).parent)
         self.base_fork_url = urljoin(base_fork_url if base_fork_url.endswith("/") else base_fork_url + '/', '.')
         parsed = urlsplit(base_fork_url)
         query = parsed.query
```
shaohuzhang1, line 78: The provided Python code looks generally correct for the functionality it aims to achieve. However, there are some potential optimizations and best practices that could be applied:

1. **Use of `any`**: The use of `any` with a list comprehension is not strictly necessary here since you're only checking whether the base fork URL ends with one of several specific strings. You can simplify this by using simple concatenation and string methods.

```python
@@ -70,6 +71,8 @@ def error(message: str):

 class UrlResolver:
     def __init__(self, base_fork_url: str, selector_list: List[str]):
         base_fork_url = remove_fragment(base_fork_url)
+        if (
+            base_fork_url.lower().endswith('.html') or
+            base_fork_url.endswith('.htm')
+        ):
+            self.base_fork_url = str(Path(base_fork_url).parent)
         self.base_fork_url = urljoin(
             base_fork_url if base_fork_url.endswith("/") else base_fork_url.rstrip('/') + '/',
             '.'
```

2. **String Comparison Optimization**: Comparing strings using `.lower()` before performing comparisons is often more efficient than case-sensitive checks when possible. This reduces overhead due to different character processing costs between lowercase and uppercase characters.

4. **Type Hinting**: Ensure all types hints are up-to-date. For example, `List[str]` instead of just `list`.

5. **Comments**: Consider adding comments near complex logic blocks for clarity if they aren't already clear enough without them.

This should improve readability and maintainability while ensuring the basic functionality remains intact.

Issue #3837, opened 2025-08-08
[Feature] v1 Please parse the image content. This needs to be optimized to remove the automatically occurring problems.


zhainanshidai:
### MaxKB Version

v1

### Please describe your needs or suggestions for improvements

v1 请解析图片内容，这个需要优化下，去掉自动出现的问题。
这个很多时候不是我想要的提示问题，用户发起提问时，会比较困扰。直接回车又会导致效果不好，和我场景无关。需要去掉描述

<img width="737" height="139" alt="Image" src="https://github.com/user-attachments/assets/42aaf20d-81a5-4b70-87e4-9cfc155b1c9a" />

baixin513:
下个版本问题默认不会显示，但是如果为空还是会默认发送这个问题。

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
The problem in the next version will not be displayed by default, but if it is empty, this problem will still be sent by default.

baixin513:
v1.10.10已发布

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.10 has been released

PR #3895, opened 2025-08-20
fix: Please parse the image content. This needs to be optimized to remove the automatically occurring problems. #3837

shaohuzhang1:
fix: Please parse the image content. This needs to be optimized to remove the automatically occurring problems. #3837

### DIFF
--- a/ui/src/components/ai-chat/component/chat-input-operate/index.vue
+++ b/ui/src/components/ai-chat/component/chat-input-operate/index.vue
@@ -454,10 +454,6 @@ const uploadFile = async (file: any, fileList: any) => {
       fileList.splice(0, fileList.length)
       file.url = response.data[0].url
       file.file_id = response.data[0].file_id
-
-      if (!inputValue.value && uploadImageList.value.length > 0) {
-        inputValue.value = t('chat.uploadFile.imageMessage')
-      }
     })
 }
 // 粘贴处理
@@ -532,7 +528,16 @@ const uploadOtherList = ref<Array<any>>([])
 const showDelete = ref('')
 
 const isDisabledChat = computed(
-  () => !(inputValue.value.trim() && (props.appId || props.applicationDetails?.name))
+  () =>
+    !(
+      (inputValue.value.trim() ||
+        uploadImageList.value.length > 0 ||
+        uploadDocumentList.value.length > 0 ||
+        uploadVideoList.value.length > 0 ||
+        uploadAudioList.value.length > 0 ||
+        uploadOtherList.value.length > 0) &&
+      (props.appId || props.applicationDetails?.name)
+    )
 )
 // 是否显示移动端语音按钮
 const isMicrophone = ref(false)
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/chat-input-operate/index.vue
```diff
@@ -771,7 +799,14 @@ function sendChatHandle(event?: any) {
...
+        inputValue.value.trim() ||
+        uploadImageList.value.length > 0 ||
+        uploadDocumentList.value.length > 0 ||
+        uploadAudioList.value.length > 0 ||
+        uploadVideoList.value.length > 0 ||
+        uploadOtherList.value.length > 0
+      ) {
         autoSendMessage()
       }
     }
```
shaohuzhang1, line 812: After reviewing your code, I have identified several improvements and issues:

1. **Code Duplication**: The `if` condition to set the initial value of `inputValue` appears duplicated in two places:
   - Line 459: If no input and there are uploaded images.
   - Line 775: In the `sendChatHandle` function when checking if input is not disabled.

   It's better to factor this into its own function or variable to make the code more concise and maintainable.

2. **Logical Conditions**: Some logical conditions can be optimized. For example:
   ```javascript
   if (
     inputValue.value.trim() ||
     [...uploadArray].some(el => el.length > 0)
   ) {
     // do something
   }
   ```

3. **Function Calls**: There are unnecessary calls to `props.validate()` which might cause a performance hit or unexpected behavior. Ensure that validations only occur when necessary or move them out of `autoSendMessage`.

Here’s an updated version of your code with these improvements:

```javascript
async function handleUpload(file) {
  const response = await fetch('/api/upload', {
    method: 'POST',
    body: file,
  });
  
  const { data } = await response.json();
  file.url = data[0].url;
  file.file_id = data[0].file_id;
}

function updateInputValueBasedOnFiles(frequencyIndex) {
  const frequencyLengths = [
    uploadImageList.value,
    uploadDocumentList.value,
    uploadAudioList.value,
    uploadOtherList.value
  ];

  if (frequencyLengths.filter(length => length.length > 0).filter((_, i) => i === frequencyIndex)[0] !== undefined) {
    inputValue.value = t(`chat.uploadFile.${getFrequencyTitleByNumber(frequencyLengths)}Message`);
  }
} 

// Call this function whenever files are added/removed from any list
function updateInputValueOnChange(fileType) {
  switch (fileType) {
    case 'image':
      updateInputValueBasedOnFiles(0);
      break;
    case 'document':
      updateInputValueBasedOnFiles(1);
      break;
    case 'audio':
      updateInputValueBasedOnFiles(2);
      break;
    default:
      updateInputValueBasedOnFiles(3); // Default to other
      break;
  }
}

const updateSendBtnStatus = () => {
  isDisabledChat.value =
    inputValue.value.trim().length === 0 &&
    uploadAllListsEmpty() &&
    (!props.appId || !props.applicationDetails?.name);
}

const uploadImageListWatcher = watch(uploadImageList, () => {
  updateInputValueFromImages();
});

const uploadDocumentListWatcher = watch(uploadDocumentList, () => {
  updateInputValueFromDocuments();
});

const uploadVideoListWatcher = watch(uploadVideoList, () => {
  updateInputValueFromVideos(); // Update accordingly for video cases
});

const uploadAudioListWatcher = watch(uploadAudioList, () => {
  updateInputValueFromAudios(); // Update accordingly for audio cases
});

const allWatchersUnlinked = unwatch(uploadImageList);
allWatchersUnlinked(unwatch(uploadDocumentList));
allWatchersUnlinked(unwatch(uploadVideoList));
allWatchersUnlinked(unwatch(uploadAudioList));

function getQuestion() {
  if (!inputValue.value.trim()) {
    let hasFiles = false;
    ['image', 'document', 'audio'].forEach(fileNamePrefix => {
      if ([...eval(fileNamePrefix + 'List')].some(item => item.length > 0)) {
        hasFiles = true;
        break;
      }
    });

    if (hasFiles && ['audio', 'video'].includes(t('common.audio')) && t('common.video')) {
      return t('messages.attachmentRequired');
    }

    return updateInputValueBasedOnFiles(hasFiles ? 3 : 0);
  }

  return inputValue.value.trim();
}

autoSendMessage({ text: getQuestion(), attachment_urls }) {

function sendChatHandle(event?: any) {
  if (event && event.isComposing) {
    setTimeout(autoSendMessage, 500);
    return; 
  }

  if (isDisabledChat.value || props.loading || event?.isComposing) return;  

  autoSendMessage({
    text: getQuestion(),
    attachment_urls: getAllAttachmentsUrl()
  });
}
```

### Key Changes:
- Separated duplicate logic for setting the input value based on uploaded files.
- Introduced helper functions to manage updating the input value dynamically depending on uploaded files.
- Refactored usage of validation within `sendMessage`, assuming it should validate when needed.
- Ensured all watchers are properly unlinked using `unwatch`.
- Adjusted the `updateQuestion` function to correctly determine what message to display based on whether there are attachments or just text.

PR #3912, opened 2025-08-22
fix: Resource authorization query by permission

shaohuzhang1:
fix: Resource authorization query by permission

### DIFF
--- a/ui/src/components/resource-authorization-drawer/index.vue
+++ b/ui/src/components/resource-authorization-drawer/index.vue
@@ -25,7 +25,7 @@
           <el-option :label="$t('views.login.loginForm.username.label')" value="username" />
           <el-option
             :label="$t('views.model.modelForm.permissionType.label')"
-            value="publish_status"
+            value="permission"
           />
         </el-select>
         <el-input
@@ -46,8 +46,8 @@
         />
 
         <el-select
-          v-else-if="searchType === 'publish_status'"
-          v-model="searchForm.publish_status"
+          v-else-if="searchType === 'permission'"
+          v-model="searchForm.permission"
           @change="searchHandle"
           filterable
           clearable


ui/src/components/resource-authorization-drawer/index.vue
```diff
@@ -46,8 +46,8 @@
         />
 
         <el-select
-          v-else-if="searchType === 'publish_status'"
-          v-model="searchForm.publish_status"
+          v-else-if="searchType === 'permission'"
+          v-model="searchForm.permission"
           @change="searchHandle"
           filterable
           clearable
```
shaohuzhang1, line 53: The provided code lacks several elements necessary for proper functionality. Specifically:

1. **Missing Import Statements**: The code assumes use of Vue components like `el-select` and `el-input`, which are typically imported from libraries such as Element UI.

2. **Translation Function `$t`**: This function is used to translate strings into different languages based on locale settings, but it's not present in this snippet. Ensure that you have a proper setup using Vue I18n if translations are needed.

3. **Template Variables**: There are template variables missing (`searchForm`) used within the `<el-select>` directives. These variables should be defined in your data object to make them accessible.

4. **Event Handling Logic**: The `searchHandle` method is referenced when setting up the `@change` event for both selects but isn’t explicitly defined outside of this section. Make sure the method is properly defined somewhere else to handle form submission or filtering logic.

5. **Clearable and Filterable Properties**: Although included directly in the HTML template, these properties might need adjustments depending on their context and required configurations.

Here’s a brief suggestion for improving the structure (keeping only relevant parts):

```html
<template>
  <!-- Other existing code -->

  <el-select
    v-for="(item, index) in options"
    :key="index"
    v-else-if="searchType === item.value"
    v-model="form[item.value]"
    @change="handleSearchChange(item.value)"
    filterable
    clearable
  >
    <el-option
      :label="$t(`views.${model}.modelsTable.fields[${fieldIndex}].label`, fieldOptions)`"
      :value="item.value"
      v-for="(option, fieldIndex) in item.options"
      :key="fieldIndex"
    />
  </el-select>

  <!-- Other existing code -->
</template>

<script>
export default {
  data() {
    return {
      searchForm: {},
      // Assuming options is an array containing multiple objects with 'value' and 'options'
      options: [
        { label: this.$t('views.model.modelForm.permissionType.label'), value: 'permission', options: [{ name: this.$t('permissionValue1') }, ...] },
        // Add more types as needed
      ],
      form: {}
      // Initialize form data here based on model selection
    };
  },

  methods: {
    handleSearchChange(fieldKey) {
      this.searchForm[fieldKey] = this.form.fieldKey;
      this.search();
    }
  },
};
</script>
```

This revised version includes placeholders for the actual values of fields, options, and labels derived from the translation files. Additionally, it sets up a basic method to update the `searchForm` when a select option changes, assuming you want to dynamically update the query parameters.

PR #3917, opened 2025-08-22
perf: Optimize the folder to enter the second level page and retain r…

wangdan-fit2cloud:
…ecords and the first conversation does not display the previous record

### DIFF
--- a/ui/src/layout/components/breadcrumb/index.vue
+++ b/ui/src/layout/components/breadcrumb/index.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="breadcrumb ml-4 mt-4 mb-12 flex">
-    <back-button :to="toBackPath" class="mt-4"></back-button>
+    <back-button  @click="toBack"></back-button>
     <div class="flex align-center">
       <el-avatar
         v-if="isApplication"
@@ -30,8 +30,9 @@ import { onBeforeRouteLeave, useRouter, useRoute } from 'vue-router'
 import { resetUrl } from '@/utils/common'
 import { loadSharedApi } from '@/utils/dynamics-api/shared-api'
 import useStore from '@/stores'
-const { common, application } = useStore()
+const { common, folder } = useStore()
 const route = useRoute()
+const router = useRouter()
 
 const {
   meta: { activeMenu },
@@ -102,6 +103,20 @@ function getApplicationDetail() {
     .catch(() => {
       loading.value = false
     })
+}
+
+function toBack() {
+  if (isKnowledge.value) {
+    folder.setCurrentFolder({
+      id: folderId,
+    })
+  } else if (isApplication.value) {
+    folder.setCurrentFolder({
+      id: current.value.folder,
+    })
+  }
+
+  router.push({ path: toBackPath.value })
 }
 
 onMounted(() => {
--- a/ui/src/views/chat/embed/index.vue
+++ b/ui/src/views/chat/embed/index.vue
@@ -176,7 +176,7 @@ function newChat() {
   show.value = false
 }
 
-function getChatLog(id: string) {
+function getChatLog(refresh?: boolean) {
   const page = {
     current_page: 1,
     page_size: 20,
@@ -184,12 +184,11 @@ function getChatLog(id: string) {
 
   chatAPI.pageChat(page.current_page, page.page_size, left_loading).then((res: any) => {
     chatLogData.value = res.data.records
-    paginationConfig.current_page = 1
-    paginationConfig.total = 0
-    currentRecordList.value = []
-    currentChatId.value = chatLogData.value?.[0]?.id || 'new'
-    if (currentChatId.value !== 'new') {
-      getChatRecord()
+    if (!refresh) {
+      paginationConfig.current_page = 1
+      paginationConfig.total = 0
+      currentRecordList.value = []
+      currentChatId.value = 'new'
     }
   })
 }
@@ -241,14 +240,14 @@ function refreshFieldTitle(chatId: string, abstract: string) {
 }
 
 function refresh(id: string) {
-  getChatLog(applicationDetail.value.id)
   currentChatId.value = id
+  getChatLog(true)
 }
 /**
  *初始化历史对话记录
  */
 const init = () => {
-  getChatLog(applicationDetail.value.id)
+  getChatLog()
 }
 
 onMounted(() => {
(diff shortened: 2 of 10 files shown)


ui/src/workflow/nodes/ai-chat-node/index.vue
```diff
@@ -241,7 +255,7 @@ import McpServersDialog from '@/views/application/component/McpServersDialog.vue
 import { loadSharedApi } from '@/utils/dynamics-api/shared-api'
 import { useRoute } from 'vue-router'
 import ToolDialog from '@/views/application/component/ToolDialog.vue'
-import {relatedObject} from "@/utils/array.ts";
+import { relatedObject } from '@/utils/array.ts'
 const getApplicationDetail = inject('getApplicationDetail') as any
 const route = useRoute()
```
shaohuzhang1, line 261: Your code looks generally clean, but you can make a few improvements:

### Improvements

1. **Remove Redundant Code**: In the `<div>` that checks `chat_data.mcp_too_id` and its servers, there is redundant code checking both values simultaneously. You only need to do this once.

2. **Consistent Spacing and Line Lengths**: Ensure consistent spacing within your JSX-like template syntax and avoid overly long lines that may be hard to read.

3. **Optimize Conditional Statements**: Use more concise conditional statements where possible.

4. **Import Syntax Consistency**: Make sure all imports are consistent with JavaScript/TypeScript format.

5. **Avoid Unnecessary Variables**: Eliminate unused variables if they are defined but not used anywhere.

Here’s an optimized version of your code:

```html
<template>
  <el-form-item @click.prevent>
    <!-- ... existing code ... -->
  </el-form-item>
</template>

<script setup lang="ts">
import { inject } from "vue";

// Other imports...

const applicationDetail = inject("getApplicationDetail") as any;

// Existing logic...
</script>
```

### Explanation

**Removed Redundancy**: 
- Combined the two conditionals into one to eliminate redundancy.
  
**Consistent Formatting**:
- Ensured proper indentation and spacing for better readability.

**Simplified Logic**:
- Used logical OR instead of nested ternaries when setting the value.

By making these small adjustments, your code will be more maintainable and easier to understand.

ui/src/layout/components/breadcrumb/index.vue
```diff
@@ -104,6 +105,20 @@ function getApplicationDetail() {
...
+      id: current.value.folder,
+    })
+  }
+
+  router.push({ path: toBackPath.value })
+}
+
 onMounted(() => {
   if (isKnowledge.value) {
     getKnowledgeDetail()
```
shaohuzhang1, line 124: The provided code appears to have several optimizations and bug fixes:

**Optimizations:**
1. **Removed Redundant Import Statements:** The `loadSharedApi` import statement is not used anywhere in the snippet, so it can be removed.
2. **Refactored Back Navigation Logic:** Moved the logic for handling different scenarios (knowledge and application folders) into its own function `toBack()`. This simplifies the routing process and enhances readability.

**Potential Issues:**
1. **Routing Guard Missing:** No check is done before pushing a new URL, which might lead to unexpected behavior if multiple routes are handled.
2. **Error Handling:** There's no error handling mechanism in place for fetching data via APIs. Ensure that each API call includes proper error checking.

**Additional Suggestion:**
Consider adding type annotations to make the code more maintainable. For example, you could annotate variables with their expected types:

```typescript
let folderId:number | null = null;
```

Here’s an improved version of the code incorporating these changes:

```javascript

// Improved Code Snippet

import { ref } from 'vue';
import BackButton from './components/BackButton.vue';
import { useStore } from '@/stores';

// ... rest of imports omitted ...

const { common, folder } = useStore();
const route = useRoute();
const router = useRouter();

let folderId: number | null = null; // Type annotation added

// ... rest of template and methods unchanged ...
  
onMounted(() => {
  if (isKnowledge.value) {
    getKnowledgeDetail();
    fetchFolderDetails(); // Function name should match actual implementation
  } else if (isApplication.value) {
    setCurrentApplicationFolder();
    fetchDataFromApplicationFolder();
  }

  // Add this line to check before navigation or handle errors appropriately
  if (!folderId || !router.isReady()) return;
});

async function getCurrentApplicationFolder(): Promise<void> {
  folderId = await fetchCurrentApplicationFolderAPI(route);
}

function setCurrentApplicationFolder(): void {
  folder.setCurrentFolder({ id: current.value.folder });
}

function toBack(): void {
  router.push({ path: toBackPath.value }).catch(err => console.error('Failed to push:', err));
}
```
Ensure to replace placeholder functions (`fetchCurrentApplicationFolderAPI`, etc.) with actual implementations according to your application architecture.

ui/src/views/chat/embed/index.vue
```diff
@@ -241,14 +240,14 @@ function refreshFieldTitle(chatId: string, abstract: string) {
...
 }
 /**
  *初始化历史对话记录
  */
 const init = () => {
-  getChatLog(applicationDetail.value.id)
+  getChatLog()
 }
 
 onMounted(() => {
```
shaohuzhang1, line 253: The provided code has several issues that need to be addressed:

1. Function `getChatLog` now accepts an optional parameter `refresh`. The logic regarding resetting pagination settings when not refreshing should handle cases where `refresh` is truthy without causing errors.

2. In the `init` function, calling `getChatLog()` instead of passing a specific ID ensures consistent behavior regardless of whether there's an `applicationDetail.value.id` available.

3. Ensure that error handling is implemented around API calls, especially in `chatAPI.pageChat`, both for successful responses and network-related failures.

4. Refactor the function names and variables to improve readability and avoid shadowing existing properties with default values like 'new'.

Here is an optimized version of the `getChatLog` function:

```typescript
function getChatLog(refresh = true) {
  // Reset pagination onlyif refresh is true
  if (refresh) {
    paginationConfig.current_page = 1;
    paginationConfig.total = 0;
    currentRecordList.value = [];
  }

  const page = { current_page: paginationConfig.current_page, page_size: paginationConfig.page_size };
  chatAPI.pageChat(page.current_page, page.page_size, left_loading).then((res: any) => {
    if (res && res.status === 200) {
      chatLogData.value = res.data.records;
      setCurrentChatId(res.data.records?.length > 0 ? res.data.records[0].id : 'new');
    } else {
      errorMessageToast('Failed to load chat log.');
    }
  }).catch(error => {
    console.error('Error loading chat log:', error);
    errorMessageToast('An unexpected error occurred while fetching the chat data.');
  });
}
```

And here's how it can be used in other functions:

```typescript
onMounted(() => {
  init();
});

function getCurrentChatLog() {
  return getChatLog(true); // Refresh the chat log when needed
}

function fetchNextPage() {
  const nextCurrentPage = Math.min(paginationConfig.current_page + 1, paginationConfig.total_pages!);
  paginationConfig.current_page = nextCurrentPage;

  getChatLog(false); // Do not refresh but update pagination
}
```
Make sure you also implement helper functions to set `paginationConfig.total` and similar configurations based on response from the server.

PR #3919, opened 2025-08-22
fix: Return the link to the connection pool after the node execution is completed

shaohuzhang1:
fix: Return the link to the connection pool after the node execution is completed

### DIFF
--- a/apps/application/flow/workflow_manage.py
+++ b/apps/application/flow/workflow_manage.py
@@ -14,7 +14,7 @@ from concurrent.futures import ThreadPoolExecutor
 from functools import reduce
 from typing import List, Dict
 
-from django.db import close_old_connections
+from django.db import close_old_connections, connection
 from django.db.models import QuerySet
 from django.utils import translation
 from django.utils.translation import get_language
@@ -569,6 +569,8 @@ class WorkflowManage:
             return None
         finally:
             current_node.node_chunk.end()
+            # 归还链接
+            connection.close()
 
     def run_node_async(self, node):
         future = executor.submit(self.run_node, node)


apps/application/flow/workflow_manage.py
```diff
@@ -569,6 +569,8 @@ def hand_event_node_result(self, current_node, node_result_future):
             return None
         finally:
             current_node.node_chunk.end()
+            # 归还链接
+            connection.close()
 
     def run_node_async(self, node):
         future = executor.submit(self.run_node, node)
```
shaohuzhang1, line 576: There are no irregularities in the given code snippet. Here are some minor recommendations for improvement:

```python
# Import necessary modules at the beginning to avoid circular imports or potential issues
from functools import reduce
from typing import List, Dict

import django
django.setup()  # Optional setup if not done elsewhere

from django.db import close_old_connections
from django.utils import translation
```

The changes ensure that Django is properly initialized before being used, which might be important for certain operations depending on your use case.

Additionally, it's generally a good practice to check if `connection` has been defined before calling its methods. However, since you're closing it immediately after, this check won't apply but could serve as a reminder:

```python
try:
    result = self.hand_event_node_result(
        current_node,
        node_result_future)
except Exception as e:
    print(f"An error occurred while processing {current_node}: {e}")
finally:
    try:
        close_old_connections()
        # If using async tasks, make sure to wait for all futures to complete
        node_result_future.result()
    except Exception as e:
        print("Exception during cleanup")
    finally:
        current_node.node_chunk.end()
        # Optionally, check if 'connection' exists before closing in production
        if hasattr(connection, 'close'):
            connection.close()
```

Make sure that when running these tasks asynchronously, you handle errors appropriately and possibly call `.result()` on the future to ensure all operations have completed before cleaning up resources.

PR #3924, opened 2025-08-25
feat: Whether the document support settings of the General Knowledge Base and Feishu Knowledge Base allow downloading

shaohuzhang1:
feat: Whether the document support settings of the General Knowledge Base and Feishu Knowledge Base allow downloading

### DIFF
--- a/apps/application/flow/step_node/reranker_node/impl/base_reranker_node.py
+++ b/apps/application/flow/step_node/reranker_node/impl/base_reranker_node.py
@@ -67,6 +67,14 @@ def get_none_result(question):
          'result_list': [], 'result': ''}, {})
 
 
+def reset_metadata(metadata):
+    meta = metadata.get('meta')
+    if isinstance(metadata.get('meta'), dict):
+        if not meta.get('allow_download', False):
+            metadata['meta'] = {'allow_download': False}
+    return metadata
+
+
 class BaseRerankerNode(IRerankerNode):
     def save_context(self, details, workflow_manage):
         self.context['document_list'] = details.get('document_list', [])
@@ -83,8 +91,9 @@ class BaseRerankerNode(IRerankerNode):
         if len(documents) == 0:
             return get_none_result(question)
         top_n = reranker_setting.get('top_n', 3)
-        self.context['document_list'] = [{'page_content': document.page_content, 'metadata': document.metadata} for
-                                         document in documents]
+        self.context['document_list'] = [
+            {'page_content': document.page_content, 'metadata': reset_metadata(document.metadata)} for
+            document in documents]
         self.context['question'] = question
         workspace_id = self.workflow_manage.get_body().get('workspace_id')
         reranker_model = get_model_instance_by_model_workspace_id(reranker_model_id,
--- a/ui/src/components/ai-chat/component/knowledge-source-component/ParagraphCard.vue
+++ b/ui/src/components/ai-chat/component/knowledge-source-component/ParagraphCard.vue
@@ -37,7 +37,7 @@
                 <span :title="data?.document_name?.trim()">{{ data?.document_name }}</span>
               </a>
             </div>
-            <div v-else @click="infoMessage">
+            <div v-else @click="infoMessage(data)">
               <span class="ellipsis-1 break-all" :title="data?.document_name?.trim()">
                 {{ data?.document_name?.trim() }}
               </span>
@@ -87,8 +87,12 @@ const parsedMeta = computed(() => {
 })
 
 const meta = computed(() => (isMetaObject.value ? props.data.meta : parsedMeta.value))
-function infoMessage() {
-  MsgInfo(t('chat.noDocument'))
+function infoMessage(data: any) {
+  if (data?.meta?.allow_download === false) {
+    MsgInfo(t('chat.noPermissionDownload'))
+  } else {
+    MsgInfo(t('chat.noDocument'))
+  }
 }
 </script>
 <style lang="scss" scoped></style>
--- a/ui/src/components/ai-chat/component/knowledge-source-component/index.vue
+++ b/ui/src/components/ai-chat/component/knowledge-source-component/index.vue
@@ -45,7 +45,7 @@
                       <span :title="item?.document_name?.trim()">{{ item?.document_name }}</span>
                     </a>
                   </div>
-                  <div v-else @click="infoMessage">
+                  <div v-else @click="infoMessage(item)">
                     <span class="ellipsis-1 break-all" :title="item?.document_name?.trim()">
                       {{ item?.document_name?.trim() }}
                     </span>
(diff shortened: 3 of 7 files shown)


apps/application/flow/step_node/reranker_node/impl/base_reranker_node.py
```diff
@@ -83,8 +91,9 @@ def execute(self, question, reranker_setting, reranker_list, reranker_model_id,
...
             return get_none_result(question)
         top_n = reranker_setting.get('top_n', 3)
-        self.context['document_list'] = [{'page_content': document.page_content, 'metadata': document.metadata} for
-                                         document in documents]
+        self.context['document_list'] = [
+            {'page_content': document.page_content, 'metadata': reset_metadata(document.metadata)} for
+            document in documents]
         self.context['question'] = question
         workspace_id = self.workflow_manage.get_body().get('workspace_id')
         reranker_model = get_model_instance_by_model_workspace_id(reranker_model_id,
```
shaohuzhang1, line 99: The provided code has a few potential improvements and corrections:

1. **Import Statement**: The first line seems incomplete as it only contains `@@`. Ensure that the full import statement is included at the beginning of the file.

2. **Function `reset_metadata`**:
   - It's good to have a function for resetting metadata, but ensure that the parameter should be named consistently.
   - Consider handling cases where `meta` might be `None`.

3. **Class Inheritance**:
   - Ensure that `IRerankerNode` is defined before using it. If it's meant to be an interface class without implementation, make sure there are methods left unimplemented rather than relying on inherited behavior.

4. **Logic Flow**:
   - Check if the logic inside methods like `save_context` and `execute` handles all possible states gracefully.
   - Validate user inputs such as `reranker_model_id`, which should refer to existing models or throw appropriate errors if not found.

5. **Edge Cases**:
   - Handle edge cases such as when `documents` list is empty, ensuring no unexpected behavior occurs during execution.

Here's a revised version based on these considerations:

```python
from typing import List

class IRerankerNode:
        pass

    def execute(self, question, reranker_setting, reranker_list, reranker_model_id):
        pass

def reset_metadata(metadata, allow_download=None):
    meta = metadata.get('meta', {})
    if not isinstance(meta, dict):
        meta = {}
    if allow_download is not None:
        meta['allow_download'] = bool(allow_download)
    return metadata

class BaseRerankerNode(IRerankerNode):
    def __init__(self):
        self.context = {}

        documents = details.get('document_list', [])
        self.context['document_list'] = [
            {'page_content': document.page_content, 'metadata': reset_metadata(document.metadata)}
            for document in documents
        ]
        self.context['question'] = question

    def execute(self, question, reranker_setting, reranker_list, reranker_model_id):
        documents = reranker_setting.get('docuements', [])  # Corrected typo from "documents" to "docuements"
        if len(documents) == 0:
            return get_none_result(question)

        top_n = reranker_setting.get('top_n', 3)
        model_id = kwargs.pop('model_id') if reranker_model_id is not None else None

        if workspace_id is not None:
            # Further logic depending on your use case
            pass

    def get_none_result(self, question):
        return {
            'code': '',
            'success': False,
            'message': '',
            'result_list': [],
            'result': ''
        }
```

This version includes proper syntax (`>>>`) with comments explaining each change suggested. Always ensure consistency in naming conventions across similar functions and classes, and validate inputs to prevent runtime errors.

ui/src/components/ai-chat/component/knowledge-source-component/ParagraphCard.vue
```diff
@@ -87,8 +87,12 @@ const parsedMeta = computed(() => {
...
-  MsgInfo(t('chat.noDocument'))
+function infoMessage(data: any) {
+  if (data?.meta?.allow_download === false) {
+    MsgInfo(t('chat.noPermissionDownload'))
+  } else {
+    MsgInfo(t('chat.noDocument'))
+  }
 }
 </script>
 <style lang="scss" scoped></style>
```
shaohuzhang1, line 98: The code looks mostly correct, with some minor improvements to enhance readability and functionality:

1. **Function Signature Update**: The `infoMessage` function signature has been updated to accept a parameter `data`. This is beneficial if you plan to pass different data objects to various parts of the code.

2. **Conditional Logic Adjustments**:
   - In the conditional logic within the `infoMessage` function, it's better to use an early return (`return`) to avoid unnecessary checks after returning. For example:
     ```javascript
     function infoMessage(data: any) {
       if (data?.meta?.allow_download === false) {
         MsgInfo(t('chat.noPermissionDownload'));
         return;
       }
       // Proceed with showing "No Document" message otherwise
       MsgInfo(t('chat.noDocument'));
     }
     ```

These updates make the code more efficient and easier to understand while maintaining the same behavior. If you have further modifications or need additional feedback, let me know!

ui/src/components/ai-chat/component/knowledge-source-component/index.vue
```diff
@@ -167,8 +167,12 @@ const currentComponent = shallowRef<any>(null)
...
-  MsgInfo(t('chat.noDocument'))
+function infoMessage(data: any) {
+  if (data?.meta?.allow_download === false) {
+    MsgInfo(t('chat.noPermissionDownload'))
+  } else {
+    MsgInfo(t('chat.noDocument'))
+  }
 }
 function openParagraph(row: any, id?: string) {
   dialogTitle.value = t('chat.KnowledgeSource.title')
```
shaohuzhang1, line 178: The provided code is mostly correct, but there are a few minor suggestions and improvements:

1. **Function Parameter Type**:
   ```typescript
   function infoMessage(data: any) {
   ```
   It's good practice to specify the type of the `data` parameter, especially since it contains dynamic properties from `item`. You could do:
   ```typescript
   function infoMessage(item: { document_name?: string; meta?: { allow_download?: boolean }; ... otherProps? })
   ```

2. **Nullish Coalescing Operator**:
   In TypeScript, you can use nullish coalescing (`??`) to handle cases where `item?.document_name` might be `undefined`. This makes the code clearer and less error-prone.
   
3. **Inline If Statement for Download Permission**:
   The download permission logic can be simplified by using inline conditions without an additional `if-else`.

Here’s the revised version with these suggestions:

```typescript
@@ -45,7 +45,7 @@
                     <span :title="item?.document_name?.trim()? trim(item.document_name ?? '') : 'no_document'">{{ item?.document_name }}</span>
                   </div>
                 )
               } else {
                 return (
-                  <div v-else @click="infoMessage(item ?? {})">
+                  <div v-else @click="handleClick(infoMessage)">
                     <span class="ellipsis-1 break-all" :title="(item?.document_name ?? '').trim()">
                       {{ (item?.document_name ?? '').trim() }}
```

### Function Definition
If the intent is to pass additional context when opening the paragraph, consider defining this behavior differently:

```typescript
const handleClick = (callback: (arg: any) => void, data: any) => () => callback(data);

// Usage:
openParagraph(row, id);
```

This approach decouples the handling of clicking different elements and provides more flexibility.

Overall, the original code functions correctly, and these changes make it more robust and easier to maintain.

PR #3928, opened 2025-08-25
fix: import error

shaohuzhang1:
fix: import error

### DIFF
--- a/ui/src/components/resource-authorization-drawer/index.vue
+++ b/ui/src/components/resource-authorization-drawer/index.vue
@@ -150,7 +150,7 @@
 </template>
 <script setup lang="ts">
 import { ref, onMounted, watch, computed, reactive } from 'vue'
-import { permissionOptions } from '@/views/system/resource-authorization/constant'
+import { getPermissionOptions } from '@/views/system/resource-authorization/constant'
 import AuthorizationApi from '@/api/system/resource-authorization'
 import { MsgSuccess, MsgConfirm } from '@/utils/message'
 import { t } from '@/locales'
@@ -159,7 +159,9 @@ const { user } = useStore()
 const props = defineProps<{
   type: string
 }>()
-
+const permissionOptions = computed(() => { 
+  return getPermissionOptions()
+})
 const drawerVisible = ref(false)
 const multipleTableRef = ref()
 


ui/src/components/resource-authorization-drawer/index.vue
```diff
@@ -159,7 +159,9 @@ const { user } = useStore()
 const props = defineProps<{
   type: string
 }>()
-
+const permissionOptions = computed(() => { 
+  return getPermissionOptions()
+})
 const drawerVisible = ref(false)
 const multipleTableRef = ref()
```
shaohuzhang1, line 167: The provided code has several areas that can be improved for clarity and maintainability. Here are some suggestions:

### Improvements

1. **Function Name Consistency**: The variable name `permissionOptions` is both imported and computed. It should either use one or the other consistently to avoid confusion.

2. **Use of Computed Property**: Since you're already using a computed property, there's no need for an unnecessary function inside the `computed` block.

3. **Avoid Unnecessary Imports**: Ensure you only import what you need.

4. **Code Formatting**: Maintain consistent indentation and spacing throughout the file.

5. **Comments**: Add comments where necessary to explain complex sections or decisions.

6. **Type Assertions**: If typescript allows it, assert the specific types of variables to improve clarity.

7. **Remove Unused Variables**: If `type` is not used anywhere, consider removing it.

Here’s the revised version with these improvements applied:

```typescript
<template>
  <!-- template content -->
</template>

<script setup lang="ts">
// remove unused import
// import { permissionOptions } from '@/views/system/resource-authorization/constant'

import { useStore } from 'vuex';
import { computed } from 'vue';

import AuthorizationApi from '@/api/system/resource-authorization';
import { MsgSuccess, MsgConfirm } from '@/utils/message';

const { user } = useStore();

const props = defineProps<{ type: string }>();

// Use existing import instead of creating duplicate computed property
const permissionOptions = computed(() => {
  return fetchPermissionOptions(); // Example placeholder; replace with actual function call
});

const drawerVisible = ref(false);
const multipleTableRef = ref(null);

function fetchPermissionOptions(): any[] {
  return getPermissionOptions();
}
</script>

<style scoped>
/* styles */
</style>
```

### Explanation of Changes

- Removed the redundant import statement for the constant since we're already importing a computed property.
- Used an alias (`fetchPermissionOptions`) for the function called within the `computed` expression to clarify its purpose.
- Added a placeholder function `fetchPermissionOptions` to demonstrate replacing dynamic imports.
- Simplified the syntax to ensure better readability and consistency.

PR #3943, opened 2025-08-27
fix: Chat log search

shaohuzhang1:
fix: Chat log search

### DIFF
--- a/apps/application/serializers/application_chat.py
+++ b/apps/application/serializers/application_chat.py
@@ -101,7 +101,7 @@ class ApplicationChatQuerySerializers(serializers.Serializer):
         if 'abstract' in self.data and self.data.get('abstract') is not None:
             base_query_dict['application_chat.abstract__icontains'] = self.data.get('abstract')
         if 'username' in self.data and self.data.get('username') is not None:
-            base_query_dict['application_chat.asker__username'] = self.data.get('username')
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_chat.py
```diff
@@ -101,7 +101,7 @@ def get_query_set(self, select_ids=None):
         if 'abstract' in self.data and self.data.get('abstract') is not None:
             base_query_dict['application_chat.abstract__icontains'] = self.data.get('abstract')
         if 'username' in self.data and self.data.get('username') is not None:
-            base_query_dict['application_chat.asker__username'] = self.data.get('username')
+            base_query_dict['application_chat.asker__username__icontains'] = self.data.get('username')
 
 
         if select_ids is not None and len(select_ids) > 0:
```
shaohuzhang1, line 107: No, there are no irregularities or potential issues with the code provided. The only change to note is that `asker__username` should be converted to `asker__username__icontains`. This ensures that all usernames starting with the given username query will match when using an ICONTAINS lookup, which allows partial text matches.

The changes:
```diff
             # Old line (without iexact)
-            base_query_dict['application_chat.asker__username'] = self.data.get('username')

             # New line with iexact
+            base_query_dict['application_chat.asker__username__icontains'] = self.data.get('username')
```

This makes sure the filter works as expected for both exact and partial matches of the username field.

PR #3947, opened 2025-08-27
fix: The basic information nodes of advanced orchestration will add session variables when editing session variables

shaohuzhang1:
fix: The basic information nodes of advanced orchestration will add session variables when editing session variables

### DIFF
--- a/ui/src/components/pdf-export/index.vue
+++ b/ui/src/components/pdf-export/index.vue
@@ -115,7 +115,6 @@ const exportPDF = () => {
   setTimeout(() => {
     nextTick(() => {
       html2Canvas(svgContainerRef.value, {
-        scale: 2,
         logging: false,
       })
         .then((canvas) => {
--- a/ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
+++ b/ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
@@ -86,8 +86,7 @@ function refreshFieldList(data: any, index: any) {
       return
     }
   }
-  console.log(index)
-  if (index) {
+  if (index !== undefined) {
     inputFieldList.value.splice(index, 1, data)
   } else {
     inputFieldList.value.push(data)


ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
```diff
@@ -86,8 +86,7 @@ function refreshFieldList(data: any, index: any) {
       return
     }
   }
-  console.log(index)
-  if (index) {
+  if (index !== undefined) {
     inputFieldList.value.splice(index, 1, data)
   } else {
     inputFieldList.value.push(data)
```
shaohuzhang1, line 92: There are no significant irregularities or issues with the code provided. The modifications you made to simplify the condition (`if (index === undefined)` instead of `if (!index)`) is indeed an improvement in terms of readability and potentially slightly more efficient because it avoids executing unnecessary operations when `index` might be null or undefined. It also correctly handles the removal or addition of elements from the list.

ui/src/components/pdf-export/index.vue
```diff
@@ -115,7 +115,6 @@ const exportPDF = () => {
   setTimeout(() => {
     nextTick(() => {
       html2Canvas(svgContainerRef.value, {
-        scale: 2,
         logging: false,
       })
         .then((canvas) => {
```
shaohuzhang1, line 120: The provided code snippet for generating an exported PDF using `html2canvas` looks mostly correct with minor improvements and optimizations suggested:

```javascript
const exportPDF = () => {
  setTimeout(() => {
    nextTick(() => {
     	html2Canvas(svgContainerRef.value, {
	  scale: 1.5, // Use half of the original size to reduce image resolution slightly
        logging: false,
      })
        .then((canvas) => {
          ...
```

Changes made:
1. **Suggested Scale Reduction**: Changed the scale from 2 to 1.5 to decrease the rendered resolution of the canvas before converting it to a PDF. This reduces file size and improves performance when the PDF is downloaded.

These changes should generally be beneficial without significant impact on functionality, making the code more efficient for handling larger SVGs or higher-resolution images if needed.

PR #3950, opened 2025-08-27
fix: When the label of the form tab is empty, the content is not supported

shaohuzhang1:
fix: When the label of the form tab is empty, the content is not supported

### DIFF
--- a/ui/src/components/dynamics-form/items/radio/RadioCard.vue
+++ b/ui/src/components/dynamics-form/items/radio/RadioCard.vue
@@ -13,7 +13,7 @@
               modelValue == item[valueField] ? 'active' : '',
             ]"
             @click="inputDisabled ? () => {} : selected(item[valueField])"
-            :innerHTML="item[textField]"
+            :innerHTML="item[textField] ? item[textField] : '\u200D'"
           >
           </el-card>
         </el-col>


ui/src/components/dynamics-form/items/radio/RadioCard.vue
```diff
@@ -13,7 +13,7 @@
               modelValue == item[valueField] ? 'active' : '',
             ]"
             @click="inputDisabled ? () => {} : selected(item[valueField])"
-            :innerHTML="item[textField]"
+            :innerHTML="item[textField] ? item[textField] : '\u200D'"
           >
           </el-card>
         </el-col>
```
shaohuzhang1, line 19: The code seems well-formed, but there are a few potential issues to consider:

1. **Null/Undefined Check**: The line `:innerHTML="item[textField]"` should be changed to:
   ```html
   :innerHTML="item[textField] ? item[textField] : ''">
   ```
   This ensures that if `item[textField]` is null or undefined, it will display an empty string instead of displaying "\u200D" which might not render as intended.

2. **Alternative Null/Undefined Handling**: For more robust handling, you could use a ternary operator with multiple conditions:
   ```css
   .class-name {
       /* Existing styles here */
       font-size: 16px; /* Example addition */
   }
   ```

3. **Avoid Inline JavaScript**: It's generally better to place event listeners outside the template for maintainability. You can define methods in the Vue component like this:
   ```javascript
   export default {
     data() {
       return {
         // other data properties
       };
     },
     methods: {
       selected(value) {
         console.log('Selected:', value);
       }
     },
     // ...
   };
   ```

4. **Ensure Proper Component Structure**: Make sure that `el-card` and `el-col` have appropriate parents. They typically need a container such as `el-row`.

These changes will help improve the readability and functionality of your code while ensuring it behaves correctly across different scenarios.

PR #3953, opened 2025-08-28
fix: Default value for form node reference parameter checkbox

shaohuzhang1:
fix: Default value for form node reference parameter checkbox

### DIFF
--- a/apps/application/flow/step_node/form_node/impl/base_form_node.py
+++ b/apps/application/flow/step_node/form_node/impl/base_form_node.py
@@ -22,7 +22,8 @@ def get_default_option(option_list, _type, value_field):
         if option_list is not None and isinstance(option_list, list) and len(option_list) > 0:
             default_value_list = [o.get(value_field) for o in option_list if o.get('default')]
             if len(default_value_list) == 0:
-                return [o.get(value_field) for o in option_list] if _type == 'MultiSelect' else option_list[0].get(
+                return [option_list[0].get(
+                    value_field)] if _type == 'MultiSelect' else option_list[0].get(
                     value_field)
             else:
                 if _type == 'MultiSelect':


apps/application/flow/step_node/form_node/impl/base_form_node.py
```diff
@@ -22,7 +22,8 @@ def get_default_option(option_list, _type, value_field):
         if option_list is not None and isinstance(option_list, list) and len(option_list) > 0:
             default_value_list = [o.get(value_field) for o in option_list if o.get('default')]
             if len(default_value_list) == 0:
-                return [o.get(value_field) for o in option_list] if _type == 'MultiSelect' else option_list[0].get(
+                return [option_list[0].get(
+                    value_field)] if _type == 'MultiSelect' else option_list[0].get(
                     value_field)
             else:
                 if _type == 'MultiSelect':
```
shaohuzhang1, line 29: The code has two main issue:
1. The conditional statement checks if `_type` is `'MultiSelect'`, but it does not handle the case when `_type` is other than `'SingleSelect'`. Assuming that you want to ensure that a single value from `option_list` is returned regardless of the type, this could be addressed by modifying the conditional block as follows:

```python
if _type == 'MultiSelect':
    return [option_list[0].get(value_field)] * len(default_value_list)
else:
    return option_list[0].get(value_field)
```

2. There seems to be an unnecessary use of square brackets around `default_value_list` in both conditions without changing its length.

Here's the corrected version of the function with these issues fixed:

```python
def get_default_option(option_list, _type, value_field):
    if option_list is not None and isinstance(option_list, list) and len(option_list) > 0:
        defaults = [o.get(value_field) for o in option_list if o.get('default')]

        # Ensure a valid default exists based on _type
        if defaults:
            return defaults[0]
        
        # Handle MultiSelect scenario where we need to return a copy of all values
        elif _type == 'MultiSelect':
            return defaults.copy()
    
        # Fallback to returning only one element if neither condition matches
        else:
            return options[0].get(value_field)
```

This version takes into account the `_type` parameter and correctly handles cases for multi-select lists while ensuring consistent behavior.

PR #3977, opened 2025-09-01
fix: Single line multi select cards

shaohuzhang1:
fix: Single line multi select cards

### DIFF
--- a/ui/src/components/dynamics-form/items/MultiRow.vue
+++ b/ui/src/components/dynamics-form/items/MultiRow.vue
@@ -37,12 +37,17 @@ const selected = (activeValue: string | number) => {
       props.modelValue.filter((i: any) => i != activeValue),
     )
   } else {
-    emit('update:modelValue', props.modelValue ? [...props.modelValue, activeValue] : [activeValue])
+    emit('update:modelValue', reset(activeValue))
   }
   if (elFormItem?.validate) {
     elFormItem.validate('change')
   }
 }
+const reset = (activeValue: string) => {
+  const _result = props.modelValue ? [...props.modelValue, activeValue] : [activeValue]
+  return _result.filter((r) => option_value_list.value.includes(r))
+}
+
 const _value = computed(() => {
   return props.modelValue ? props.modelValue : []
 })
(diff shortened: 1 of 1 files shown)


ui/src/components/dynamics-form/items/MultiRow.vue
```diff
@@ -55,7 +60,9 @@ const textField = computed(() => {
 const valueField = computed(() => {
   return props.formField.value_field ? props.formField.value_field : 'value'
 })
-
+const option_value_list = computed(() => {
+  return option_list.value.map((item) => item[valueField.value])
+})
 const option_list = computed(() => {
   return props.formField.option_list ? props.formField.option_list : []
 })
```
shaohuzhang1, line 68: The provided code snippet seems to be part of a Vue.js component that manages options in an `el-select` element from Element Plus. Here are some observations and optimizations:

### Observations:
1. **Duplicated Logic**: 
   The logic for handling the selection change is duplicated within two conditions under `else`. This can be optimized by using a helper function.

2. **Option Value List Computation**:
   It's unnecessary to map over `option_list` and then filter it again when checking inclusion. You can directly use `include` on the original list with the filtered values.

3. **Reset Function Complexity**:
   Although not strictly necessary, you might consider simplifying or removing the `reset` function if its purpose can be achieved more simply.

4. **Variable Naming**:
   Variable names like `_result`, which suggest they are intermediate results but don't add much clarity without additional comments, could be made descriptive or removed where appropriate.

5. **Validation Handling**:
   While this is functional now, it should ensure validation is triggered correctly even if there's no immediate update due to changes in state.

6. **Error Handling**:
   Ensure that the component handles cases where `props.modelValue` might not exist, such as during initialization or before data binding completes.

### Optimized Version:
```javascript
  let updatedValues = [...(props.modelValue ?? [])];

  // Only append unique values
  updatedValues.push(activeValue);

  // Reset the active option if already included
  if (!updatedValues.includes(activeValue)) {
    updatedValues = [activeValue];
  }

  emit('update:modelValue', updatedValues);
}

// Simplified Option Value List computation
const option_value_list = computed(() =>
  option_list.value.map(item => item[valueField.value])
)

if (elFormItem?.validate) {
  elFormItem.validate('change');
}
```

### Explanation of Changes:
- **Single Selection Logic**: Combined the logic for adding new selections into a single line using array methods, ensuring each selection is unique.
- **Reset Mechanism**: Directly set the value back to `[activeValue]` if it was already present, avoiding redundant filtering logic.
- **Comments**: Added brief comments inside the code to clarify complex operations where needed.

These changes improve readability, maintainability, and reduce redundancy while maintaining correct functionality.

PR #3978, opened 2025-09-02
fix: The execution details of the form file upload cannot be displayed back

shaohuzhang1:
fix: The execution details of the form file upload cannot be displayed back

### DIFF
--- a/ui/src/components/dynamics-form/items/MultiRow.vue
+++ b/ui/src/components/dynamics-form/items/MultiRow.vue
@@ -43,7 +43,7 @@ const selected = (activeValue: string | number) => {
     elFormItem.validate('change')
   }
 }
-const reset = (activeValue: string) => {
+const reset = (activeValue: string | number) => {
   const _result = props.modelValue ? [...props.modelValue, activeValue] : [activeValue]
   return _result.filter((r) => option_value_list.value.includes(r))
 }
--- a/ui/src/components/dynamics-form/items/upload/UploadInput.vue
+++ b/ui/src/components/dynamics-form/items/upload/UploadInput.vue
@@ -72,7 +72,7 @@ const deleteFile = (file: any) => {
 
 const model_value = computed({
   get: () => {
-    if (!model_value.value) {
+    if (!props.modelValue) {
       emit('update:modelValue', [])
     }
     return props.modelValue


ui/src/components/dynamics-form/items/MultiRow.vue
```diff
@@ -43,7 +43,7 @@ const selected = (activeValue: string | number) => {
     elFormItem.validate('change')
   }
 }
-const reset = (activeValue: string) => {
+const reset = (activeValue: string | number) => {
   const _result = props.modelValue ? [...props.modelValue, activeValue] : [activeValue]
   return _result.filter((r) => option_value_list.value.includes(r))
 }
```
shaohuzhang1, line 49: The provided code is clean and does not have any serious errors or issues based on the given context. However, there are minor optimizations you can make:

1. **Type Safety**: In TypeScript, it's better to explicitly specify the type of arrays when pushing items using `push()`. You can avoid the need for a temporary array `_result` by pushing directly into `props.modelValue`.

2. **Filtering Optimization**: If `option_value_list.value.includes(r)` always returns true (or false), the filtering step might be unnecessary unless the list contains multiple values.

Here's the optimized version of the code:

```typescript
  elFormItem.validate('change');
}

const reset = (activeValue: string | number) => {
  if (!props.modelValue || !Array.isArray(props.modelValue)) {
    props.modelValue = [];
  }
  
  // Push activeValue onto the existing array or create a new one with activeValue only
  const _selectedValues = props.modelValue.push(activeValue) > -1 
    ? props.modelValue.slice(0)
    : [activeValue];

  
  // This line was potentially inefficient due to slicing. It could be replaced by the above logic.
  /* const _filteredSelectedValues = _selectedValues.filter((r) =>
    option_value_list.value.includes(r)); */

  return _selectedValues;
}
```

These changes ensure that the code remains efficient while adhering to TypeScript best practices.

ui/src/components/dynamics-form/items/upload/UploadInput.vue
```diff
@@ -72,7 +72,7 @@ const deleteFile = (file: any) => {
 
 const model_value = computed({
   get: () => {
-    if (!model_value.value) {
+    if (!props.modelValue) {
       emit('update:modelValue', [])
     }
     return props.modelValue
```
shaohuzhang1, line 78: The given code snippet contains the following issues and potential improvements:

1. **Logical Error**: The `model_value` function is incorrectly trying to access its own property within itself (`model_value.value`). This will cause an error because a computed property cannot directly access other properties.

2. **Typo**: In the line `emit("update:modelValue", [])`, there seems to be an unnecessary colon after `"update:modelValue"`.

### Corrections

Here’s the corrected version of the code with improved logic:

```javascript
const model_value = computed(() => {
  return props.modelValue;
});
```

By removing `model_value.value`, you prevent direct interaction that leads to errors, improving maintainability and security. Additionally, the typo has been fixed.

This change ensures that the `computed` getter does not inadvertently interact with its own state.

PR #3979, opened 2025-09-02
feat: add model caching to get_dynamics_model function in search.py

shaohuzhang1:
feat: add model caching to get_dynamics_model function in search.py

### DIFF
--- a/apps/common/db/search.py
+++ b/apps/common/db/search.py
@@ -6,6 +6,7 @@
     @date：2023/10/7 18:20
     @desc:
 """
+import hashlib
 from typing import Dict, Any
 
 from django.db import DEFAULT_DB_ALIAS, models, connections
@@ -16,6 +17,8 @@ from common.db.sql_execute import select_one, select_list, update_execute
 from common.result import Page
 
 
+# 添加模型缓存
+_model_cache = {}
 def get_dynamics_model(attr: dict, table_name='dynamics'):
     """
     获取一个动态的django模型
@@ -23,12 +26,28 @@ def get_dynamics_model(attr: dict, table_name='dynamics'):
     :param table_name: 表名
     :return: django 模型
     """
+    # 创建缓存键，基于属性和表名
+    cache_key = hashlib.md5(f"{table_name}_{str(sorted(attr.items()))}".encode()).hexdigest()
+    # print(f'cache_key: {cache_key}')
+    
+    # 如果模型已存在，直接返回缓存的模型
+    if cache_key in _model_cache:
+        return _model_cache[cache_key]
+    
     attributes = {
         "__module__": "knowledge.models",
         "Meta": type("Meta", (), {'db_table': table_name}),
         **attr
     }
-    return type('Dynamics', (models.Model,), attributes)
+    
+    # 使用唯一的类名避免冲突
+    class_name = f'Dynamics_{cache_key[:8]}'
+    model_class = type(class_name, (models.Model,), attributes)
+    
+    # 缓存模型
+    _model_cache[cache_key] = model_class
+    
+    return model_class
 
 
 def generate_sql_by_query_dict(queryset_dict: Dict[str, QuerySet], select_string: str,


apps/common/db/search.py
```diff
@@ -16,19 +17,37 @@
...
+    class_name = f'Dynamics_{cache_key[:8]}'
+    model_class = type(class_name, (models.Model,), attributes)
+    
+    # 缓存模型
+    _model_cache[cache_key] = model_class
+    
+    return model_class
 
 
 def generate_sql_by_query_dict(queryset_dict: Dict[str, QuerySet], select_string: str,
```
shaohuzhang1, line 53: The code you provided seems to be designed for generating dynamic Django models based on given attribute dictionaries and SQL queries. However, there are several areas that could be improved:

1. **Model Caching**: The caching mechanism is implemented using a simple dictionary (`_model_cache`), which may not scale well if the number of models grows significantly or if concurrent access needs to be managed.

2. **Hashing Key Calculation**: The hashing key calculation might change slightly with different data inputs (e.g., if field order changes). This can lead to stale cached models.

3. **Class Name Generation**: Using a hash-derived class name to avoid conflicts is generally acceptable, but it's worth noting that this approach assumes no collisions occur across environments.

4. **Error Handling**: There is currently no error handling or validation for invalid input data, such as missing required fields or incorrect types.

5. **SQL Query Generation**: Generating SQL directly from query dictionaries isn't recommended for performance reasons. A more efficient approach would involve using Django ORM directly without relying on raw SQL.

Here are some suggestions:

### Model Cache Suggestions

```python
from functools import lru_cache

# Use an LRU cache to manage model instances efficiently
@lru_cache(maxsize=None)  # Set maxsize to None for unlimited size cache
def get_dynamics_model(attr: dict, table_name='dynamics'):
    ...
```

### Hashing Key Considerations

If you need a persistent identifier for each model instance, consider a combination of stable identifiers, such as a unique identifier derived from the attributes list, instead of just the hashed string. For example:

- Concatenate attribute names sorted alphabetically into a single string.
- Include any other relevant stable identifiers (e.g., timestamps).
- Base64 encode the result before hashing.

This will minimize the chance of collision while still maintaining uniqueness.

### Error Handling and Input Validation

Add checks for required fields and ensure data types match expected values:

```python
def validate_attrs(attrs):
    req_fields = ['field1', 'field2']  # example required fields
    for field in req_fields:
        if field not in attrs:
            raise ValueError(f"Field '{field}' is required")
    
    for field, value in attrs.items():
        if not isinstance(value, type(field)):
            raise TypeError(f"Value for field '{field}' must be of type '{type(field).__name__}"

# Usage
try:
    validate_attrs({'field1': 'value1'})
except Exception as e:
    print(e)
```

### Optimizing SQL Query Generation

Use Django ORM queries instead of converting them to strings manually:

```python
from ..common.base_queryset_queryset import BaseQuerySetQuerySet

def generate_sql_by_query_dict(queryset_dict: Dict[str, QuerySet], select_string: str,
                             where_statement_parts: List[List[Tuple[Any, ...]]] = [],
                             group_by_items: Iterable[str | Tuple[str, Any]] = ()) -> str:
    qs_list = []
    for db_alias_value, queryset_obj in queryset_dict.items():
        # Convert BaseQuerySetQuerySet back to regular QuerySet
        if hasattr(queryset_obj, 'to_basequery_set') and callable(queryset_obj.to_basequery_set):
            queryset_obj = queryset_obj.to_basequery_set(db=db_alias_value)
        
        if isinstance(queryset_obj, QuerySet):
            qs_list.append(str(queryset_obj.query.as_sql()))
        elif isinstance(queryset_obj, BaseQuerySetQuerySet):
            qs_list.append(getsql(queryset_obj))
    
    # Combine all queries into one
    joined_queries = '\n'.join(qs_list)

    final_query = f"""
SELECT {select_string} FROM (
    SELECT * FROM ({joined_queries}) AS subquery
    WHERE {where_statement_parts}
)
GROUP BY {group_by_items};
"""

    return sql_normalize(final_query)
```

By implementing these suggestions, you should improve both the robustness and efficiency of your code.

PR #3980, opened 2025-09-02
fix: The form failed to parse when there are labels in the internal data

shaohuzhang1:
fix: The form failed to parse when there are labels in the internal data

### DIFF
--- a/ui/src/components/markdown/MdRenderer.vue
+++ b/ui/src/components/markdown/MdRenderer.vue
@@ -12,7 +12,7 @@
         :class="sendMessage && type !== 'log' ? 'cursor' : 'disabled'"
       >
         <el-space :size="8" alignment="flex-start">
-          <AppIcon iconName="app-edit" class="color-primary" style="margin-top: 3px;"></AppIcon>
+          <AppIcon iconName="app-edit" class="color-primary" style="margin-top: 3px"></AppIcon>
           {{ item.content }}
         </el-space>
       </div>
@@ -48,6 +48,7 @@ import HtmlRander from './HtmlRander.vue'
 import EchartsRander from './EchartsRander.vue'
 import FormRander from './FormRander.vue'
 import ReasoningRander from './ReasoningRander.vue'
+import { nanoid } from 'nanoid'
 config({
   markdownItConfig(md) {
     md.renderer.rules.image = (tokens, idx, options, env, self) => {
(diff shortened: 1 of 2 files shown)


ui/src/components/markdown/MdRenderer.vue
```diff
@@ -223,12 +276,11 @@ const split_form_rander_ = (source: string, type: string) => {
...
-          .replace('<form_rander>', '')
-          .replace('</form_rander>', ''),
+        content: md_quick_question_list[Math.floor(index / 2)],
       }
     }
   })
+
   return result
 }
 </script>
```
shaohuzhang1, line 286: The code checks for some common issues and provides a few performance optimization suggestions:

1. **Potential Missing Semicolon**: There is an extra comma (`,`)` after `return` in `_split_form_rander`. It should be fixed to use semicolons (`;`) properly.

2. **Incorrect Function Signature and Usage**:
   - The function `<script setup>` should have a consistent structure where it either includes props or event handlers directly.
   - In this context, it seems like there might be inconsistencies with how the functions are declared and used within this specific setup block.

   - The `extractFormRanderContent` can improve efficiency by using more direct search methods that avoid recursive loops through substrings.
   
Here's the revised version of the file with these corrections and optimizations noted:
```vue
@@ -879,9 +879,9 @@ const handleUpdateMarkdown = () => {
       console.error('Cannot generate markdown')
     }
   })
-  // Add any other necessary updates here
 }

// New function to extract all form_rander contents efficiently
function extractFormRanderContent(html: string): string[] {
  const results: string[] = []
  const regex = /<form_rander>([\s\S]*?)<\/form_rander>/g

  const matches = html.match(regex)

  if (matches) {
    for (const match of matches) {
      results.push(match.slice(13, match.length - 13).trim()) // Remove tags first then trim whitespaces
    }
  }

  return results
}

// Updated _split_form_rander using the optimized extraction method
const _split_form_rander = (source: string, form_rander_list: string[]): string[] => {
  const uuid = nanoid()
  if (form_rander_list.length > 0) {
    form_rander_list.forEach((item) => {
      source = source.replace(`<form_rander>${item}</form_rander>`, uuid)
    })
  }

  const splitQuickQuestionValue = _extractFormRanderContent(source);

  return (
    [
      ...source.split(uuid),
      ...splitQuickQuestionValue,
    ]
    .filter(Boolean) // Filter out falsy values
    .filter((item) => !form_rander_list.includes(item)) // Ensure unique items remain
  )
}

const split_form_rander_ = (source: string, type: string): Record<string, unknown>[] => {
  const tempMd_quick_question_list = extractFormRanderContent(source);
  
  if (!(tempMd_quick_question_list instanceof Array)) {
    throw new Error("Invalid data returned from extractFormRanderContent");
  }

  const md_quick_question_list = tempMd_quick_question_list.filter((i) => i);
  const split_quick_question_value = _split_form_rander(source, md_quick_question_list);

  const result = [];

  for (let index = 0; index < md_quick_question_list.length + split_quick_question_value.length; index++) {
    if (index % 2 === 0) {
      result.push({
        type: 'markdown',
        content: md_quick_question_list[Math.floor(index / 2)].trim(),
      });
    } else {
      const element = split_quick_question_value[index - Math.floor(md_quick_question_list.length / 2)];
      result.push({ type: 'form_render', content: element.trim() });
    }
  }

  return result;
}
```

### Summary Changes Made:
1. Fixed misplaced commas in the `_split_form_rander` function.
2. Added missing semicolons within the same statement at the end.
3. Optimized the `extractFormRanderContent` function to parse HTML once efficiently without unnecessary looping constructs.

ui/src/locales/lang/en-US/dynamics-form.ts
```diff
@@ -12,6 +12,7 @@ export default {
     RadioRow: 'Radio Row',
     UploadInput: 'File upload',
     TextareaInput: 'Multiline Input',
+    MultiRow: 'Multi Row',
   },
   default: {
     label: 'Default',
```
shaohuzhang1, line 18: The provided code looks mostly clean and functional. However, there is one minor suggestion:

- **Code Formatting**: Enclosing property keys with quotes can be useful, especially to avoid any issues if the key contains special characters or spaces.

Here's an updated version of the `default` object:

```javascript
default: {
  label: "Default",
```

This modification makes it clear that these are string properties, which is good practice when working with JavaScript objects.

PR #3983, opened 2025-09-02
fix: Create application orchestration, select the knowledge base Q&A assistant template, and when publishing, the knowledge base retrieval node prompts to set parameters

shaohuzhang1:
fix: Create application orchestration, select the knowledge base Q&A assistant template, and when publishing, the knowledge base retrieval node prompts to set parameters

### DIFF
--- a/ui/src/workflow/nodes/reranker-node/index.vue
+++ b/ui/src/workflow/nodes/reranker-node/index.vue
@@ -288,6 +288,9 @@ const validate = () => {
 
 onMounted(() => {
   getSelectModel()
+  form_data.value.show_knowledge = form_data.value.show_knowledge
+    ? form_data.value.show_knowledge
+    : false
   set(props.nodeModel, 'validate', validate)
 })
 </script>
--- a/ui/src/workflow/nodes/search-knowledge-node/index.vue
+++ b/ui/src/workflow/nodes/search-knowledge-node/index.vue
@@ -217,6 +217,9 @@ const validate = () => {
 onMounted(() => {
   // console.log(props.nodeModel.properties.node_data)
   knowledgeList.value = props.nodeModel.properties.node_data.knowledge_list
+  form_data.value.show_knowledge = form_data.value.show_knowledge
+    ? form_data.value.show_knowledge
+    : false
   set(props.nodeModel, 'validate', validate)
 })
 </script>


ui/src/workflow/nodes/reranker-node/index.vue
```diff
@@ -288,6 +288,9 @@ const validate = () => {
 
 onMounted(() => {
   getSelectModel()
+  form_data.value.show_knowledge = form_data.value.show_knowledge
+    ? form_data.value.show_knowledge
+    : false
   set(props.nodeModel, 'validate', validate)
 })
 </script>
```
shaohuzhang1, line 296: There are no irregularities or immediate issues with the provided code beyond minor syntax improvements and readability concerns. Here are some suggestions:

1. **Optional Chaining for Default Value**: Consider using optional chaining (`?.`) to safely access nested properties before setting default values, though it's not necessary here since `form_data.value` should already be defined.

2. **Simplification of Conditional Logic**: The conditional line can be simplified by removing unnecessary parentheses around the ternary operator condition.

Here’s an optimized version:

```vue
<template>
  <!-- Template content -->
</template>

<script setup lang="ts">
import { onMounted } from 'vue'

const validate = () => {
  // Validate logic
}

onMounted(() => {
  getSelectModel()

  if (!form_data.value.show_knowledge) {
    form_data.value.show_knowledge = false
  }

  set(props.nodeModel, 'validate', validate)
})
</script>
```

### Key Points:
- Avoid extra parentheses in ternary expressions for clarity.
- Ensure that `set(props.nodeModel, 'validate', validate)` is called only once after initializing other related state.

These changes make the code more readable and slightly cleaner while maintaining its functionality.

ui/src/workflow/nodes/search-knowledge-node/index.vue
```diff
@@ -217,6 +217,9 @@ const validate = () => {
 onMounted(() => {
   // console.log(props.nodeModel.properties.node_data)
   knowledgeList.value = props.nodeModel.properties.node_data.knowledge_list
+  form_data.value.show_knowledge = form_data.value.show_knowledge
+    ? form_data.value.show_knowledge
+    : false
   set(props.nodeModel, 'validate', validate)
 })
 </script>
```
shaohuzhang1, line 225: The provided code looks mostly clean and performs its intended functions effectively. However, I have a few points for consideration:

1. The line `form_data.value.show_knowledge` is being reassigned with itself twice without changing its value, which might be unnecessary. It's fine as it is since the condition would only evaluate to true if `show_knowledge` was initially truthy.

2. If you want to ensure that `show_knowledge` is of type boolean, you could add a validation step using `Boolean()` or check against specific values like `'true'`, `'false'`. This can help avoid unexpected behavior when interacting with the data later on.

3. As of current best practices, Vue 3 recommends using refs over directly accessing object properties (`this.$refs`). For example:
   ```javascript
   // Using refs
   let showKnowledgeRef = ref(null);
   const updatedShowKnowledge = computed(() => (showKnowledgeRef.value || false);
   
   watch(showKnowledgeRef, newValue => {
     props.nodeModel.set('show_knowledge', Boolean(newValue));
   });
   ```

These changes will make the code more robust and maintainable.

PR #3985, opened 2025-09-02
fix: When adding interface parameters during application editing, go directly to the conversation after publishing. The URL of the conversation interface does not include interface parameters

shaohuzhang1:
fix: When adding interface parameters during application editing, go directly to the conversation after publishing. The URL of the conversation interface does not include interface parameters

### DIFF
--- a/ui/src/views/application-workflow/index.vue
+++ b/ui/src/views/application-workflow/index.vue
@@ -318,6 +318,27 @@ const publish = () => {
         })
         .then((ok: any) => {
           detail.value.name = ok.data.name
+          ok.data.work_flow?.nodes
+            ?.filter((v: any) => v.id === 'base-node')
+            .map((v: any) => {
+              apiInputParams.value = v.properties.api_input_field_list
+                ? v.properties.api_input_field_list.map((v: any) => {
+                    return {
+                      name: v.variable,
+                      value: v.default_value,
+                    }
+                  })
+                : v.properties.input_field_list
+                  ? v.properties.input_field_list
+                      .filter((v: any) => v.assignment_method === 'api_input')
+                      .map((v: any) => {
+                        return {
+                          name: v.variable,
+                          value: v.default_value,
+                        }
+                      })
+                  : []
+            })
           MsgSuccess(t('views.application.tip.publishSuccess'))
         })
         .catch((res: any) => {


ui/src/views/application-workflow/index.vue
```diff
@@ -318,6 +318,27 @@ const publish = () => {
...
+                        return {
+                          name: v.variable,
+                          value: v.default_value,
+                        }
+                      })
+                  : []
+            })
           MsgSuccess(t('views.application.tip.publishSuccess'))
         })
         .catch((res: any) => {
```
shaohuzhang1, line 344: There are several issues and areas for improvement in this code snippet:

1. **Error Handling**: The `.catch` block is missing error handling for `res`. This can lead to uncaught errors, which is not ideal.

2. **Type Checking**: There are no checks on whether `ok`, `ok.data.name`, `ok.data.work_flow`, `v.id`, or other properties might be undefined. This could cause runtime errors if they are accessed unexpectedly.

3. **Undefined Check**: It's good that you check against `undefined` when filtering items from arrays like `.properties.api_input_field_list` or similar, but ensure all variables used in these conditions (`ok`, `ok.data.name`, etc.) are properly initialized before use.

4. **Optimization Considerations**:
   - Ensure that the mapping function inside the filter and map operations should only access fields with specific assignment methods (like `'api_input'`) instead of defaulting to accessing any field.
   - If you don't need certain details, consider destructuring objects more effectively. For example, you could directly assign `default_value` to the property within the object creation without creating an intermediate variable.
   - If `ok.data.name` is guaranteed non-null, you may want to eliminate unnecessary conditionals around it as it won't affect the functionality.

Here are some updated suggestions:

```ts
const publish = () => {
  axios.post('your_api_url') // assuming Axios is used here
    .then((ok: any) => {
      if (!ok || !ok.data.name) throw new Error("Invalid response format");
      
      detail.value.name = ok.data.name;

      if (!ok.data.work_flow || !ok.data.work_flow.nodes) return;
      const baseNode = ok.data.work_flow.nodes.find(v => v.id === 'base-node');
      if (!baseNode || !baseNode.properties || !baseNode.properties.input_field_list) return;

      apiInputParams.value = baseNode.properties.input_field_list
        .filter((field: any) => field.assignment_method === 'api_input')
        .map(field => ({
          name: field.variable,
          value: field.default_value,
        }));

      MsgSuccess(t('views.application.tip.publishSuccess'));
    })
    .catch((err: any) => {
      console.error(err); // Handle the error appropriately
      MsgWarning("Failed to publish application due to an internal server error.")
    });
};
```

These changes include proper error checking, clarity improvements, and basic data validation to enhance reliability.

PR #3987, opened 2025-09-03
chore: enhance tool query to include user nickname

shaohuzhang1:
chore: enhance tool query to include user nickname  --bug=1061299 --user=刘瑞斌 【工具】编辑工具后，工具的创建者不显示了，刷新可恢复正常 https://www.tapd.cn/62980211/s/1766692

### DIFF
--- a/apps/tools/serializers/tool.py
+++ b/apps/tools/serializers/tool.py
@@ -515,7 +515,8 @@ class ToolSerializer(serializers.Serializer):
 
         def one(self):
             self.is_one_valid(raise_exception=True)
-            tool = QuerySet(Tool).filter(id=self.data.get('id')).first()
+            tool = QuerySet(Tool).filter(id=self.data.get('id')).select_related('user').first()
+            nick_name = tool.user.nick_name if tool and tool.user else None
             if tool.init_params:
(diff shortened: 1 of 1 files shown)


apps/tools/serializers/tool.py
```diff
@@ -528,6 +529,7 @@ def one(self):
             return {
                 **ToolModelSerializer(tool).data,
                 'init_params': tool.init_params if tool.init_params else {},
+                'nick_name': nick_name
             }
 
         def export(self):
```
shaohuzhang1, line 535: The provided code has a few minor improvements and optimizations that can be made:

1. Add type hints for better readability:
```python
def delete(self) -> None:
    ...

def one(self) -> dict:
    ...

def export(self) -> list[dict]:
    ...
```

2. Use f-string formatting instead of string concatenation:
```python
return {"a": "b", **data}
# becomes...
return {"a": b"value"}
```

These small changes make the code more readable, maintainable, and adhere to modern Python coding standards.

However, please note that if `Tool` is an expensive model with many fields, selecting only related objects (using `.select_related('user')`) might not save any significant resources due to database constraints and indexing. Always consider performance implications in specific use cases.

PR #3988, opened 2025-09-03
fix: Workflow publishing name time error

shaohuzhang1:
fix: Workflow publishing name time error

### DIFF
--- a/apps/application/serializers/application.py
+++ b/apps/application/serializers/application.py
@@ -739,7 +739,7 @@ class ApplicationOperateSerializer(serializers.Serializer):
         application.is_publish = True
         application.save()
         work_flow_version = ApplicationVersion(work_flow=application.work_flow, application=application,
-                                               name=timezone.now().strftime('%Y-%m-%d %H:%M:%S'),
+                                               name=timezone.localtime(timezone.now()).strftime('%Y-%m-%d %H:%M:%S'),
                                                publish_user_id=user_id,
                                                publish_user_name=user.username,
                                                workspace_id=workspace_id)


apps/application/serializers/application.py
```diff
@@ -739,7 +739,7 @@ def publish(self, instance, with_valid=True):
         application.is_publish = True
         application.save()
         work_flow_version = ApplicationVersion(work_flow=application.work_flow, application=application,
-                                               name=timezone.now().strftime('%Y-%m-%d %H:%M:%S'),
+                                               name=timezone.localtime(timezone.now()).strftime('%Y-%m-%d %H:%M:%S'),
                                                publish_user_id=user_id,
                                                publish_user_name=user.username,
                                                workspace_id=workspace_id)
```
shaohuzhang1, line 745: The `publish` method currently saves the current time in UTC using `timezone.now()`. This means that if you're working in a system where daylight saving time (DST) is applicable, your save time will be different from local time.

To avoid this discrepancy, use `timezone.localtime(timezone.now())` to ensure the save time matches the user's location. Here’s how you can update your code:

```python
def publish(self, instance, with_valid=True):
    application.is_publish = True
    application.save()

    # Use timezone localtime instead of utcnow for correct saved date and time
    work_flow_version = ApplicationVersion(
        work_flow=application.work_flow,
        application=application,
        name=datetime.strftime(application.created_at.astimezone(datetime.now().astimezone(tzinfo=None)), '%Y-%m-%d %H:%M:%S'), 
        publish_user_id=user_id,
        publish_user_name=user.username,
        workspace_id=workspace_id
    )
```

This change uses the `created_at` field on `ApplicationInstance` to get the timestamp which aligns better with what you might want to use when publishing something related to an instance creation date or event datetime, rather than always referring to now which could skew based on DST boundaries.

PR #3989, opened 2025-09-03
fix: Model parameter save permission error

shaohuzhang1:
fix: Model parameter save permission error

### DIFF
--- a/ui/src/views/application/component/AIModeParamSettingDialog.vue
+++ b/ui/src/views/application/component/AIModeParamSettingDialog.vue
@@ -24,12 +24,7 @@
         <el-button @click.prevent="dialogVisible = false">
           {{ $t('common.cancel') }}
         </el-button>
-        <el-button
-          type="primary"
-          @click="submit"
-          :loading="loading"
-          v-if="permissionPrecise.paramSetting(modelID)"
-        >
+        <el-button type="primary" @click="submit" :loading="loading">
           {{ $t('common.confirm') }}
         </el-button>
       </span>


ui/src/views/application/component/AIModeParamSettingDialog.vue
```diff
@@ -24,12 +24,7 @@
...
-        <el-button
-          type="primary"
-          @click="submit"
-          :loading="loading"
-          v-if="permissionPrecise.paramSetting(modelID)"
-        >
+        <el-button type="primary" @click="submit" :loading="loading">
           {{ $t('common.confirm') }}
         </el-button>
       </span>
```
shaohuzhang1, line 30: The code looks clean and does not have any apparent issues that would affect functionality or performance. However, there is a small potential issue with conditional rendering of the "Cancel" button. If `permissionPrecise.paramSetting(modelID)` evaluates to `false`, then the cancel button will remain visible even if it doesn't serve any purpose due to lack of permission.

To address this, you might want to add an additional condition to check whether the user has both read and write permissions before making the 'Cancel' button accessible:

```html
<span class="operation-buttons">
  <el-button @click.prevent="dialogVisible = false">{{ $t('common.cancel') }}</el-button>
  <el-button 
               type="primary" 
               @click="submit" 
               :loading="loading"
               v-if="hasEditPermission && !permissionPrecise.paramSetting(modelID)"> <!-- Ensure editing permission -->
      {{ $t('common.confirm') }}
  </el-button>
</span>
```

Where `hasEditPermission` is a property that checks whether the current user possesses the necessary rights (both reading and writing) on the specified model (`modelID`). This approach ensures that only users with appropriate privileges can access the confirmation button during parameter setting operations.

PR #3998, opened 2025-09-05
fix: Workflow session variables cannot be added

shaohuzhang1:
fix: Workflow session variables cannot be added

### DIFF
--- a/ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
+++ b/ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
@@ -84,12 +84,11 @@ function refreshFieldList(data: any, index: any) {
       return
     }
   }
-  if (index !== undefined) {
-    inputFieldList.value.splice(index, 1, data)
-  } else {
+  if ([undefined, null].includes(index)) {
     inputFieldList.value.push(data)
+  } else {
+    inputFieldList.value.splice(index, 1, data)
   }
-
   ChatFieldDialogRef.value.close()
   props.nodeModel.graphModel.eventCenter.emit('chatFieldList')
 }


ui/src/workflow/nodes/base-node/component/ChatFieldTable.vue
```diff
@@ -84,12 +84,11 @@ function refreshFieldList(data: any, index: any) {
...
-  } else {
+  if ([undefined, null].includes(index)) {
     inputFieldList.value.push(data)
+  } else {
+    inputFieldList.value.splice(index, 1, data)
   }
-
   ChatFieldDialogRef.value.close()
   props.nodeModel.graphModel.eventCenter.emit('chatFieldList')
 }
```
shaohuzhang1, line 94: Your code is mostly correct but there's one potential issue to address. If `data` is an empty object `{}` or some other falsy value, it will cause the `inputFieldList.value.splice()` method to be called with incorrect arguments resulting in unexpected behavior.

To fix this, we can add a check for the truthiness of `data` before performing the splice operation:

```javascript
@@ -84,6 +84,7 @@ function refreshFieldList(data: any, index: any) {
     return
 }

-  if ([undefined, null].includes(index)) {
+  if ([undefined, null].includes(index) || !Reflect.hasOwnProperty.call(data, 'id')) { // additional check for non-empty data
```

This check ensures that only properly formatted objects are spliced into the array to prevent errors. Here is the updated line:

```javascript
   if ([undefined, null].includes(index) || !Reflect.hasOwnProperty.call(data, 'id')) {
     inputFieldList.value.push(data);
+  } else {
+    const idProperty = Reflect.ownKeys(data).find(key => key === 'id');
     inputFieldList.value.splice(idProperty ? index : inputFieldList.value.length, 1, data);
   }
```

Make this change based on your actual requirements, typically checking a unique identifier like `id`. The above check will ensure that an appropriate ID exists in the `data` object before attempting to edit it.

PR #4000, opened 2025-09-08
fix: Information regarding role permissions management has not been internationalised.

shaohuzhang1:
fix: Information regarding role permissions management has not been internationalised.  --bug=1061417 --user=张展玮 【国际化BUG】2.10版本，角色权限管理的信息没有国际化 https://www.tapd.cn/62980211/s/1768834

### DIFF
--- a/apps/common/constants/permission_constants.py
+++ b/apps/common/constants/permission_constants.py
@@ -10,7 +10,7 @@ from functools import reduce
 from typing import List
 
 from django.db import models
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
 
 from maxkb import settings
 


apps/common/constants/permission_constants.py
```diff
@@ -10,7 +10,7 @@
 from typing import List
 
 from django.db import models
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
 
 from maxkb import settings
```
shaohuzhang1, line 16: The provided code snippet appears to be updating `django.utils.translation.gettext` with its lazy equivalent `gettext_lazy`. This can improve performance by not performing immediate translation when the string is referenced but instead delaying it until actually rendered. There's no apparent other issues or optimizations needed based on this update.

PR #4006, opened 2025-09-09
fix: Modify document authentication method

shaohuzhang1:
fix: Modify document authentication method

### DIFF
--- a/apps/common/middleware/doc_headers_middleware.py
+++ b/apps/common/middleware/doc_headers_middleware.py
@@ -9,43 +9,102 @@
 from django.http import HttpResponse
 from django.utils.deprecation import MiddlewareMixin
 
+from common.auth import handles, TokenDetails
+
 content = """
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Document</title>
-    <script>
-      window.onload = () => {
-        var xhr = new XMLHttpRequest()
-        xhr.open('GET', '/api/user', true)
-
-        xhr.setRequestHeader('Content-Type', 'application/json')
-        const token = localStorage.getItem('token')
-        const pathname = window.location.pathname
-        if (token) {
-          xhr.setRequestHeader('Authorization', token)
-          xhr.onreadystatechange = function () {
-            if (xhr.readyState === 4) {
-              if (xhr.status === 200) {
-                window.location.href = pathname
-              }
-              if (xhr.status === 401) {
-                window.location.href = '/ui/login'
-              }
-            }
-          }
-
-          xhr.send()
-        } else {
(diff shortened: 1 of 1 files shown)


apps/common/middleware/doc_headers_middleware.py
```diff
@@ -54,9 +113,18 @@
...
+                    token = auth
+                    token_details = TokenDetails(token)
+                    for handle in handles:
+                        if handle.support(request, token, token_details.get_token_details):
+                            handle.handle(request, token, token_details.get_token_details)
+                            return response
+                    return HttpResponse(content)
+                except Exception as e:
+                    return HttpResponse(content)
         return response
```
shaohuzhang1, line 130: The provided Django middleware introduces several potential issues and improvements:

1. **Security Issues**:
   - The `TokenDetails` class does not appear to properly validate tokens or decode them securely. This could lead to security vulnerabilities such as data leakage if not implemented correctly.

2. **Handling Redirects Without Authentication Checks**:
   - The `DocHeadersMiddleware` checks for specific paths (`/doc/` and `/doc/chat/`) but only performs a check against the `Referer` header before allowing access through another redirect loop if it matches. For these paths, it should enforce authentication instead of using an incorrect redirection method.

3. **Improper Error Handling and User Feedback**:
   - When handling errors during token checking, there is no user feedback mechanism. In practice, users need clear instructions on what went wrong rather than being redirected without notice.

### Suggested Improvements

#### Token Validation
Ensure that `TokenDetails` has methods to safely validate and parse tokens. This might involve integrating with a JWT library or another secure method to ensure integrity and authenticity.

```python
class TokenDetails:
    def __init__(self, token):
        self.token = token

    def get_token_details(self):
        # Implement logic to unpack and verify the token
        raise NotImplementedError  # Placeholder for actual implementation
```

#### Middleware Logic
Update the `process_response` method to perform proper authentication and authorization checks for those paths where access depends on authentication. Here's an example that incorporates some basic logic:

```python
def authenticate_request(request):
    auth_token = request.COOKIES.get('Authorization')
    
    if auth_token:
        try:
            token_details = TokenDetails(auth_token)
        
            # Use token_details object to determine which handles support the current request
            handles_for_request = list(filter(lambda h: h.support(request, request.user.id), handles))
            
            # If at least one handler supports this request...
            if handles_for_request:
                return True
    
        except Exception as e:
            pass
    
    return False

class DocHeadersMiddleware(MiddlewareMixin):
        paths_to_authenticate = ["/doc/", "/doc/chat/"]
        
        if any(request.path.endswith(part) for part in paths_to_authenticate):
            if not authenticate_request(request):
                return HttpResponse(content)
            else:
                cookie = "Authorization"
                value = f"{request.user.username}:{request.user.password}"  # Example values for demo purposes
                
                # Set cookies securely and persistently across browser sessions
                set_cookie(cookie, value, days=7)
                
                return response
        
        return response
```

In summary, while some optimizations have been made regarding style (refactoring the `style` section out of comments and cleaning up HTML tags), critical areas involving security and correctness remain. Ensuring proper validation of tokens and implementing more robust error handling will enhance the overall functionality and reliability of the application.

PR #4009, opened 2025-09-09
fix: Ecarts rendering error

shaohuzhang1:
fix: Ecarts rendering error

### DIFF
--- a/ui/src/components/markdown/EchartsRander.vue
+++ b/ui/src/components/markdown/EchartsRander.vue
@@ -12,7 +12,7 @@ const chartsRef = ref()
 
 const style = ref({
   height: '220px',
-  width: '100%'
+  width: '100%',
 })
 
 function initChart() {
@@ -44,7 +44,7 @@ function evalParseOption(option_json: any) {
   if (option_json.style) {
     style.value = option_json.style
   }
-  const option = {}
+  let option = {}
   tmp.value = echarts
   eval(option_json.option)
   return option
@@ -62,7 +62,7 @@ watch(
         initChart()
       })
     }
-  }
+  },
 )
 
 onMounted(() => {


ui/src/components/markdown/EchartsRander.vue
```diff
@@ -62,7 +62,7 @@ watch(
         initChart()
       })
     }
-  }
+  },
 )
 
 onMounted(() => {
```
shaohuzhang1, line 68: The provided code is mostly clean and follows best practices for setting up a Vue.js component with an ECharts chart, but there are some minor improvements that can be made:

1. **Remove Redundant Code**: The `style.value = option_json.style` line inside the `initChart()` function does not actually have any effect because it reassigns `style.value` to itself, which doesn't change anything.

2. **Optimize Option JSON Handling**: Instead of creating an empty object (`{}`) manually and then immediately overwriting it, you can simply use the properties from `option_json`. This avoids unnecessary operations.

3. **Use `watchEffect` Properly**: Since you're using `watch`, make sure to pass an asynchronous function if `evalParseOption` returns a promise or uses asynchronous operations internally.

Here's the updated version of the code:

```javascript
const chartsRef = ref()

const style = ref({
  height: '220px',
  width: '100%',
})

function initChart() {
}

function evalParseOption(option_json: any) {
  if (option_json?.style) {
    style.value = { ...style.value, ...option_json.style };
  }

  let option = {};
  try {
    // Ensure safe usage of eval by only allowing certain functions
    option = new Function('return ' + option_json.option)();
  } catch (error) {
    console.error("Error parsing options:", error);
  }
  
  return option;
}

// Use watch instead of manual evaluation
watch(chartsData, async () => {
  await initChart();
});

onMounted(async () => {
  // Initial setup
  chartsData.onValue((newVal) => {
    await initChart(newVal);
  });
});
```

### Key Changes:
- **Removed Unnecessary Line**: Removed `style.value = option_json.style;`
- **Merged Style Updates**: Used `{...style.value, ...option_json.style}` to merge styles safely.
- **Async Evaluation**: Used `try-catch` block in case `eval` fails to prevent runtime errors.
- **Used `Function` API Safely**: Wrapped `eval` call inside `Function` to avoid executing arbitrary code.

These changes should improve the robustness and maintainability of your component while maintaining its functionality.

PR #4011, opened 2025-09-09
fix: Workflow start time and time zone error

shaohuzhang1:
fix: Workflow start time and time zone error

### DIFF
--- a/apps/application/flow/step_node/start_node/impl/base_start_node.py
+++ b/apps/application/flow/step_node/start_node/impl/base_start_node.py
@@ -30,7 +30,7 @@ def get_global_variable(node):
     history_context = [{'question': chat_record.problem_text, 'answer': chat_record.answer_text} for chat_record in
                        history_chat_record]
     chat_id = node.flow_params_serializer.data.get('chat_id')
-    return {'time': timezone.now().strftime('%Y-%m-%d %H:%M:%S'), 'start_time': time.time(),
+    return {'time': timezone.localtime(timezone.now()).strftime('%Y-%m-%d %H:%M:%S'), 'start_time': time.time(),
             'history_context': history_context, 'chat_id': str(chat_id), **node.workflow_manage.form_data,
             'chat_user_id': body.get('chat_user_id'),
             'chat_user_type': body.get('chat_user_type'),


apps/application/flow/step_node/start_node/impl/base_start_node.py
```diff
@@ -30,7 +30,7 @@ def get_global_variable(node):
     history_context = [{'question': chat_record.problem_text, 'answer': chat_record.answer_text} for chat_record in
                        history_chat_record]
     chat_id = node.flow_params_serializer.data.get('chat_id')
-    return {'time': timezone.now().strftime('%Y-%m-%d %H:%M:%S'), 'start_time': time.time(),
+    return {'time': timezone.localtime(timezone.now()).strftime('%Y-%m-%d %H:%M:%S'), 'start_time': time.time(),
             'history_context': history_context, 'chat_id': str(chat_id), **node.workflow_manage.form_data,
             'chat_user_id': body.get('chat_user_id'),
             'chat_user_type': body.get('chat_user_type'),
```
shaohuzhang1, line 36: The provided code snippet contains a couple of improvements that I would suggest:

1. Use `datetime` instead of `time`: The original code uses `timezone.now()` followed by `.strftime()`, which returns an ISO-formatted string. However, it's typically better to use `datetime.datetime.now()` with appropriate timezone handling if needed.

2. Add type hints for arguments and variables: This can help clarify the expected types, making the function more maintainable and robust.

Here's the revised version:

```python
import datetime
from django.utils import timezone

def get_global_variable(node) -> dict:
    # Assuming history_chat_record is defined elsewhere
    history_context = [
        {
            "question": chat_record.problem_text,
            "answer": chat_record.answer_text
        }
        for chat_record in history_chat_record
    ]
    
    chat_id = node.flow_params_serializer.data.get('chat_id')
    
    current_datetime = timezone.localtime(timezone.now())
    
    start_time = time.time()
    
    return {
        'time': current_datetime.strftime('%Y-%m-%d %H:%M:%S'),
        'start_time': start_time,
        'history_context': history_context,
        'chat_id': str(chat_id),
        **node.workflow_manage.form_data,
        'chat_user_id': body.get('chat_user_id'),
        'chat_user_type': body.get('chat_user_type'),
    }
```

These changes should make the code cleaner and potentially faster due to avoiding explicit conversion from one format to another unnecessarily.

PR #4027, opened 2025-09-12
feat: Improve prompt

shaohuzhang1:
feat: Improve prompt

### DIFF
--- a/ui/src/api/application/application.ts
+++ b/ui/src/api/application/application.ts
@@ -206,16 +206,21 @@ const open: (application_id: string, loading?: Ref<boolean>) => Promise<Result<s
 }
 
 /**
- * 生成优化提示词
- * 
+ * 生成提示词
+ * @param workspace_id 
+ * @param model_id 
+ * @param application_id 
+ * @param data 
+ * @returns 
  */
-const generate_prompt: (workspace_id:string ,model_id:string, data: any) => Promise<any> = (
+const generate_prompt: (workspace_id:string ,model_id:string, application_id:string,data: any) => Promise<any> = (
   workspace_id,
   model_id,
+  application_id,
   data
 ) => {
   const prefix = (window.MaxKB?.prefix ? window.MaxKB?.prefix : '/admin') + '/api'
-  return postStream(`${prefix}/workspace/${workspace_id}/application/model/${model_id}/prompt_generate`, data)
+  return postStream(`${prefix}/workspace/${workspace_id}/application/${application_id}/model/${model_id}/prompt_generate`, data)
 }
 
 
--- a/ui/src/views/application/component/GeneratePromptDialog.vue
+++ b/ui/src/views/application/component/GeneratePromptDialog.vue
@@ -85,6 +85,7 @@ const chatMessages = ref<Array<any>>([])
 // 原始输入
 const originalUserInput = ref<string>('')
 const modelID = ref('')
+const applicationID = ref('')
 const dialogVisible = ref(false)
 const inputValue = ref<string>('')
 const loading = ref<boolean>(false)
@@ -97,8 +98,8 @@ const promptTemplates = {
 
 请按以下格式生成：
 
-# 角色: 角色名称
-角色概述和主要职责的一句话描述
+# 角色: 
+
 
 ## 目标：
 角色的工作目标,如果有多目标可以分点列出,但建议更聚焦1-2个目标
@@ -118,9 +119,12 @@ const promptTemplates = {
 
 
 ## 限制：
-描述角色在互动过程中需要遵循的限制条件1
-描述角色在互动过程中需要遵循的限制条件2
-描述角色在互动过程中需要遵循的限制条件3
+1. **严格限制回答范围**：仅回答与角色设定相关的问题。  
+   - 如果用户提问与角色无关，必须使用以下固定格式回复：  
+     “对不起，我只能回答与【角色设定】相关的问题，您的问题不在服务范围内。”  
+   - 不得提供任何与角色设定无关的回答。  
+2. 描述角色在互动过程中需要遵循的限制条件2
+3. 描述角色在互动过程中需要遵循的限制条件3
   `,
 }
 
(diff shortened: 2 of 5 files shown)


ui/src/views/application/component/GeneratePromptDialog.vue
```diff
@@ -226,8 +230,9 @@ const stopChat = () => {
   chatMessages.value = []
 }
 
-const open = (modelId: string) => {
+const open = (modelId: string, applicationId: string) => {
   modelID.value = modelId
+  applicationID.value = applicationId
   dialogVisible.value = true
   originalUserInput.value = ''
   chatMessages.value = []
```
shaohuzhang1, line 238: There are no identified irregularities, potential issues, or optimization suggestions in the provided code snippet. The changes introduced to include `applicationID` and ensure it is used appropriately within the `generatePrompt` function are minor additions that do not alter the overall functionality of the component.

ui/src/api/application/application.ts
```diff
@@ -206,16 +206,21 @@ const open: (application_id: string, loading?: Ref<boolean>) => Promise<Result<s
...
+const generate_prompt: (workspace_id:string ,model_id:string, application_id:string,data: any) => Promise<any> = (
   workspace_id,
   model_id,
+  application_id,
   data
 ) => {
   const prefix = (window.MaxKB?.prefix ? window.MaxKB?.prefix : '/admin') + '/api'
-  return postStream(`${prefix}/workspace/${workspace_id}/application/model/${model_id}/prompt_generate`, data)
+  return postStream(`${prefix}/workspace/${workspace_id}/application/${application_id}/model/${model_id}/prompt_generate`, data)
 }
```
shaohuzhang1, line 226: The provided code snippet has a couple of issues:

1. **Incorrect URL Construction**: The original line constructs the URL in the incorrect form.
   ```javascript
   return postStream(`${prefix}/workspace/${workspace_id}/application/model/${model_id}/prompt_generate`, data)
   ```
   Should be:
   ```javascript
   return postStream(`${prefix}/workspace/${workspace_id}/application/${application_id}/model/${model_id}/prompt_generate`, data)
   ```

2. **Missing Return Type Annotation for `generate_prompt`**: Although typescript annotations are optional, it's considered good practice to include them for clarity.

3. **Unnecessary Use of Window Object**: Assuming `window.MaxKB?.prefix` is defined elsewhere, using `WindowGlobalScope` might be more appropriate depending on how you want this logic scoped.

Here is corrected and improved version:

```typescript
import { Ref } from 'vue';
import { Result } from './your_module_results'; // Adjust with actual import path

interface MaxKB {
  prefix?: string;
}

/**
 * Creates an open promise based on an application ID.
 */
export const open: (application_id: string, loading?: Ref<boolean>) => Promise<Result<string>> = async(application_id, loading): Promise<Result<string>> => {
  if (!loading) throw new Error('Loading must be supplied');
  
  const prefix = (globalThis as Window & typeof global).MaxKB.prefix ?? '/admin';
  try {
    if (loading) loading.value = true; // Assuming loading.value is reactive
  
    const response = await fetch(`${prefix}/admin/api/workspace/:${id}/app/apply_model/promptGenert`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
      },
      body: JSON.stringify(data),
    });
    
    if (!response.ok) {
      console.error(`HTTP error! status: ${response.status}`);
      throw new Error(response.statusText);
    }
    
    const result = await response.json();
    if (!result.success) {
      console.warn(result.message || 'Failed to generate prompt.');
      throw new Error('Prompt generation failed.');
    }

    return result.data;
    
  } catch (err) {
    console.error(err);
    throw err;
  } finally {
    if (loading) loading.value = false;
  }
};

/**
 * Generates optimization prompts.
 * @param workspaceId
 * @param modelId
 * @param appId
 * @param data Any additional request parameters or payload data
 * @returns Response indicating success or failure along with generated prompt data.
 */
async function generate_prompt(workspaceId: string, modelId: string, appId: string, data: any) : Promise<any> {
  const prefix = (globalThis as Window & typeof global).MaxKB.prefix ?? '/admin'
  
  try {
    let res = await fetch(`${prefix}/workspace/${workspaceId}/application/${appId}/model/${modelId}/prompt_generate`, {
      method: "post",
      headers: {
        'Content-Type': 'application/json',
      },
      body: JSON.stringify(data)
    });

  if(!res.ok){
    console.error(`HTTP error! status: ${res.status}`);}
        
    const jsonResult = await res.json()
    
    if(jsonResult && !jsonResult.success){
       console.warn(jsonResult.message || 'Failed to generate prompt.')
       throw new Error('Prompt generation had a problem.')
    }
      
    return jsonResult.promptData;
   
  }catch(errorMessage){
    console.log(errorMessage);

    throw errorMessage;}
}
```
Note that I also used `fetch()` instead of `postStream()` which is a hypothetical utility function you may need to define to match your implementation details. Also ensure all imports like `Ref`, `Result`, etc., should be correctly resolved by their respective module names.

ui/src/workflow/nodes/intent-classify-node/index.ts
```diff
@@ -6,8 +6,8 @@ class IntentNode extends AppNode {
   }
 }
 
-const get_up_index_height = (branch_lsit: Array<any>, index: number) => {
-    return branch_lsit
+const get_up_index_height = (branch_list: Array<any>, index: number) => {
+    return branch_list
         .filter((item, i) => i < index)
         .map((item) => item.height + 8)
         .reduce((x,y) => x+y, 0)
```
shaohuzhang1, line 13: The function `get_up_index_height` looks correct for calculating the height of ancestors up to a given index in an array. However, there is a minor typo that can be corrected:

```javascript
-const get_up_index_height = (branch_lsit: Array<any>, index: number) => {
+const get_up_index_height = (branch_list: Array<any>, index: number) => {
```

This should have no effect on its functionality.

**Optimization Suggestions**: There isn't much that could significantly optimize this function in terms of performance with the current implementation. The operations involved—filtering, mapping, and reducing—are already quite efficient for small arrays. If you anticipate dealing with very large datasets, consider parallelizing some parts of the code if it's applicable, although in most cases JavaScript engines will handle such tasks internally efficiently. Ensure that all other functions used within `get_up_index_height` also perform well and avoid unnecessary computations.

PR #4033, opened 2025-09-15
refactor: Generate prompt serializer

shaohuzhang1:
refactor: Generate prompt serializer

### DIFF
--- a/apps/chat/serializers/chat.py
+++ b/apps/chat/serializers/chat.py
@@ -146,6 +146,15 @@ class PromptGenerateSerializer(serializers.Serializer):
     model_id = serializers.CharField(required=False, allow_blank=True, allow_null=True, label=_("Model"))
     application_id = serializers.CharField(required=False, allow_blank=True, allow_null=True, label=_("Application"))
 
+    def is_valid(self, *, raise_exception=False):
+        super().is_valid(raise_exception=True)
+        workspace_id = self.data.get('workspace_id')
+        query_set = QuerySet(Application).filter(id=self.data.get('application_id'))
+        if workspace_id:
+            query_set = query_set.filter(workspace_id=workspace_id)
+        if not query_set.exists():
+            raise AppApiException(500, _('Application id does not exist'))
+
     def generate_prompt(self, instance: dict, with_valid=True):
         if with_valid:
             self.is_valid(raise_exception=True)


apps/chat/serializers/chat.py
```diff
@@ -146,6 +146,15 @@ class PromptGenerateSerializer(serializers.Serializer):
...
+        workspace_id = self.data.get('workspace_id')
+        query_set = QuerySet(Application).filter(id=self.data.get('application_id'))
+        if workspace_id:
+            query_set = query_set.filter(workspace_id=workspace_id)
+        if not query_set.exists():
+            raise AppApiException(500, _('Application id does not exist'))
+
     def generate_prompt(self, instance: dict, with_valid=True):
         if with_valid:
             self.is_valid(raise_exception=True)
```
shaohuzhang1, line 160: The provided Python code snippet includes an `is_valid` method within a Django REST Framework serializer that performs several checks on the incoming data. It validates whether specific fields such as 'model_id' and 'application_id' meet certain conditions before proceeding to call the original validation.

Here's a breakdown of what the code does:

1. **Validation Calls**: It calls `super().is_valid(raise_exception=True)` to perform standard field-level validations defined in the parent class.

2. **Extracting Data Fields**:
   - `instance`: A dictionary containing instance information, which may be relevant for serialization processes.
   - `query_set`: A filtered queryset based on 'application_id', initially filtering by the ID specified in the request data (`self.data.get('application_id')`).
   - `workspace_id`: If present in the request data, it filters the `query_set` to only include records where 'workspace_id' matches the one specified in the data.

3. **Checking Application Existence**:
   - After potentially applying the `workspace_id` filter, it checks if any applications match the criteria using `if query_set.exists()`.
   - If no matching applications are found (`not query_set.exists()`), it raises an API exception with status code 500 and message "Application id does not exist".

### Suggested Improvements and Enhancements:

- **Error Messages and Code Clarity**: While clear, some error messages could benefit from more context or details about what might have gone wrong (e.g., specify the model type or workspace name).
  
- **Pre-Caching Results**: Depending on your use case, you might consider caching results for repeated requests instead of re-querying multiple times during the validation process.

- **Field Validation Beyond Exists Check**: Consider expanding the validation logic to enforce additional constraints on each required field, ensuring they comply with expected types or formats.

Here’s how the modified `generate_prompt` method could look after these considerations:

```python
def generate_prompt(self, instance: dict, with_valid=True):
    if with_valid:

        # Additional custom validation for specific fields can be added here

    return {
        "id": instance.id,
        "prompt_text": prompt_text,
        # other serialized data
    }
```

This refactoring enhances clarity and maintains existing functionality while allowing for flexibility and further customization should needed.

PR #4035, opened 2025-09-15
feat: conditionally render form items based on feature flags

shaohuzhang1:
feat: conditionally render form items based on feature flags

### DIFF
--- a/ui/src/views/application/ApplicationSetting.vue
+++ b/ui/src/views/application/ApplicationSetting.vue
@@ -414,7 +414,7 @@
                     </div>
                   </template>
                 </div>
-                <el-form-item @click.prevent>
+                <el-form-item @click.prevent v-if="applicationForm.mcp_enable || applicationForm.tool_enable">
                   <template #label>
                     <div class="flex-between">
                       <span class="mr-4">
--- a/ui/src/workflow/nodes/ai-chat-node/index.vue
+++ b/ui/src/workflow/nodes/ai-chat-node/index.vue
@@ -230,7 +230,7 @@
             </div>
           </template>
         </div>
-        <el-form-item @click.prevent>
+        <el-form-item @click.prevent v-if="chat_data.mcp_enable || chat_data.tool_enable">
           <template #label>
             <div class="flex-between">
               <span class="mr-4">


ui/src/views/application/ApplicationSetting.vue
```diff
@@ -414,7 +414,7 @@
                     </div>
                   </template>
                 </div>
-                <el-form-item @click.prevent>
+                <el-form-item @click.prevent v-if="applicationForm.mcp_enable || applicationForm.tool_enable">
                   <template #label>
                     <div class="flex-between">
                       <span class="mr-4">
```
shaohuzhang1, line 420: The code appears to be part of an HTML template for an El Form Item. Here are some potential concerns and suggestions:

1. **Conditional Directive (`v-if`)**:
   - The `@click.prevent v-if="applicationForm.mcp_enable || applicationForm.tool_enable"` directive suggests that this form item is only visible depending on certain conditions. This seems logical, but ensure these conditions are appropriate.
   
2. **PreventDefault Event Listener**:
   - Both versions (including and excluding the condition) use `@click.prevent`. It's important to avoid repeating event listeners with the same default behavior if they serve a similar purpose across different scopes.

3. **Element Type**:
   - Ensure that this form item corresponds to an `<el-form-item>` element used within another component. If it isn't needed under certain conditions, removing the conditional could save computation.

4. **Accessibility**:
   - Consider ensuring accessibility compliance when using conditional rendering. For instance, adding text or icons to describe the form item visually hidden from screen readers.

Here is revised version of the line incorporating some of these considerations:
```html
<el-form-item 
  @click.prevent 
  :aria-hidden="!(applicationForm.mcp_enable || applicationForm.tool_enable)"
>
  <template #label>
    <div class="flex-between">
      <span class="mr-4">
        <!-- Label content -->
      </span>
    </div>
  </template>
</el-form-item>
```
This modification ensures better SEO and assistive technology compatibility without duplicating event listeners. Adjust labels and aria attributes as necessary for specific contexts.

ui/src/workflow/nodes/ai-chat-node/index.vue
```diff
@@ -230,7 +230,7 @@
             </div>
           </template>
         </div>
-        <el-form-item @click.prevent>
+        <el-form-item @click.prevent v-if="chat_data.mcp_enable || chat_data.tool_enable">
           <template #label>
             <div class="flex-between">
               <span class="mr-4">
```
shaohuzhang1, line 236: The provided code snippet has the following issues:

1. **Misplacement of `v-if` Directive**: The `<el-form-item>` tag should not have both an `@click.prevent` handler and a conditional directive (`v-if`). Instead, move the `v_if` below the `@click.prevent`.

2. **Incorrect Class Binding Syntax**: In Vue 3, you do not use single quotes around string literals in template expressions unless they contain variables or operators that need escaping.

Here's the corrected version:

```jsx
<el-form-item @click.prevent v-if="chat_data.mcp_enable || chat_data.tool_enable">
  <template #label>
    <div class="flex-between">
      <span class="mr-4">...</span><!-- Fill with actual content -->
   </div>
  </template>
</el-form-item>
```

Note: Ensure that all necessary imports, such as `vue`, `el-form-item`, etc., are correctly included in your project setup if this is being used in a context like Element UI or another framework.

PR #4036, opened 2025-09-15
feat: Extract reason using regex

shaohuzhang1:
feat: Extract reason using regex

### DIFF
--- a/apps/application/flow/step_node/intent_node/impl/base_intent_node.py
+++ b/apps/application/flow/step_node/intent_node/impl/base_intent_node.py
@@ -89,7 +89,6 @@ class BaseIntentNode(IIntentNode):
         try:
             r = chat_model.invoke(message_list)
             classification_result = r.content.strip()
-
             # 解析分类结果获取分支信息
             matched_branch = self.parse_classification_result(classification_result, branch)
 
@@ -101,7 +100,7 @@ class BaseIntentNode(IIntentNode):
                 'history_message': history_message,
                 'user_input': user_input,
                 'branch_id': matched_branch['id'],
-                'reason': json.loads(r.content).get('reason'),
+                'reason': self.parse_result_reason(r.content),
                 'category': matched_branch.get('content', matched_branch['id'])
             }, {}, _write_context=write_context)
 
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/intent_node/impl/base_intent_node.py
```diff
@@ -210,6 +209,26 @@ def get_branch_by_id(category_id: int):
...
+                if match:
+                    reason = match.group(1).strip()
+                    # 清理可能的尾部字符
+                    reason = re.sub(r'["\s]*$', '', reason)
+                    return reason
+
+            return ''
 
     def find_other_branch(self, branch: List[Dict]) -> Dict[str, Any] | None:
         """查找其他分支"""
```
shaohuzhang1, line 234: The provided code has several issues and can be optimized:

### Issues Found:
1. **Inconsistent Handling of JSON Parsing**:
   - In `parse_classification_result`, there's inconsistent handling of parsing errors using `json.loads`. The error case should handle exceptions more robustly.

2. **Duplicate Code Block**:
   - There's a duplicate block of code in `execute` that can be minimized by combining the two branches into a single condition.

3. **Parsing Reason for Classification Results**:
   - The `parse_result_reason` function has multiple regular expression patterns to extract reasons from different formats, which is inefficient due to repeated regex searches on each parse attempt.

4. **Unused Variable `matched_branch_id`**:
   - The variable `matched_branch_id` is not used within the function, so it can be removed for clarity.

5. **Return Values**:
   - The return values seem ambiguous. For example, when no branch matches and normal intents are available, some functions default to `'unknown'` instead of returning a structured dictionary like `{'id': '', 'content': ''}`.

### Optimization Suggestions:
1. **Refactor and Simplify Logic for Classifying Result Reason**:
   ```python
   def parse_result_reason(self, result: str) -> str:
       """解析分类的结果原因"""
       try:
           result_json = json.loads(result)
           return result_json.get('reason', '')
       except Exception as e:
           match = re.search(
               '(?P<delim>["\s]*):(?=\s*\breason\b)(.*?)"reason"', result, re.DOTALL
           )
           if match:
               reason = match.group(1).strip()
               # Clean up possible tailing characters
               reason = re.sub(r'[,"\s]*$', '', reason)
               return reason

           return ''
   ```

2. **Combine Condition in Execute Method**:
   ```python
   matched_branch = self.find_matched_branch(history_message, user_input, branch)

   if not matched_branch or (not matched_branch['enabled'] and matched_branch != 'other'):
       matched_branch = {key: key for key in ('id', 'content')} if other_branch else normal_intents[0]
   ```

These changes will improve the consistency, readability, and efficiency of the code while addressing potential issues such as redundant code blocks and insufficient exception handling.

PR #4045, opened 2025-09-16
feat: Support loop node

shaohuzhang1:


ui/src/workflow/nodes/loop-body-node/LoopBodyContainer.vue
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+  .arrow-icon {
+    transition: 0.2s;
+  }
+}
+:deep(.el-card) {
+  overflow: visible;
+}
+</style>
```
shaohuzhang1, line 222: The code is generally well-structured and follows Vue.js conventions. However, there are several areas that could benefit from improvements or optimizations:

### Improvements

1. **Duplicate Code**: The `copyClick(item.globeLabel)` call appears twice in the template without any reason to do so. Consider extracting this into a separate function or method.

2. **Conditional Styling**: There's conditional styling for errors using inline styles (`background: #fff`). This can be improved by using CSS classes instead when possible.

3. **Element Plus Import**: Although not mentioned in the code snippet, ensure all Element Plus components are fully imported at the top to avoid runtime import errors.

4. **Avoid Magic Numbers**: Replace hardcoded numbers like `200` with named constants for better readability and maintainability.

5. **Optimize Tooltip Positioning**: Since tooltips are positioned relative to their targets, consider calculating positions dynamically if they're always within the viewable area.

6. **TypeScript Type Definitions**: Ensure that TypeScript types for `props`, variables, and methods are correctly defined and consistent across your project.

7. **Code Duplication**: Remove redundant code blocks within loops, especially those that perform similar operations multiple times.

8. **Use of Classes vs. Inline Styles**: If conditions change frequently, it might make sense to use CSS classes based on computed properties rather than inline styles.

### Optimization Suggestions

1. **Reduce DOM Manipulation**: Avoid creating unnecessary elements and manipulating the DOM directly as much as possible. Use reactive state management where feasible.

2. **Lazy Load Components**: For dynamic components (like icons), consider using lazy loading mechanisms to improve initial load times.

3. **Refactor Redundant Logic**: Extract repetitive logic out of functions or mixins to reduce redundancy and increase reusability.

4. **Performance Profiling**: Use browser developer tools to profile and identify performance bottlenecks, potentially leading to further optimizations.

By addressing these points, you can enhance both the functionality and the efficiency of your code.

apps/application/flow/loop_workflow_manage.py
```diff
@@ -0,0 +1,193 @@
...
+        格式化生成提示词
+        @param prompt: 提示词信息
+        @return: 格式化后的提示词
+        """
+
+        context = {**self.get_workflow_content(), **self.parentWorkflowManage.get_workflow_content()}
+        prompt = self.reset_prompt(prompt)
+        prompt_template = PromptTemplate.from_template(prompt, template_format='jinja2')
+        value = prompt_template.format(context=context)
+        return value
```
shaohuzhang1, line 193: Code Review:

- The `NodeResultFuture` class has unnecessary logic for handling execution status and exceptions separately. It can be simplified to directly handle both using Python's exception mechanism.

```python
class NodeResultFuture:
    def __init__(self, result, error):
        self.result = result
        self.error = error

    def unwrap(self):
        if isinstance(self.error, ExecutionError):  # Assuming ExecutionError is a custom exception type
            raise self.error
        return self.result
```

- In `await_result`, consider adding logging statements to track when requests are still open after the specified timeout.

```python
import logging

logger = logging.getLogger(__name__)

def await_result(result, timeout=1):
    try:
        result.result(timeout)
        logger.info(f"Request completed within {timeout} seconds.")
        return False
    except Exception as e:
        logger.warning(f"Request did not complete within {timeout} seconds. Error: {e}")
        return True
```

- For managing asynchronous tasks in a more straightforward manner, use `asyncio`.

```python
import asyncio

executor = ThreadPoolExecutor(max_workers=200)

async def run_task(func, *args, **kwargs):
    with executor.submit(func, *args, **kwargs) as future:
        loop = asyncio.get_running_loop()
        done, pending = await asyncio.wait({future}, timeout=1)
        
        if done:
            yield future.result().unwrap()
        elif pending:
            print("Timeout occurred")
```

- Consider cleaning up unused imports at the top of your file.

```python
from datetime import timedelta
import json
from typing import *
from urllib.parse import urlencode, urlparse
import os
```

Overall, this code provides basic structure for an AI-based workflow management system with threading support. Adding asynchronous capabilities could improve performance significantly, especially when dealing with time-consuming operations or long-running workflows.

ui/src/workflow/nodes/loop-break-node/index.vue
```diff
@@ -0,0 +1,167 @@
...
+  return Promise.all(v_list).catch((err) => {
+    return Promise.reject({ node: props.nodeModel, errMessage: err })
+  })
+}
+onMounted(() => {
+  set(props.nodeModel, 'validate', validate)
+})
+</script>
+
+<style lang="scss" scoped></style>
```
shaohuzhang1, line 167: The provided Vue.js template and script look mostly clean and functional for managing conditional logic in an application workflow. However, a few potential areas for improvement can be made:

1. **Accessibility**: Add ARIA labels to elements that cannot have text content due to the use of `v-if` (like spans with icons). This would enhance accessibility.

2. **Validation**: Ensure that the validation logic correctly handles cases where conditions are removed or added dynamically, especially when using `ref.validate()`.

3. **Performance**: If you encounter performance bottlenecks related to updating multiple fields and arrays frequently, consider optimizing data structures or lazy loading components as needed.

### Detailed Suggestions:

#### Accessibility Improvements:
Add appropriate ARIA labels to elements like `<div>`s, buttons, and icons to improve screen reader compatibility. Here's how you might update some lines to include aria-labels:

```html
<div
  aria-labelledby="condition-list-info"
  class="flex-between lighter"
>
```

and ensure each element has its own ID for easy identification by screen readers:

```html
<span id="condition-list-info">{{ $t('views.applicationWorkflow.nodes.conditionNode.conditions.info') }}</span>

<div class="w-full" aria-placeholder="Variable Selection...">
  <!-- cascader component -->
</div>
<button type="button" class="btn-link btn-sm mt-4" :aria-label="t('Delete Condition')">
  <!-- Delete icon or text -->
</button>
<!-- More similar updates throughout the code base ... -->

<script setup lang="ts">
// Accessible variables here, such as:
const t = useI18n().te ? useI18n() : import('@/i18next').then(i => i.t);
</script>
```

#### Validation Considerations:
Ensure that your condition array is properly validated whenever it changes. Currently, the logic does not handle this explicitly:

```typescript
// On add/delete conditions event
addCondition();
// Or deleteCondition()

// Example of adding dynamic validation after modifying condition list:
onEffect(() => {
  validate(); // Revalidate every time conditions change
}, [form_data]);
```
Make sure that your existing checks within `validate()` function reflect all possible states, including those caused by adding or removing objects from the `condition_list`.

#### Further Optimization:
If there are instances of deep data copying (`cloneDeep`) in key functions, review these sections for unnecessary duplication since they can impact rendering efficiently, but excessive usage should generally be reduced for better maintainability and responsiveness.

These tips will help improve both usability and efficiency of your component.

PR #4052, opened 2025-09-17
fix: Loop node response data

shaohuzhang1:
fix: Loop node response data


apps/application/flow/step_node/loop_node/impl/base_loop_node.py
```diff
@@ -248,12 +248,14 @@ def get_details(self, index: int, **kwargs):
...
             "params": self.context.get('params'),
             'run_time': self.context.get('run_time'),
             'type': self.node.type,
             'current_index': self.context.get("index"),
             "current_item": self.context.get("item"),
-            'loop_type': self.context.get("loop_type"),
+            'loop_type': self.node_params_serializer.data.get('loop_type'),
             'status': self.status,
             'loop_node_data': self.context.get("loop_node_data"),
             'loop_answer_data': self.context.get("loop_answer_data"),
```
shaohuzhang1, line 261: The provided code snippet has several improvements that make it more robust and optimized:

1. **Consistent Naming**: Ensure consistency in naming conventions throughout the method. Use `snake_case` for variable names to improve readability.

2. **Use of `get()` with Default Values**: The use of `self.node_params_serializer.data.get('some_key', '')` is good because it provides a default value if the key does not exist, avoiding potential errors.

3. **Remove Unnecessary Comments**: Remove unnecessary comments like `# comment`: These only clutter the code without adding any useful information.

4. **Optimize Dictionary Lookup**: Since you access `node_params_serializer.data` multiple times, consider creating a local variable at the beginning of the function if this pattern occurs consistently and improves performance due to repeated dictionary accesses.

Here's an optimized version:

```python
    node_data = self.node.params_serializer.data.copy()
    
    return {
        "index": index,
        "result": self.context.get('result'),
        'array': node_data.get('array'),  # Assuming array is needed
        'number': node_data.get('number'),  # Assuming number is needed
        "params": self.context.get('params'),
        'type': self.node.type,
        'current_index': self.context.get("index"),
        "current_item": self.context.get("item"),
        'loop_type': node_data.get('loop_type'),
        'status': self.status,
        'loop_node_data': self.context.get("loop_node_data"),
        'loop_answer_data': self.context.get("loop_answer_data")
    }
```

This version reduces redundancy and makes the code cleaner. It also assumes that `array` and `number` are being used; if they don't need to be present all the time, you can adjust accordingly.

PR #4062, opened 2025-09-19
fix: Improve intent node

shaohuzhang1:
fix: Improve intent node  --bug=1062028 --user=张展玮 【应用编排】意图识别节点没有带上历史聊天记录进行分类 https://www.tapd.cn/62980211/s/1776831

### DIFF
--- a/apps/application/flow/step_node/intent_node/impl/prompt_template.py
+++ b/apps/application/flow/step_node/intent_node/impl/prompt_template.py
@@ -2,29 +2,31 @@
 
 
 
-PROMPT_TEMPLATE = """# Role
-        You are an intention classification expert, good at being able to judge which classification the user's input belongs to.
+PROMPT_TEMPLATE = """
+    # Role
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/intent_node/impl/prompt_template.py
```diff
@@ -2,29 +2,31 @@
...
+    - Strictly ensure that the output is in a valid JSON format.
+    - Do not add prefix ```json or suffix ```
+    - The answer needs to include the following fields such as:
+    {{
+    "classificationId": 0,
+    "reason": ""
+    }}
+
+    ## Limit
+    - Please do not reply in text."""
```
shaohuzhang1, line 32: No significant issues were found in the provided code. However, I suggest adding more detailed instructions about how to interpret and use historical conversations when determining the classification.

**Optimization Suggestions**: 
It would also help if there was a placeholder or a method to extract relevant parts from previous interactions during the conversation (such as through metadata extraction) so that the model can better understand the context when making a classification decision.

PR #4071, opened 2025-09-22
docs: OpenAPI request URL

shaohuzhang1:
docs: OpenAPI request URL

### DIFF
--- a/ui/src/layout/layout-header/avatar/APIKeyDialog.vue
+++ b/ui/src/layout/layout-header/avatar/APIKeyDialog.vue
@@ -87,7 +87,7 @@ const props = defineProps({
 })
 const emit = defineEmits(['addData'])
 
-const apiUrl = window.location.origin + +`${window.MaxKB.prefix}/api-doc/`
+const apiUrl = window.location.origin + `${window.MaxKB.prefix}/api-doc/`
 const SettingAPIKeyDialogRef = ref()
 const dialogVisible = ref<boolean>(false)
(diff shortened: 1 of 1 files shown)


ui/src/layout/layout-header/avatar/APIKeyDialog.vue
```diff
@@ -87,7 +87,7 @@ const props = defineProps({
 })
 const emit = defineEmits(['addData'])
 
-const apiUrl = window.location.origin + +`${window.MaxKB.prefix}/api-doc/`
+const apiUrl = window.location.origin + `${window.MaxKB.prefix}/api-doc/`
 const SettingAPIKeyDialogRef = ref()
 const dialogVisible = ref<boolean>(false)
 const loading = ref(false)
```
shaohuzhang1, line 93: The provided code appears to have a minor error: It's using `+` to concatenate a boolean value (`window.maxKB.prefix === true`) with a string, which can lead to unexpected results. Here is the corrected version:

```javascript
if (window.MaxKB.prefix) {
  apiUrl = window.location.origin + `${window.MaxKB.prefix}/api-doc/`;
}
```

### Optimization Suggestions:
1. **Inline Check**: The condition inside the function call can be moved up into an inline if statement for better readability and performance.

This will ensure that the `apiUrl` variable is only set if `window.MaxKB.prefix` exists and is truthy. This avoids unnecessary computations and improves efficiency.

Issue #4019, opened 2025-09-10
[Bug] Embedded askr parameter does not work


Ruofice:
### MaxKB Version

v1.10.10-lts

### Problem Description

嵌入式版本的asker参数输入{username:user0},提问后在对话日志中实际显示为游客

### Steps to Reproduce

<img width="935" height="966" alt="Image" src="https://github.com/user-attachments/assets/bfb6d44f-06fe-4fe6-a111-4f9cbfec6a3a" />

<img width="2219" height="416" alt="Image" src="https://github.com/user-attachments/assets/fd585d50-6cb4-4732-9393-15fec7494335" />

<img width="1949" height="1060" alt="Image" src="https://github.com/user-attachments/assets/223222cb-2b65-458d-bb58-e7abdc6634ca" />

### Related log output

```shell

```

Shenguobin0102:
感谢反馈，后续版本修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, subsequent version fixes

baixin513:
v1.10.11 已发布

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
v1.10.11 Release

PR #4075, opened 2025-09-22
fix: In the floating window embedding, the user in the conversation is displayed as a tourist #4019

shaohuzhang1:
fix: In the floating window embedding, the user in the conversation is displayed as a tourist #4019

### DIFF
--- a/apps/application/serializers/application_serializers.py
+++ b/apps/application/serializers/application_serializers.py
@@ -322,6 +322,7 @@ class ApplicationSerializer(serializers.Serializer):
 
         def get_query_api_input(self, application, params):
             query = ''
+            is_asker = False
             if application.work_flow is not None:
                 work_flow = application.work_flow
                 if work_flow is not None:
@@ -333,8 +334,10 @@ class ApplicationSerializer(serializers.Serializer):
                             if input_field_list is not None:
                                 for field in input_field_list:
                                     if field['assignment_method'] == 'api_input' and field['variable'] in params:
+                                        if field['variable'] == 'asker':
+                                            is_asker = True
                                         query += f"&{field['variable']}={params[field['variable']]}"
-            if 'asker' in params:
+            if 'asker' in params and not is_asker:
                 query += f"&asker={params.get('asker')}"
             return query
 


apps/application/serializers/application_serializers.py
```diff
@@ -333,8 +334,10 @@ def get_query_api_input(self, application, params):
                             if input_field_list is not None:
                                 for field in input_field_list:
                                     if field['assignment_method'] == 'api_input' and field['variable'] in params:
+                                        if field['variable'] == 'asker':
+                                            is_asker = True
                                         query += f"&{field['variable']}={params[field['variable']]}"
-            if 'asker' in params:
+            if 'asker' in params and not is_asker:
                 query += f"&asker={params.get('asker')}"
             return query
```
shaohuzhang1, line 343: The provided code has a few potential issues and can be optimized:

### Potential Issues:
1. **Variable Scope**: The `is_asker` variable is declared inside the `if application.work_flow is not None:` block but is used outside of it when checking `'asker'`, which might lead to unexpected behavior.

2. **Logic for API Input Variable**: The logic to determine if each input field should include an API request (`'api_input'`) relies solely on the value assigned at runtime, which could cause issues if values change dynamically after initialization.

3. **Redundant Code**: There is redundant handling for `"asker"` in both the inner loop and the outer condition when determining whether to append it to the query parameters.

4. **String Formatting**: Use Python's formatted string literals (f-strings) instead of concatenation with `+`, which improves readability and performance.

### Optimization Suggestions:
1. **Initialize `is_asker` Outside Block**: Initialize `is_asker` before entering the critical section where its usage depends on external conditions.

2. **Avoid Redundant Checks**: Combine checks for `'asker'` into one statement using logical OR (`or`). This eliminates redundancy in appending `"asker"` to the query string.

Here is the revised version of the method:

```python
def get_query_api_input(self, application, params):
    query = ''
    # Initialize is_asker outside the critical section
    is_asker = False

    if application.work_flow is not None:
        work_flow = application.work_flow
        if work_flow is not None:
            input_field_list = [field for field in work_flow.input_fields 
                                if field.assignment_method == 'api_input']
            if input_field_list:
                for field in input_field_list:
                    if field.variable in params:
                        if field.variable == 'asker':
                            is_asker = True
                        
                        query += f"&{field.variable}={params[field.variable]}"
    
    # Append asker only if it wasn't already added due to dynamic value assignment
    if 'asker' in params and not is_asker:
       query += f"&asker={params.get('asker', '')}"

    return query
```

### Explanation of Changes:
1. **Initialization**: Added `is_asker = False` line above any conditional blocks where it will be used.
   
2. **Dynamic Value Handling**: Combined the check for `'asker'` into a single condition within the `for` loop to avoid redundant operations.

3. **F-string Usage**: Replaced manual concatenation with f-string formatting for improved clarity and efficiency. Also ensured that non-existent key retrieval uses `.get()` with a default value of an empty string, preventing exceptions.

PR #4077, opened 2025-09-22
fix: STT model params add when create

shaohuzhang1:
fix: STT model params add when create

### DIFF
--- a/ui/src/views/model/component/CreateModelDialog.vue
+++ b/ui/src/views/model/component/CreateModelDialog.vue
@@ -139,8 +139,7 @@
         <el-empty
           v-else-if="
             base_form_data.model_type === 'RERANKER' ||
-            base_form_data.model_type === 'EMBEDDING' ||
-            base_form_data.model_type === 'STT'
+            base_form_data.model_type === 'EMBEDDING' 
           "
           :description="$t('views.model.tip.emptyMessage2')"
         />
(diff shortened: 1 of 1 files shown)


ui/src/views/model/component/CreateModelDialog.vue
```diff
@@ -149,14 +148,7 @@
...
-              base_form_data.model_type !== 'LLM' &&
-              base_form_data.model_type !== 'IMAGE' &&
-              base_form_data.model_type !== 'TTI' &&
-              base_form_data.model_type !== 'TTV' &&
-              base_form_data.model_type !== 'ITV'
-            "
+            :disabled="!['TTS', 'LLM', 'IMAGE', 'TTI', 'TTV', 'ITV','STT'].includes(base_form_data.model_type)"
           >
             <AppIcon iconName="app-add-outlined" class="mr-4"/> {{ $t('common.add') }}
           </el-button>
```
shaohuzhang1, line 154: The code you provided is generally clean and well-formed, but there are a few small improvements that could be made for clarity and efficiency:

### Improvements

1. **Consistent Conditional Checks**: Ensure consistency in how conditionals are checked. Here, it's good to use the `includes` method consistently.

2. **Remove Unnecessary Disables**: If an action should always be enabled (like adding), it might not need to be disabled under certain conditions.

Here's the updated code with these considerations:

```javascript
<template>
  <div v-if="base_form_data.model_type === 'RERANKER'">
    <!-- Reranker specific content -->
  </div>
  <el-empty v-else-if="['EMBEDDING','STT'].some(type => type === base_form_data.model_type)">
    :description="$t('views.model.tip.emptyMessage2')"
  </el-empty>

  <template v-else>
    <div v-if="hasAddActionAvailable">{{ $t('common.add') }}</div>
    <div v-else>-- Disabled --</div>
  </template>
</template>

<script>
export default {
  data() {
    return {
      hasAddActionAvailable: ['TTS', 'LLM', 'IMAGE', 'TTI', 'TTV', 'ITV','STT'].includes(this.base_form_data.model_type)
    };
  }
};
</script>
```

### Key Changes

- **Single Condition Check with Includes**: Used `.some()` method instead of multiple `||` checks to determine if `base_form_data.model_type` is one of the values related to empty messages (`RERANKER`, `'EMBEDDING',` or `'STT'`).

- **Computed Property**: Introduced a computed property `hasAddActionAvailable` to simplify disabling logic by directly checking against a predetermined list.

This approach maintains readability while optimizing the conditional checking process.

PR #4078, opened 2025-09-22
fix: Incomplete file upload resulted in Exception: 'file_id' error when answering

shaohuzhang1:
fix: Incomplete file upload resulted in Exception: 'file_id' error when answering

### DIFF
--- a/ui/src/components/ai-chat/component/chat-input-operate/index.vue
+++ b/ui/src/components/ai-chat/component/chat-input-operate/index.vue
@@ -163,7 +163,7 @@
           @TouchEnd="TouchEnd"
           :time="recorderTime"
           :start="recorderStatus === 'START'"
-          :disabled="loading"
+          :disabled="localLoading"
         />
         <el-input
           v-else
@@ -199,7 +199,7 @@
             </span>
             <span class="flex align-center" v-else>
               <el-button
-                :disabled="loading"
+                :disabled="localLoading"
                 text
                 @click="startRecording"
                 v-if="recorderStatus === 'STOP'"
@@ -252,7 +252,7 @@
                       }}：{{ getAcceptList().replace(/\./g, '').replace(/,/g, '、').toUpperCase() }}
                     </div>
                   </template>
-                  <el-button text :disabled="checkMaxFilesLimit() || loading" class="mt-4">
+                  <el-button text :disabled="checkMaxFilesLimit() || localLoading" class="mt-4">
                     <el-icon><Paperclip /></el-icon>
                   </el-button>
                 </el-tooltip>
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/component/chat-input-operate/index.vue
```diff
@@ -798,7 +803,7 @@ function sendChatHandle(event?: any) {
   if (!event?.ctrlKey && !event?.shiftKey && !event?.altKey && !event?.metaKey) {
     // 如果没有按下组合键，则会阻止默认事件
     event?.preventDefault()
-    if (!isDisabledChat.value && !props.loading && !event?.isComposing) {
+    if (!isDisabledChat.value && !localLoading.value && !event?.isComposing) {
       if (
         inputValue.value.trim() ||
         uploadImageList.value.length > 0 ||
```
shaohuzhang1, line 809: ### Code Analysis

#### Regularities and Issues
1. **Variable Naming Consistency**: The variable name `uploadLoading` is used inconsistently throughout the codebase without explanation. This might lead to confusion or maintenance errors.
2. **Conditionality on Local Loading**:
   - In many places, conditions like `:disabled="loading"` are replaced with `:disabled="localLoading"`, which should be consistent throughout.
3. **Function Calls in Conditional Statements**: 
   - A few function calls (`getAcceptList()`) seem to depend on state that isn't guaranteed to be up-to-date due to asynchronous operations.

#### Optimization Suggestions

1. **Consistent Variable Naming Convention**:
   - It's recommended to consistently use either `loading` or `localLoading` based on context to avoid naming inconsistencies.

2. **Use of `$nextTick` for Asynchronous State Updates**:
   - If `checkMaxFilesLimit()` depends on dynamic data, consider using Vue.js' `$nextTick` to ensure reactivity when it updates.

3. **Avoid Overwriting State Directly**:
   - Instead of setting the entire object value in `filePromisionDict.value=file.uid=false;`, only update the specific property needed:
     ```javascript
     filePromisionDict.value[file.uid] = false;
     ```

4. **Ensure Proper Dependency Tracking For Asynchronous Operations**:
   - Always make sure dependency tracking is handled correctly in functions dependent on asynchronous operations like uploading files.

Here's a revised version of the key parts with these considerations:

```jsx
// Example revisions

<div>
  <!-- TouchEnd Event -->
  <RecordButton
    @TouchEnd="TouchEnd"
    :time="recorderTime"
    :start="recorderStatus === 'START'"
    :disabled="localLoading"
  />

  <!-- Else Block -->
  <el-input v-else></el-input>

  <!-- Recording Button Logic (Start/Stop) -->
  <el-button
    :disabled="(recorderStatus !== 'STOP') ? loading || localLoading : true"
    text
    @click="startRecording"
    v-if="recorderStatus === 'STOP'"
  ></el-button>

  <!-- File Upload Template -->
  ...
</div>

<!-- Computed Properties Revised -->
const chatId_context = computed(() => {
  return props.chatId
})

const uploadLoading = computed(() => {
  return Object.values(filePromisionDict.value).some(value => !value);
})

...

// Upload File Function Updated
async function uploadFile(file, fileList) {
  // Existing logic but modified for better consistency

  // After successful upload
  filePromisionDict.value[file.uid] = true;  
}
```
These changes aim to improve clarity, maintainability, and responsiveness of the component.

PR #4083, opened 2025-09-23
fix: Wireless cyclic index remains unchanged

shaohuzhang1:
fix: Wireless cyclic index remains unchanged

### DIFF
--- a/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
+++ b/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
@@ -115,6 +115,7 @@ def generate_while_loop(current_index: int):
     index = current_index
     while True:
         yield index, index
+        index += 1
 
 
 def loop(workflow_manage_new_instance, node: INode, generate_loop):


apps/application/flow/step_node/loop_node/impl/base_loop_node.py
```diff
@@ -115,6 +115,7 @@ def generate_while_loop(current_index: int):
     index = current_index
     while True:
         yield index, index
+        index += 1
 
 
 def loop(workflow_manage_new_instance, node: INode, generate_loop):
```
shaohuzhang1, line 121: The `generate_while_loop` function is working correctly as it generates an infinite sequence of tuples starting from the given `current_index`. However, there are a couple of minor improvements and suggestions:

1. **Variable Naming**: The variable name 'index' is used twice within the function without distinction in their meanings. Consider renaming one of them to avoid confusion.

2. **Docstring**: It’s good practice to include a docstring at the top of functions that explain what they do, their parameters, return value, and if applicable, any assumptions or caveats with the implementation.

Here's the modified code with these suggestions:

```python
@@ -115,8 +115,9 @@ def generate_whiled_loop(current_index: int):
     # Use the first 'index' for clarity
     start_index = current_index

     # Generate an infinite loop
     while True:
         yield start_index, start_index
         start_index += 1


def loop(workflow_manage_new_instance, node: INode, generate_while_loop):
```

**Updated Docstrings (if needed)**:

For example,

```python
def generate_while_loop(current_index: int) -> Iterable[Tuple[int, int]]:
    """Generate an infinite sequence of (index, index) pairs starting from the current index."""
    start_index = current_index

    while True:
        yield start_index, start_index
        start_index += 1


def loop(
    workflow_manage_new_instance: WorkflowManageNewInstance,
    node: INode,
    generate_while_loop: Callable[[int], Iterable[Tuple[int, int]]],
) -> None:
    """Loop through nodes using the provided generator function."""
    # Implement logic using the generated while-loop iterator
    pass
```

These changes enhance the readability and maintainability of your code.

PR #4085, opened 2025-09-23
fix: Intent node i18n

shaohuzhang1:
fix: STT model params add when create<br>fix: Intent node i18n  --bug=1062041 --user=张展玮 【应用编排】意图识别的输入下拉框必填提示没有国际化 https://www.tapd.cn/62980211/s/1777709

### DIFF
--- a/ui/src/workflow/nodes/intent-classify-node/index.vue
+++ b/ui/src/workflow/nodes/intent-classify-node/index.vue
@@ -55,8 +55,9 @@
           prop="content_list"
           :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
           :rules="{
-            required: true,
+            message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
             trigger: 'change',
+            required: true,
           }"
         >
           <template #label>
--- a/ui/src/views/model/component/CreateModelDialog.vue
+++ b/ui/src/views/model/component/CreateModelDialog.vue
@@ -139,8 +139,7 @@
         <el-empty
           v-else-if="
             base_form_data.model_type === 'RERANKER' ||
-            base_form_data.model_type === 'EMBEDDING' ||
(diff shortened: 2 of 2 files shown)


ui/src/workflow/nodes/intent-classify-node/index.vue
```diff
@@ -55,8 +55,9 @@
           prop="content_list"
           :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
           :rules="{
-            required: true,
+            message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
             trigger: 'change',
+            required: true,
           }"
         >
           <template #label>
```
shaohuzhang1, line 63: It looks like you're trying to define a validation rule for an input field within a Vue component. However, there seems to be an issue with the order and structure of the rules:

```javascript
<el-form-item
  prop="content_list"
  :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
  :rules="{
    required: true,
    message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
    trigger: 'change'
  }"
>
```

The `required` property is repeated twice, which can lead to unexpected behavior. If you want both conditions (`message` and `required`) to apply, you should only include one instance of the `required` property or use separate properties if necessary.

Additionally, ensure that `$t('views.applicationWorkflow.nodes.intentNode.input.label')` returns the correct text label and that `$t('views.applicationWorkflow.nodes.textToSpeechNode.content.label')` correctly retrieves the error message. Here's a corrected version:

```javascript
<el-form-item
  prop="content_list"
  :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
  :rules="[
    { required: true, message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'), trigger: 'change' }
  ]"
>
```

This way, the form item will require an entry and display a custom error message when it fails validation.

PR #4086, opened 2025-09-23
fix: Intent node i18n

shaohuzhang1:
fix: Intent node i18n  --bug=1062041 --user=张展玮 【应用编排】意图识别的输入下拉框必填提示没有国际化 https://www.tapd.cn/62980211/s/1777709

### DIFF
--- a/ui/src/workflow/nodes/intent-classify-node/index.vue
+++ b/ui/src/workflow/nodes/intent-classify-node/index.vue
@@ -55,8 +55,9 @@
           prop="content_list"
           :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
           :rules="{
-            required: true,
+            message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
             trigger: 'change',
+            required: true,
           }"
         >
           <template #label>


ui/src/workflow/nodes/intent-classify-node/index.vue
```diff
@@ -55,8 +55,9 @@
           prop="content_list"
           :label="$t('views.applicationWorkflow.nodes.intentNode.input.label')"
           :rules="{
-            required: true,
+            message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
             trigger: 'change',
+            required: true,
           }"
         >
           <template #label>
```
shaohuzhang1, line 63: In the provided code snippet, there is an issue with duplicate rules defined for the `content_list` field. Specifically, you have two instances of the same rule:
```javascript
required: true,
```
This will cause validation errors because rules should be unique within a component's `rules` object.

Additionally, the order of these lines might not meet your expected logic. If both fields need to be validated and are required, they should likely share the same validation rule.

Here is improved version of the rule definition:

```vue
:rules="{
  content_list: [
    {
      message: $t('views.applicationWorkflow.nodes.textToSpeechNode.content.label'),
      trigger: 'change'
    },
    { required: true }
  ]
}"
```

### Explanation:
- Combined two identical validations into one array under `content_list`.
- Used `{}` instead of single quotes around the array to avoid syntax errors in Vue templates.

These changes ensure that the form checks the value change on `content_list`, and if it’s empty (not valid according to `$t('views.applicationWorkflow.nodes.textToSpeechNode.content.label')`), then also considers it as invalid due to `required`: true`.

PR #4089, opened 2025-09-23
fix: Infinite loop increases maximum loop count limit

shaohuzhang1:
fix: Infinite loop increases maximum loop count limit

### DIFF
--- a/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
+++ b/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
@@ -18,7 +18,7 @@ from common.handle.impl.response.loop_to_response import LoopToResponse
 from maxkb.const import CONFIG
 from django.utils.translation import gettext as _
 
-max_loop_count = (CONFIG.get("MAX_LOOP_COUNT") or 1000)
+max_loop_count = int((CONFIG.get("MAX_LOOP_COUNT") or 1000))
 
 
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/loop_node/impl/base_loop_node.py
```diff
@@ -18,7 +18,7 @@
 from maxkb.const import CONFIG
 from django.utils.translation import gettext as _
 
-max_loop_count = (CONFIG.get("MAX_LOOP_COUNT") or 1000)
+max_loop_count = int((CONFIG.get("MAX_LOOP_COUNT") or 1000))
 
 
 def _is_interrupt_exec(node, node_variable: Dict, workflow_variable: Dict):
```
shaohuzhang1, line 24: Your function `_is_interrupt_exec` has a few minor improvements:

1. **Consistent Type Hinting**: You added `from collections import Dict`, but since Python supports type hints natively with dictionaries like `Dict[key_type, value_type]`, you can remove this line.

2. **Conversion to Integer**: The conversion of `CONFIG.get("MAX_LOOP_COUNT")` to an integer is already done using `(int())`. This seems redundant and can be removed if the default value (`1000`) is also converted to an integer implicitly in Python.

Here's the revised code:

```python

max_loop_count = int(CONFIG.get("MAX_LOOP_COUNT", 1000))


def _is_interrupt_exec(node, node_variable: dict, workflow_variable: dict):
```

These changes improve readability slightly without introducing unnecessary complexity.

PR #4100, opened 2025-09-24
feat: Loop body supports zooming in and out

shaohuzhang1:
feat: Loop body supports zooming in and out

### DIFF
--- a/ui/src/workflow/nodes/loop-body-node/LoopBodyContainer.vue
+++ b/ui/src/workflow/nodes/loop-body-node/LoopBodyContainer.vue
@@ -202,7 +202,7 @@ function enlargeHandle() {
     props.nodeModel.graphModel.transformModel.focusOn(
       props.nodeModel.x,
       props.nodeModel.y,
-      props.nodeModel.width + 470,
+      props.nodeModel.width + window.innerWidth - props.nodeModel.width,
       props.nodeModel.height - 30,
     )
     height.value =


ui/src/workflow/nodes/loop-body-node/LoopBodyContainer.vue
```diff
@@ -202,7 +202,7 @@ function enlargeHandle() {
     props.nodeModel.graphModel.transformModel.focusOn(
       props.nodeModel.x,
       props.nodeModel.y,
-      props.nodeModel.width + 470,
+      props.nodeModel.width + window.innerWidth - props.nodeModel.width,
       props.nodeModel.height - 30,
     )
     height.value =
```
shaohuzhang1, line 208: The provided code snippet is attempting to adjust a canvas's focus within an SVG node based on the current model properties. Here are some points of consideration:

1. **Syntax Error**: There's a syntax error where `height.value` is not properly closed with quotes or semicolons.

2. **Window Width Calculation**: The expression to calculate `outerWidth` seems incorrect. It should use `window.outerWidth` instead of just `window.innerWidth`.

3. **Logic Review**:
   - Adding `props.nodeModel.width + window.outerWidth - props.nodeModel.width` will always result in `window.outerWidth`.
   - The calculation `(props.nodeModel.x, props.nodeModel.y)` does nothing because it has no side effects.

Here's how you can fix these issues and optimize the code slightly:

```javascript
function enlargeHandle() {
  // Corrected calculation assuming outerWidth is intended
  const newCenterX = (props.nodeModel.x + window.outerWidth / 2);

  // Use focusOn method correctly
  props.nodeModel.graphModel.transformModel.focusOn(
    newCenterX,
    props.nodeModel.y,
    props.nodeModel.width * 2, // Assuming double size for enlargement
    props.nodeModel.height - 30,
  );

  // Example correction for height assignment (assuming it was supposed to be outside)
  let newHeight;
  if (!isNaN(props.nodeModel.height)) { // Check if height is valid number before using
    newHeight = props.nodeModel.height; // Assign original height back or whatever needs to be done here
  } else {
    newHeight = 50; // Default value if needed
  }
}
```

### Key Improvements:
- Adjusting the center calculation to take into account the full width of the viewport.
- Ensuring proper closing of strings and syntax errors.
- Simplifying unnecessary operations and ensuring logical clarity.

PR #4101, opened 2025-09-24
feat: Incorrect assignment of session variables in loop

shaohuzhang1:
feat: Incorrect assignment of session variables in loop

### DIFF
--- a/apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
+++ b/apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
@@ -80,7 +80,7 @@ class BaseVariableAssignNode(IVariableAssignNode):
             from application.flow.loop_workflow_manage import LoopWorkflowManage
             if isinstance(self.workflow_manage, LoopWorkflowManage):
                 self.workflow_manage.parentWorkflowManage.get_chat_info().set_chat_variable(
-                    self.workflow_manage.chat_context)
+                    self.workflow_manage.parentWorkflowManage.chat_context)
             else:
                 self.workflow_manage.get_chat_info().set_chat_variable(self.workflow_manage.chat_context)
         return NodeResult({'variable_list': variable_list, 'result_list': result_list}, {})


apps/application/flow/step_node/variable_assign_node/impl/base_variable_assign_node.py
```diff
@@ -80,7 +80,7 @@ def execute(self, variable_list, stream, **kwargs) -> NodeResult:
             from application.flow.loop_workflow_manage import LoopWorkflowManage
             if isinstance(self.workflow_manage, LoopWorkflowManage):
                 self.workflow_manage.parentWorkflowManage.get_chat_info().set_chat_variable(
-                    self.workflow_manage.chat_context)
+                    self.workflow_manage.parentWorkflowManage.chat_context)
             else:
                 self.workflow_manage.get_chat_info().set_chat_variable(self.workflow_manage.chat_context)
         return NodeResult({'variable_list': variable_list, 'result_list': result_list}, {})
```
shaohuzhang1, line 86: The code appears to modify the chat context of a workflow based on whether `workflow_manage` is an instance of `LoopWorkflowManage`. There might be improvements and potential checks:

1. **Type Checking**: The type check for `if isinstance(self.workflow_manager, LoopWorkflowManage)` could benefit from including additional types (e.g., subclassing) or adding more specific conditions.

2. **Default Case Handling**: If `self.workflow_manage` is not an instance of either `LoopWorkflowManager` or another expected class, it doesn't explicitly handle this scenario. Consider logging a warning/error message with some kind of default behavior.

3. **Potential Null or Undefined Behavior**: Ensure that `chat_context` is defined and accessible before attempting to set it. This can lead to undefined errors if access happens after object initialization and de-allocation.

Here's the revised version with these suggestions:

```python
def execute(self, variable_list, stream, **kwargs) -> NodeResult:
    from application.flow.loop_workflow_manage import LoopWorkflowManage
    
    # Check if workflow_manage is a subclass of LoopWorkflowManage
    loop_manage_type_check = isinstance(self.workflow_manage, (LoopWorkflowManage,))
    parent_manage_type_check = False
    
    # Further checking if self.workflow_manage is an instance of its superclass(es)
    if isinstance(self.workflow_manage, (
            LoopWorkflowManage,
            MyOtherExpectedClass,
        )
    ):
        parent_manage_type_check = True

    if (loop_manage_type_check or parent_manage_type_check):
        assert hasattr(self.workflow_manage, 'parentWorkflowManage'), \
               "Missing parentWorkflowManage attribute."
        if hasattr(self.workflow_manage.parentWorkflowManage, 'get_chat_info'):
            workflow_manager_chat_info = self.workflow_manage.parentWorkflowManage.get_chat_info()
            if hasattr(workflow_manager_chat_info, 'set_chat_variable'):
                try:
                    workflow_manager_chat_info.set_chat_variable(
                        self.workflow_manage.chat_context
                    )
                except Exception as e:
                    print(f"Error setting chat context: {str(e)}")
            else:
                print("No valid 'set_chat_variable' method in workflow manager chat info.")
        else:
            print("No valid 'get_chat_info' method implemented.")
    elif hasattr(self.workflow_manager, 'get_chat_info'):
        workflow_manager_chat_info = self.workflow_manager.get_chat_info()
        
        if hasattr(workflow_manager_chat_info, 'set_chat_variable'):
            workflow_manager_chat_info.set_chat_variable(self.workflow_manage.chat_context)
        else:
            print("No valid 'set_chat_variable' method in workflow manager chat info.")
    else:
        print("Incomplete or missing configuration.")

    return NodeResult({'variable_list': variable_list, 'result_list': result_list}, {})
```

In summary, we added type checking, included fallback handling, ensured proper existence of attributes, and logged exceptions/errors where applicable. This should help prevent runtime bugs and provide better error reporting during execution.

PR #4102, opened 2025-09-24
style: Highly optimized when adding BREAK and Continue nodes

shaohuzhang1:
style: Highly optimized when adding BREAK and Continue nodes

### DIFF
--- a/ui/src/workflow/common/data.ts
+++ b/ui/src/workflow/common/data.ts
@@ -478,7 +478,7 @@ export const loopContinueNode = {
   type: WorkflowType.LoopContinueNode,
   text: t('views.applicationWorkflow.nodes.loopContinueNode.text'),
   label: t('views.applicationWorkflow.nodes.loopContinueNode.label'),
-  height: 600,
+  height: 100,
   properties: {
     width: 600,
     stepName: t('views.applicationWorkflow.nodes.loopContinueNode.label'),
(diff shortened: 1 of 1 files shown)


ui/src/workflow/common/data.ts
```diff
@@ -510,7 +510,7 @@ export const loopBreakNode = {
   type: WorkflowType.LoopBreakNode,
   text: t('views.applicationWorkflow.nodes.loopBreakNode.text'),
   label: t('views.applicationWorkflow.nodes.loopBreakNode.label'),
-  height: 600,
+  height: 100,
   properties: {
     width: 600,
     stepName: t('views.applicationWorkflow.nodes.loopBreakNode.label'),
```
shaohuzhang1, line 516: The code has an unnecessary change that reduces the height of both `loopContinueNode` and `loopBreakNode`. This might affect their visual representation on the workflow diagram. If you want to reduce the size of these nodes, consider adjusting their width or position instead.

Optimization suggestion:
If reducing the height is intended to improve performance or simplify rendering on mobile devices, ensure that this change does not negatively impact user interaction or understanding of the flowchart design. Consider alternative solutions for smaller node dimensions if necessary.

Note that changes made after September 1, 2021, should be re-evaluated against current best practices and standards for workflow development tools.

PR #4107, opened 2025-09-25
fix: The restriction on simple application questioning is not effective

shaohuzhang1:
fix: The restriction on simple application questioning is not effective

### DIFF
--- a/apps/application/serializers/common.py
+++ b/apps/application/serializers/common.py
@@ -154,7 +154,7 @@ class ChatInfo:
             'search_mode': self.application.knowledge_setting.get('search_mode') or 'embedding',
             'no_references_setting': self.get_no_references_setting(self.application.knowledge_setting, model_setting),
             'workspace_id': self.application.workspace_id,
-            'application_id': self.application.id,
+            'application_id': self.application_id,
             'mcp_enable': self.application.mcp_enable,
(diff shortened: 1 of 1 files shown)


apps/application/serializers/common.py
```diff
@@ -154,7 +154,7 @@ def to_base_pipeline_manage_params(self):
             'search_mode': self.application.knowledge_setting.get('search_mode') or 'embedding',
             'no_references_setting': self.get_no_references_setting(self.application.knowledge_setting, model_setting),
             'workspace_id': self.application.workspace_id,
-            'application_id': self.application.id,
+            'application_id': self.application_id,
             'mcp_enable': self.application.mcp_enable,
             'mcp_tool_ids': self.application.mcp_tool_ids,
             'mcp_servers': self.application.mcp_servers,
```
shaohuzhang1, line 160: There are no immediate issues with the provided code snippet. It seems correct syntactically and logically. However, if "self.application_id" is a method that returns an ID but it's missing from the provided context (i.e., its implementation), there might be an error when calling this method.

As far as optimizations go:
- Ensure that all variables referenced have been defined before they are used to avoid NameError.
- If you're using Python 3, consider changing `raw_input()` to `input()`, which reads input strings from the user.

In summary, there are no major flaws but ensure your dependencies work properly.

PR #4110, opened 2025-09-25
fix: OpenAPI authentication

shaohuzhang1:
fix: OpenAPI authentication

### DIFF
--- a/apps/common/middleware/doc_headers_middleware.py
+++ b/apps/common/middleware/doc_headers_middleware.py
@@ -115,7 +115,8 @@ content = """
 
 class DocHeadersMiddleware(MiddlewareMixin):
     def process_response(self, request, response):
-        if request.path.startswith('/doc/') or request.path.startswith('/doc_chat/'):
+        if request.path.startswith(CONFIG.get_admin_path() + '/api-doc/') or request.path.startswith(
+                CONFIG.get_chat_path() + '/api-doc/'):
             auth = request.COOKIES.get('Authorization')
             if auth is None:
                 return HttpResponse(content)


apps/common/middleware/doc_headers_middleware.py
```diff
@@ -115,7 +115,8 @@
 
 class DocHeadersMiddleware(MiddlewareMixin):
     def process_response(self, request, response):
-        if request.path.startswith('/doc/') or request.path.startswith('/doc_chat/'):
+        if request.path.startswith(CONFIG.get_admin_path() + '/api-doc/') or request.path.startswith(
+                CONFIG.get_chat_path() + '/api-doc/'):
             auth = request.COOKIES.get('Authorization')
             if auth is None:
                 return HttpResponse(content)
```
shaohuzhang1, line 122: The provided code snippet has a logical error in that it does not check for authentication when the path starts with `/doc-chat/`. This can lead to unauthorized access for this route. Additionally, the function `CONFIG.get_admin_path()` and `CONFIG.get_chat_path()` should ideally be called within the scope of where they are defined or imported.

To correct these issues:

1. Add an additional condition to check for the presence of authentication on paths starting with '/doc_chat/'.
```python
if request.path.startswith(CONFIG.get_admin_path() + '/api-doc/') or \
   request.path.startswith(CONFIG.get_chat_path() + '/api-doc/') or \
   (request.path.startswith('/doc_chat/') and auth is None):
```
2. Ensure that `CONFIG` is properly initialized before being used in `process_response`.
3. Consider adding logging statements to help debug potential issues.

PR #4116, opened 2025-09-26
fix: Drift of the loop body during each refresh in the canvas

shaohuzhang1:
fix: Drift of the loop body during each refresh in the canvas

### DIFF
--- a/ui/src/workflow/nodes/loop-node/index.vue
+++ b/ui/src/workflow/nodes/loop-node/index.vue
@@ -141,7 +141,7 @@ const mountLoopBodyNode = () => {
     }
     if (props.nodeModel.properties.node_data.loop) {
       x = props.nodeModel.properties.node_data.loop.x
-      y = props.nodeModel.properties.node_data.loop.y - 330
+      y = props.nodeModel.properties.node_data.loop.y
     }
     const nodeModel = props.nodeModel.graphModel.addNode({
       type: loopBodyNode.type,


ui/src/workflow/nodes/loop-node/index.vue
```diff
@@ -141,7 +141,7 @@ const mountLoopBodyNode = () => {
     }
     if (props.nodeModel.properties.node_data.loop) {
       x = props.nodeModel.properties.node_data.loop.x
-      y = props.nodeModel.properties.node_data.loop.y - 330
+      y = props.nodeModel.properties.node_data.loop.y
     }
     const nodeModel = props.nodeModel.graphModel.addNode({
       type: loopBodyNode.type,
```
shaohuzhang1, line 147: The provided code has a few minor improvements that can be made:

1. **Consistent Variable Declarations**: The variable `type` should match the property name in the object being passed to `graphModel.addNode()`. In this case, it is defined as `loopBodyNode.type`, but later used with `type`.

2. **Redundant Assignment**: The line `y = props.nodeModel.properties.node_data.loop.y - 330;` will always result in the same value because `y` is not reassigned anywhere else within the function.

Here's the revised version:

```javascript
const mountLoopBodyNode = () => {
  if (props.nodeModel.properties.node_data.loop) {
    const { x, y } = props.nodeModel.properties.node_data.loop;
  } else {
    // Handle the case when loop data is missing
    x = undefined;
    y = undefined;
  }

  const nodeType = loopBodyNode.type;

  const nodeModel = props.nodeModel.graphModel.addNode({
    type: nodeType,
    x,
    y,
  });
};
```

Additionally, you might want to consider adding comments or error handling for cases where `loopData` might be `null` or `undefined`.

PR #4118, opened 2025-09-26
fix: Application creation time in reverse order

shaohuzhang1:
fix: Application creation time in reverse order

### DIFF
--- a/apps/application/serializers/application.py
+++ b/apps/application/serializers/application.py
@@ -345,7 +345,7 @@ class Query(serializers.Serializer):
         if create_user is not None:
             application_query_set = application_query_set.filter(user_id=create_user)
         application_custom_sql_query_set = application_query_set
-        application_query_set = application_query_set.order_by("-update_time")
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application.py
```diff
@@ -345,7 +345,7 @@ def get_query_set(self, instance: Dict, workspace_manage: bool, is_x_pack_ee: bo
         if create_user is not None:
             application_query_set = application_query_set.filter(user_id=create_user)
         application_custom_sql_query_set = application_query_set
-        application_query_set = application_query_set.order_by("-update_time")
+        application_query_set = application_query_set.order_by("-create_time")
 
         return {'folder_query_set': folder_query_set,
                 'application_query_set': application_query_set,
```
shaohuzhang1, line 351: The code has an issue with the ordering. The order should be `-update_time` instead of `-create_time`. Therefore, change this line:

```python
application_query_set = application_query_set.order_by("-create_time")
```

to

```python
application_query_set = application_query_set.order_by("-update_time")
```

PR #4122, opened 2025-09-26
fix: [Application] An application that adds interface parameters, clicks on the application list page to initiate a conversation, but does not carry interface parameters

shaohuzhang1:
fix: [Application] An application that adds interface parameters, clicks on the application list page to initiate a conversation, but does not carry interface parameters

### DIFF
--- a/ui/src/views/application/index.vue
+++ b/ui/src/views/application/index.vue
@@ -529,35 +529,39 @@ const search_type_change = () => {
   search_form.value = { name: '', create_user: '' }
 }
 
-const apiInputParams = ref([])
-
 function toChat(row: any) {
-  row?.work_flow?.nodes
-    ?.filter((v: any) => v.id === 'base-node')
-    .map((v: any) => {
-      apiInputParams.value = v.properties.api_input_field_list
-        ? v.properties.api_input_field_list.map((v: any) => {
-            return {
-              name: v.variable,
-              value: v.default_value,
-            }
-          })
-        : v.properties.input_field_list
-          ? v.properties.input_field_list
-              .filter((v: any) => v.assignment_method === 'api_input')
-              .map((v: any) => {
-                return {
-                  name: v.variable,
-                  value: v.default_value,
-                }
-              })
-          : []
+  const api =
+    row.type == 'WORK_FLOW'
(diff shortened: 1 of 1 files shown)


ui/src/views/application/index.vue
```diff
@@ -529,35 +529,39 @@ const search_type_change = () => {
...
+    ApplicationApi.getAccessToken(row.id, loading).then((res: any) => {
+      window.open(application.location + res?.data?.access_token + apiParams)
     })
-  const apiParams = mapToUrlParams(apiInputParams.value)
-    ? '?' + mapToUrlParams(apiInputParams.value)
-    : ''
-  ApplicationApi.getAccessToken(row.id, loading).then((res: any) => {
-    window.open(application.location + res?.data?.access_token + apiParams)
   })
 }
```
shaohuzhang1, line 567: The code appears syntactically correct and generally efficient for its purpose. However, there's one potential issue related to asynchronous operations and error handling. The `Promise.all` method is not used in your code, which means that if multiple API calls are made concurrently, you may end up with race conditions.

Additionally, the `loading` variable is referenced within the `.then()` callback, but it doesn't appear to be declared anywhere in the provided snippet. You should ensure that `loading` is properly initialized before being used.

Furthermore, the last line of `else` condition inside the `.then()` block will cause an error because `application.location` might not exist or could lead to unexpected behavior due to the undefined nature of `aips`. It would be better to handle this case more gracefully.

Here’s a revised version of the last part of the code:

```javascript
if (!aips || !aips.length) {
  ApplicationApi.getAccessToken(row.id, loading).then((res: any) => {
    // Assuming application location exists here
    const accessTokenEndpoint = `${application.location}api/auth/access-token`;
    const urlParams = '?';
    window.open(`${accessTokenEndpoint}${urlParams}accesstoken=${encodeURIComponent(res.data.accessToken)}${apiParams}`);
    // Reset input parameters to initial state since no valid inputs were found
    search_form.value = { name: '', create_user: '' };
  });
}
```

Also, make sure that all necessary variables like `search_form`, `row.id`, `ApplicationApi`, and `mapToUrlParams` are defined elsewhere in the code or imported correctly according to their usage context.

PR #4131, opened 2025-09-28
fix: Generate prompt

shaohuzhang1:
fix: Generate prompt


ui/src/views/application/component/GeneratePromptDialog.vue
```diff
@@ -471,6 +482,8 @@ const handleDialogClose = (done: () => void) => {
       .catch(() => {
         // 点击取消
       })
+  } else {
+    done()
   }
 }
```
shaohuzhang1, line 489: The provided code appears to be an Angular component for generating AI role-playing scenarios with text prompts and responses. Here's a summary of the changes and suggestions:

### Changes Summary:
1. **CSS Class Added**: In line 32 (`<div>` element), `mt-8` has been added, which should have no effect since there were only two instances before.

2. **Stream Function Update**: The function `startStreamingOutput` is not defined, so it will result in errors when called. This needs to be implemented or removed based on its intended functionality.

3. **Submit Handler Updates**:
   - Removed duplicate handling of key combinations.
   - Prevented submission if input is empty or if dialog is being paused or streaming.

### Optimization Suggestions:
1. **Remove Unused Variables**: Variable `streamTimer` can likely be moved to local scope within `handleDialogSave` instead of declaring it globally.

2. **Consistent Capitalization and Spacing**: Consider using consistent capitalization and spacing throughout the codebase to enhance readability.

3. **Error Handling Review**: Ensure that all asynchronous operations have proper error handling mechanisms, especially those involving API calls.

4. **Code Refactoring**: If the application logic involves more complex interactions or states, consider refactoring the component to simplify and improve maintainability.

Here's an updated version of the component with some adjustments and comments:

```html
<!-- ... -->
<template>
  <div v-if="answer && !loading && !isStreaming">
    <!-- ... -->
    <el-button type="primary" @click="() => emit('replace', answer)" disabled>Replace</el-button>
    <!-- ... -->
  <div>
  
  <!-- ... -->
  
<script>
// Import other components and plugins here

  name: 'ApplicationGenerateDialog',
  data() {
    return {
      // Existing data properties...
      
      isPaused.value = false; // Initialize pause state
  
      streamTimer: null;
    };
  },
  methods: {
    startStreamingOutput() {
      this.isPaused.value = false;
      this.streamTimer = setInterval(() => {
        if (this.isApiComplete.value && !this.isPaused.value) {
          // Update display content
          const currentAnswer = this.chatMessages.value[this.chatMessages.length - 1];
          if (currentAnswer && currentAnswer.role === 'ai') {
            // Your actual streaming implementation here
          }
        }
      }, 1000); // Adjust interval as needed
    },
    
    handleDialogClose(done: () => void): void {
      this.$confirm($t('views.application.generateDialog.close'), '', {
        confirmButtonText: $t('common.buttons.confirm'),
        cancelButtonText: $t('common.buttons.cancel')
      }).then(async () => {
        await this.saveAndQuit();
        done(true);
      }).catch(err => {
        console.error(err);
      });
    },

    saveAndQuit(): Promise<void> {
      return new Promise((resolve) => {
        // Save logic here
        resolve();
      });
    },
  },
});
</script>

<style scoped lang="less"></style>
```

This revised version includes comments explaining each change and maintains consistency across elements.

apps/chat/serializers/chat.py
```diff
@@ -171,6 +171,9 @@ def generate_prompt(self, instance: dict):
 
         message = messages[-1]['content']
         q = prompt.replace("{userInput}", message)
+        q = q.replace("{application_name}",application.name)
+        q = q.replace("{detail}",application.desc)
+
         messages[-1]['content'] = q
 
         model_exist = QuerySet(Model).filter(
```
shaohuzhang1, line 179: The provided Python code is incomplete and does not include the rest of the `generate_prompt` function definition. I'll address the partial code to highlight any immediate issues.

**Issue 1:** The function is missing its closing parenthesis after `messages[-1]`. This can cause a syntax error.

```python
# Corrected line
messages[-1].append({'content': q})
```

**Optimization Suggestion:** If you want to use an immutable list when appending, you might consider converting it back to a tuple:

```python
# Recommended change
messages.append(Message(content=q))
```

However, if you prefer leaving the type as `list`, the current implementation works correctly.

**Overall Feedback:** Ensure that all necessary imports (e.g., for querying models) are included at the beginning of your file. Also verify that `application.name` and `application.desc` are defined and accessible within context since they are used in replacing placeholders with actual values.

PR #4138, opened 2025-09-29
feat: Raise error when can not generate prompt

shaohuzhang1:
feat: Raise error when can not generate prompt

### DIFF
--- a/ui/src/views/application/component/GeneratePromptDialog.vue
+++ b/ui/src/views/application/component/GeneratePromptDialog.vue
@@ -309,6 +309,12 @@ const getWrite = (reader: any) => {
         if (split) {
           for (const index in split) {
             const chunk = JSON?.parse(split[index].replace('data:', ''))
+            if (chunk.error) {
+              loading.value = false
+              stopStreaming()
+              middleAnswer.content = chunk.error
+              return Promise.reject(new Error(chunk.error))
+            }
             if (!chunk.is_end) {
               // 实时将新接收的内容添加到完整内容中
               fullContent.value += chunk.content


ui/src/views/application/component/GeneratePromptDialog.vue
```diff
@@ -309,6 +309,12 @@ const getWrite = (reader: any) => {
...
             const chunk = JSON?.parse(split[index].replace('data:', ''))
+            if (chunk.error) {
+              loading.value = false
+              stopStreaming()
+              middleAnswer.content = chunk.error
+              return Promise.reject(new Error(chunk.error))
+            }
             if (!chunk.is_end) {
               // 实时将新接收的内容添加到完整内容中
               fullContent.value += chunk.content
```
shaohuzhang1, line 320: This code appears to be intended for processing a stream of JSON objects sent from an asynchronous operation, likely using WebSocket communication with an AI backend. The primary task is to parse each chunk received, which contains potential error messages or content, and update UI elements accordingly. Here’s a review:

### Irregularities/Issues:
1. **Lack of Error Handling**: There's no general error handling outside the chunk parsing logic; errors within individual chunks may not propagate correctly.
2. **Overwriting `middleAnswer.content` Early**: If there's any error while parsing (e.g., non-JSON structure), it should overwrite `middleAnswer.content`, but this can lead to unexpected behavior if multiple errors occur consecutively.
3. **Stopping Streaming Without Checking Errors**: There's a call to `stopStreaming()` without checking if `loading.value` was set to `false`. This could happen unexpectedly during early stages of streaming.

### Optimization Suggestions:
1. **Single Stream Processing Function**: Combine all logic related to receiving and handling data into a single function that processes chunks one at a time. This might make the code cleaner and more readable.
   ```javascript
   const handleChunk = async (chunk): Promise<void> => {
     try {
       if (chunk.error) {
         loading.value = false;
         stopStreaming();
         middleAnswer.content = chunk.error;
         throw new Error(chunk.error);
       }

       if (!chunk.is_end) {
         fullContent.value += chunk.content;
       }
     } catch (error) {
       // Handle any overall errors here
       console.error('Error processing chunk:', error);
       loading.value = true;
     }
   };

   const getWrite = (reader: any) => {
     if (split) {
       for (const index in split) {
         handleChunk(JSON.parse(split[index].replace('data:', '')));
       }
     }
   };
   ```

2. **Use Promises Properly**: Instead of returning promises directly, consider chaining them properly to maintain clarity and flow control.
   ```javascript
   const processStream = async () => {
       reader.onmessage = async ({ data }) => {
         const chunk = JSON.parse(data.replace('data:', ''));
         try {
           await handleChunk(chunk);
           resolve();
         } catch (error) {
           reject(error);
         }
       };
     });
   };
   ```

3. **Ensure UI Updates Are Synchronized**: Use Vue.js methods like `$nextTick` to ensure DOM updates are complete before making further changes.

By addressing these points, you can enhance the robustness and reliability of your streaming processing logic.

PR #4148, opened 2025-10-09
fix: No history while use openai chat

shaohuzhang1:
fix: No history while use openai chat  --bug=1061932 --user=张展玮 【系统API】调用 openai 接口对话，无法获取历史聊天记录信息 https://www.tapd.cn/62980211/s/1781929

### DIFF
--- a/apps/chat/serializers/chat.py
+++ b/apps/chat/serializers/chat.py
@@ -222,9 +222,21 @@ class OpenAIChatSerializer(serializers.Serializer):
     def generate_chat(chat_id, application_id, message, chat_user_id, chat_user_type):
         if chat_id is None:
             chat_id = str(uuid.uuid1())
-        chat_info = ChatInfo(chat_id, chat_user_id, chat_user_type, [], [],
-                             application_id)
-        chat_info.set_cache()
+            chat_info = ChatInfo(chat_id, chat_user_id, chat_user_type, [], [],
+                                 application_id)
+            chat_info.set_cache()
+        else:
+            chat_info = ChatInfo.get_cache(chat_id)
+            if chat_info is None:
+                ser = ChatSerializers(data={
+                    'chat_id': chat_id,
+                    'chat_user_id': chat_user_id,
+                    'chat_user_type': chat_user_type,
+                    'application_id': application_id
+                })
+                ser.is_valid(raise_exception=True)
+                chat_info = ser.re_open_chat(chat_id)
+                chat_info.set_cache()
         return chat_id
 
     def chat(self, instance: Dict, with_valid=True):


apps/chat/serializers/chat.py
```diff
@@ -222,9 +222,21 @@ def get_message(instance):
...
+                    'chat_user_id': chat_user_id,
+                    'chat_user_type': chat_user_type,
+                    'application_id': application_id
+                })
+                ser.is_valid(raise_exception=True)
+                chat_info = ser.re_open_chat(chat_id)
+                chat_info.set_cache()
         return chat_id
 
     def chat(self, instance: Dict, with_valid=True):
```
shaohuzhang1, line 242: There are several concerns and minor optimizations in your code:

1. **Duplicate UUID Generation**: The code generates a new UUID each time `generate_chat` is called, even if a chat ID was provided. This can lead to inefficiency when creating many chats.

2. **Lazy Loading of Cache**: If multiple instances request the same chat info without caching it explicitly, it could still be retrieved from the cache on subsequent requests. Consider using memoization or explicit caching mechanisms.

3. **Serialization Logic**: The serialization logic within `get_cache` seems redundant with how you handle cases where no chat_info exists in the database. You might want to simplify this part.

4. **ChatInfo Initialization**: The way `set_cache` and `re_open_chat` are used indicates that they have internal state management that's not well-integrated into `get_cache`. It may make sense to ensure these methods work correctly together without redundancy or overhead.

```python
@@ -222,17 +238,35 @@ def generate_chat(chat_id=None, application_id=None, message=None, chat_user_id=N
     @staticmethod
     def get_cache(chat_id):
         # Check if chat_id exists in Redis/Database/Caching System
-        # Implement cache lookup logic here
+        try:
+            return ChatCache.objects.get(id=chat_id).load()  # Assuming ChatCache is a ORM model
+        except ChatCache.DoesNotExist:
+            return None

     def set_cache(self):
-        # Implement cache store logic here
+        self.save()

     @staticmethod
     def re_open_chat(chat_id):
-        # Restore previously saved chat data based on chat_id
+        return ChatInfo.load_from_database(chat_id)  # Assumes there's an API like load_from_database

class ChatSerializers(serializers.Serializer):
    chat_id = serializers.UUIDField(required=True)
    chat_user_id = serializers.CharField(required=True)
    chat_user_type = serializers.ChoiceField(choices=['user', 'bot'], required=True)
    messages = serializers.ListSerializer(child=models.TextField(), allow_empty=False)
    history = serializers.JSONField(default={})
    application_id = serializers.IntegerField(required=True)

    def create(self, validated_data):
        # Create a new ChatInfo instance and save it to DB
+
+        chat_id = uuid.uuid4().hex
+        chat_info_instance = ChatInfo(
+            id=chat_id,
+            chat_user_id=validated_data['chat_user_id'],
+            chat_user_type=validated_data['chat_user_type'],
+            application_id=validated_data['application_id']
+        )
+        chat_info_instance.save()
+
+        for _message in validated_data.get('messages', []):
+            Message.objects.create(text=_message, parent_message=None, sender=self.context['request'].user.id, chat_info=chat_info_instance)
 
     def update(self, instance, validated_data):
-        # Update existing chat information
-        # Implementation details depend on specific schema
+        updates = validated_data.copy()
+        if 'messages' not in updates:
+            del updates['messages']  # Prevent updating messages directly
```

This revised version should fix the issues mentioned above. It now properly handles lazy loading/cache retrieval, optimizes serializing/chat creation process, ensures proper separation between data fetching strategies, and provides more comprehensive exception handling around serialization errors.

PR #4154, opened 2025-10-09
chore: update tool_params handling in index.vue

shaohuzhang1:
chore: update tool_params handling in index.vue  --bug=1062339 --user=刘瑞斌 【MCP】MCP节点，从“引用变量” 切换到 “自定义” 没有清空 “引用变量” 的值 https://www.tapd.cn/62980211/s/1782327

### DIFF
--- a/ui/src/workflow/nodes/mcp-node/index.vue
+++ b/ui/src/workflow/nodes/mcp-node/index.vue
@@ -138,7 +138,7 @@
                   v-model="item.source"
                   size="small"
                   style="width: 85px"
-                  @change="form_data.tool_params[form_data.params_nested] = {}"
+                  @change="form_data.tool_params[form_data.params_nested] = {}; form_data.tool_params[form_data.params_nested][item.label.label]"
                 >
                   <el-option
                     :label="$t('views.applicationWorkflow.nodes.replyNode.replyContent.reference')"
@@ -207,6 +207,7 @@
                   v-model="item.source"
                   size="small"
                   style="width: 85px"
+                  @change="form_data.tool_params[item.label.label] = ''"
                 >
                   <el-option
                     :label="$t('views.applicationWorkflow.nodes.replyNode.replyContent.reference')"


ui/src/workflow/nodes/mcp-node/index.vue
```diff
@@ -207,6 +207,7 @@
                   v-model="item.source"
                   size="small"
                   style="width: 85px"
+                  @change="form_data.tool_params[item.label.label] = ''"
                 >
                   <el-option
                     :label="$t('views.applicationWorkflow.nodes.replyNode.replyContent.reference')"
```
shaohuzhang1, line 213: After reviewing the provided code snippet, I've identified a couple of potential issues:

### Issue 1: Overwriting Form Data Immediately After Changes

In both cases where an `@change` event is triggered on the select component, you're immediately overwriting all parameters under `form_data.tool_params[form_data.params_nested]`. If this nested object contains multiple properties or values, it might lead to loss of previously set data.

**Suggestion**: Change the line 

```javascript
form_data.tool_params[form_data.params_nested] = {};
```

to

```javascript
const currentParams = form_data.tool_params[form_data.params_nested];
// Ensure params exists before merging with new item label value
if (currentParams) {
    currentParams[item.label.label] = '';
} else {
    form_data.tool_params[form_data.params_nested] = { [item.label.label]: '' };
}
```

This prevents overwriting other keys in `params_nested`, ensuring that only necessary changes are made.

### Optimizations

#### Avoiding Redundant Operations in Select Component

The first version of the change handler already clears the entire nested object (`tool_params`) when `source` changes. However, if another parameter within the same level needs to be preserved after changing one, you could add conditional logic within this approach to avoid resetting non-relevant keys.

**Note:** If there's a specific requirement that all child nodes of `params_nested` need to reset every time its parent source changes, then the above modification aligns with your intentions. Otherwise, consider whether additional conditional checks can help manage updates more precisely at different levels.

PR #4164, opened 2025-10-10
chore: enhance re_findall function to handle None and empty string patterns

shaohuzhang1:


apps/common/utils/split_model.py
```diff
@@ -169,7 +169,12 @@ def parse_level(text, pattern: str):
...
-    if not pattern or not pattern.strip():
+    # 检查 pattern 是否为空或无效
+    if pattern is None:
+        return []
+
+    # 如果是字符串类型，检查是否为空字符串
+    if isinstance(pattern, str) and (not pattern or not pattern.strip()):
         return []
 
     try:
```
shaohuzhang1, line 180: The code has a few minor issues that require adjustments:

1. The function name `re_findall` does not match the expected behavior compared to Python's built-in `re.findall`. This could cause confusion for users.

2. There is no explicit documentation or comments explaining why certain parts of the function are included or what they do.

3. The use of `None` as a condition makes sense for empty patterns but should be commented to clarify its purpose.

4. The check for an empty string and whitespace-only strings at the beginning of the pattern might be unnecessary and should be removed since it would behave correctly with standard regular expressions.

Here is the revised version with these improvements:

```python
def re_finds_all(pattern, text):
    """
    Find all non-overlapping matches of the pattern in the string using re module.
    
    :param pattern: Regex pattern as a string or compiled regex object
    :param text: String to search within
    :return: List of all matched substrings
    """
    
    if pattern is None:
        return []
    
    try:
        import re
        return re.findall(pattern, text)
    except TypeError as e:
        print(f"Error while finding pattern: {e}")
        return []

# Example usage:
# result = re_finds_all(r'\b\w+\b', "Hello World")
# print(result)  # Output: ['Hello', 'World']
```

### Key Changes:
- **Function Name**: Renamed to `re_finds_all` to clearly indicate its intended functionality.
- **Comments**: Added docstring to describe the function and its parameters.
- **Error Handling**: Included a basic error handling mechanism for catching type errors when attempting to find the pattern in the text.
- **Removed Unnecessary Check**: Removed the additional check for empty or whitespace-only strings.

PR #4165, opened 2025-10-10
chore: enhance re_findall function to handle None and empty string patterns

shaohuzhang1:


apps/common/utils/split_model.py
```diff
@@ -169,7 +169,12 @@ def parse_level(text, pattern: str):
...
-    if not pattern or not pattern.strip():
+    # 检查 pattern 是否为空或无效
+    if pattern is None:
+        return []
+
+    # 如果是字符串类型，检查是否为空字符串
+    if isinstance(pattern, str) and (not pattern or not pattern.strip()):
         return []
 
     try:
```
shaohuzhang1, line 180: The function `re_findall` has several improvements to enhance readability and robustness:

1. **Early Return on Invalid Pattern**:
   - Added a check for `pattern is None` to handle cases where the pattern might be explicitly set to `None`.
2. **Check for Empty String Patterns**:
   - Changed the condition inside the main pattern check from using `not pattern.strip()` to just `(not pattern)`. This makes it more concise.

Here's the revised version of the code with these changes:

```python
import re

def parse_level(text, pattern: str):
    pass  # Placeholder implementation


def re_findall(pattern, text):
    # Early return if no pattern or invalid pattern is provided
    if pattern is None:
        return []

    # Check if pattern is an empty string
    if pattern == "":
        return []

    try:
        return re.findall(pattern, text)
    except Exception as e:
        print(f"Error parsing regex {pattern}: {e}")
        return []
```

### Key Changes:
- **Pattern Null Check**: Ensures that the function returns an empty list if the pattern is not provided or is explicitly set to `None`.
- **Empty String Check**: Adds a simple comparison to check if the pattern is an empty string.

These changes make the function more defensive and easier to understand, especially when dealing with unexpected input values.

PR #4168, opened 2025-10-10
fix: In workflow applications, the display of custom icons for sub applications and tools in the box below the variable selection is incorrect

shaohuzhang1:
fix: In workflow applications, the display of custom icons for sub applications and tools in the box below the variable selection is incorrect

### DIFF
--- a/ui/src/workflow/common/NodeCascader.vue
+++ b/ui/src/workflow/common/NodeCascader.vue
@@ -10,9 +10,12 @@
   >
     <template #default="{ node, data }">
       <span class="flex align-center" @wheel="wheel">
-        <component :is="iconComponent(`${data.type}-icon`)" class="mr-8" :size="18" />{{
-          data.label
-        }}</span
+        <component
+          :is="iconComponent(`${data.type}-icon`)"
+          class="mr-8"
+          :size="18"
+          :item="data"
+        />{{ data.label }}</span
       >
     </template>
   </el-cascader>
--- a/ui/src/workflow/common/app-node.ts
+++ b/ui/src/workflow/common/app-node.ts
@@ -75,6 +75,7 @@ class AppNode extends HtmlResize.view {
     }
     result.push({
       value: this.props.model.id,
+      icon: this.props.model.properties.node_data?.icon,
       label: this.props.model.properties.stepName,
       type: this.props.model.type,
       children: this.props.model.properties?.config?.fields || [],


ui/src/workflow/common/NodeCascader.vue
```diff
@@ -10,9 +10,12 @@
...
-        }}</span
+        <component
+          :is="iconComponent(`${data.type}-icon`)"
+          class="mr-8"
+          :size="18"
+          :item="data"
+        />{{ data.label }}</span
       >
     </template>
   </el-cascader>
```
shaohuzhang1, line 21: This code looks mostly correct with a couple of minor adjustments to enhance readability and maintainability:

1. **Attribute Order**: It seems like you're trying to ensure attributes (`class` and `@wheel`) are consistently placed within specific tags (e.g., `<a>`, `<button>`, etc.). If this is a requirement, make sure each element has its properties in the same order.

2. **Whitespace Consistency**: The addition of spaces after commas (`,"size": "18"`), colons (`:`), and semi-colons (;) can improve the readability, but it's already present.

4. **Code Formatting**: Ensure consistent indentation. In JavaScript objects, `{...}` typically use two-space indentation while strings might need to be single-quoted (`'string'`). However, if you prefer triple quotes (`"""```"), they must span at least three lines.

Here is a slightly formatted version for better readability:

```html
<el-cascader>
  <template #default="{ node, data }">
    <span class="flex align-center" @wheel="wheel">
      <component :is="iconComponent(`${data.type}-icon`)" 
                 class="mr-8" 
                 :size="18"><>{{ data.label }}</></span
    />
  </template>
</el-cascader>
```

In summary, there are no immediate technical issues, but ensuring proper formatting and attribute ordering can certainly help keep the code clean and easier to understand.

ui/src/workflow/common/app-node.ts
```diff
@@ -75,6 +75,7 @@ class AppNode extends HtmlResize.view {
     }
     result.push({
       value: this.props.model.id,
+      icon: this.props.model.properties.node_data?.icon,
       label: this.props.model.properties.stepName,
       type: this.props.model.type,
       children: this.props.model.properties?.config?.fields || [],
```
shaohuzhang1, line 81: The provided code snippet seems to contain a small issue that could be resolved without altering its main functionality. Let's address it:

```javascript
@@ -75,6 +75,9 @@
       },
       result.push({
         value: this.props.model.id,
+        icon: this.props.model.properties.node_data.icon || '',
         label: this.props.model.properties.stepName,
         type: this.props.model.type,
         children: this.props.model.properties.config.fields || [],
     });
```

Potential issue resolved: 
1. **Optional Chaining**: In JavaScript/TypeScript, using `?.` can prevent errors if an object is nullish (`null` or `undefined`). The line where you retrieve the icon should use `|| ''` to ensure a default string is returned in case of a missing property.

Optimization suggestion (not required but recommended):
2. **Inline Conditional Logic**: Consider moving the conditional logic inside the array literal, especially for better readability and potentially reducing nested checks.

This change will handle optional properties gracefully while maintaining the original intent of providing a node with an optional icon.

PR #4169, opened 2025-10-10
fix: Translate fixed name

shaohuzhang1:
fix: Translate fixed name

### DIFF
--- a/ui/src/locales/lang/en-US/layout.ts
+++ b/ui/src/locales/lang/en-US/layout.ts
@@ -23,7 +23,10 @@ export default {
     authorize: 'Authorized',
     inner_admin: 'System Admin',
     inner_wsm: 'Workspace Manager',
-    inner_user: 'Common User',
+    inner_user: 'Regular User',
+    root: 'Root Directory',
+    default_workspace: 'Default Workspace',
+    default_user_group: 'Default User Group',
   },
   time: {
     daysLater: 'days later',
--- a/ui/src/utils/common.ts
+++ b/ui/src/utils/common.ts
@@ -1,4 +1,5 @@
 import {nanoid} from 'nanoid'
+import {t} from '@/locales'
 
 /**
  * 数字处理
@@ -84,6 +85,22 @@ export function downloadByURL(url: string, name: string) {
   a.click()
   document.body.removeChild(a)
 }
+
+// 替换固定数据国际化
+const i18n_default_name_map:any = {
+  "系统管理员": 'layout.about.inner_admin',
+  "工作空间管理员": 'layout.about.inner_wsm',
+  "普通用户": 'layout.about.inner_user',
+  "根目录": 'layout.about.root',
+  "默认工作空间": 'layout.about.default_workspace',
+  "默认用户组": 'layout.about.default_user_group',
+}
+
+export function i18n_name(name: string) {
+  const key = i18n_default_name_map[name]
+  return key ? t(key) : name
+}
+
 
 // 截取文件名
 export function cutFilename(filename: string, num: number) {
--- a/ui/src/views/application/index.vue
+++ b/ui/src/views/application/index.vue
@@ -177,7 +177,7 @@
                   </template>
                   <template #subTitle>
                     <el-text class="color-secondary lighter" size="small">
-                      {{ $t('common.creator') }}: {{ item.nick_name }}
+                      {{ $t('common.creator') }}: {{ i18n_name(item.nick_name) }}
                     </el-text>
                   </template>
                 </CardBox>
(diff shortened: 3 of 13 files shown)


ui/src/utils/common.ts
```diff
@@ -85,6 +86,22 @@ export function downloadByURL(url: string, name: string) {
...
+
+export function i18n_name(name: string) {
+  const key = i18n_default_name_map[name]
+  return key ? t(key) : name
+}
+
+
 // 截取文件名
 export function cutFilename(filename: string, num: number) {
   const lastIndex = filename.lastIndexOf('.')
```
shaohuzhang1, line 107: The code looks mostly correct, but here are some minor suggestions and points to consider:

1. **Whitespace Issues**: The code has unnecessary whitespace around imports, after type annotations, comments, and before closing curly braces (`}`).

2. **Function Naming Convention**: Functions should be named using camelCase, such as `downloadByUrl`, rather than PascalCase.

3. **Error Handling for `document.body.removeChild`**: Adding error handling can prevent crashes if the element is not found unexpectedly (`a`). You could handle this with a try-catch block or by ensuring that the anchor element exists when calling `.removeChild()`.

4. **Unused Variable**: There's a variable `name` declared in the loop of `cutFilename` that isn't used anywhere else in the method. Consider removing it.

5. **Optimization**: If there's a possibility that `filenames` might contain many items, ensure that the loop doesn't unnecessarily compute the same index multiple times or modify elements directly within the loop without considering whether they've already been processed.

Here’s the cleaned-up code version incorporating these suggestions:

```javascript
import nanoid from 'nanoid';

/**
 * Number processing functions
 */
export function downloadByURL(url: string, name?: string) {
  if (!url || !name) return;

  const link = document.createElement('a');
  link.href = url;
  link.download = name!;
  link.click();
  document.body.removeChild(link);

  // Cleanup
  setTimeout(() => {
    window.URL.revokeObjectURL(link.download);
  }, 0);
}

// Replace fixed data localization
const i18n_default_name_map = {
  "System Administrator": 'layout.about.inner_admin',
  "Workspace Administrator": 'layout.about.inner_wsm',
  "Normal User": 'layout.about.inner_user',
  "Root Directory": 'layout.about.root',
  "Default Workspace": 'layout.about.default_workspace',
  "Default User Group": 'layout.about.default_user_group',
};

export function i18nName(name: string): string | undefined {
  const key = i18n_default_name_map[name];
  return key ? t(key) : name;
}

function cutFilename(filename: string, num: number): string {
  let lastIndex = filename.lastIndexOf('.');
  while (lastIndex > -1 && --num >= 0) {
    lastIndex = filename.lastIndexOf('.', lastIndex - 1);
  }
  return filename.substr(lastIndex).replace(/[^a-zA-Z0-9_.]/g, '_');
}
```

These changes enhance readability and potentially make the code more robust and efficient.

ui/src/locales/lang/en-US/layout.ts
```diff
@@ -23,7 +23,10 @@ export default {
...
     inner_admin: 'System Admin',
     inner_wsm: 'Workspace Manager',
-    inner_user: 'Common User',
+    inner_user: 'Regular User',
+    root: 'Root Directory',
+    default_workspace: 'Default Workspace',
+    default_user_group: 'Default User Group',
   },
   time: {
     daysLater: 'days later',
```
shaohuzhang1, line 32: Sure, please supply the piece of code you'd like me to review.

ui/src/views/application/index.vue
```diff
@@ -330,6 +330,7 @@ import ApplicationApi from '@/api/application/application'
 import { MsgSuccess, MsgConfirm, MsgError } from '@/utils/message'
 import useStore from '@/stores'
 import { t } from '@/locales'
+import { i18n_name } from '@/utils/common'
 import { useRouter, useRoute } from 'vue-router'
 import { isWorkFlow } from '@/utils/application'
 import { resetUrl } from '@/utils/common'
```
shaohuzhang1, line 336: The code provided appears to be part of a Vue.js application that uses internationalization (i18n). The changes made include using an `i18n_name` function instead of directly calling `$t('common.creator').toLowerCase()`.

### Analysis:
1. **Template Usage**:
   - In both templates, the same string (`$t('common.creator')`) is being used with the nick name. This means that if `"common.creator"` changes its translation in the locale files, these strings will need to be updated manually.

2. **Use of `auto-tooltip` Component**:
   - If the intention was to provide tooltips for usernames, then replacing `{{ item.nick_name }}` with `{{ i18n_name(item.nick_name) }}" would make more sense since tooltips should typically display the translated value.
   
3. **Functionality Check**:
   - Ensure that `i18n_name` function correctly translates the input text into the desired language, especially considering potential variations across different locales like Chinese and English.

   - Consider caching results or lazy-loading translations where appropriate to improve performance, particularly if `item.nick_name` or other frequently accessed data is expensive to translate.

Overall, the main issue here is the consistency of how localization functions are called, which can lead to mismatches between expected translations and rendered output.

PR #4177, opened 2025-10-11
fix: The tool in the application export loop body has not been exported

shaohuzhang1:
fix: The tool in the application export loop body has not been exported

### DIFF
--- a/apps/application/serializers/application.py
+++ b/apps/application/serializers/application.py
@@ -683,7 +683,10 @@ class ApplicationOperateSerializer(serializers.Serializer):
             application = QuerySet(Application).filter(id=application_id).first()
             tool_id_list = [node.get('properties', {}).get('node_data', {}).get('tool_lib_id') for node
                             in
-                            application.work_flow.get('nodes', []) if
+                            application.work_flow.get('nodes', []) + reduce(lambda x, y: [*x, *y], [
+                                n.get('properties', {}).get('node_data', {}).get('loop_body', {}).get('nodes', []) for n
+                                in
+                                application.work_flow.get('nodes', []) if n.get('type') == 'loop-node'], []) if
                             node.get('type') == 'tool-lib-node']
             tool_list = []
             if len(tool_id_list) > 0:
@@ -760,7 +763,7 @@ class ApplicationOperateSerializer(serializers.Serializer):
         work_flow_version.save()
         access_token = hashlib.md5(
             str(uuid.uuid7()).encode()).hexdigest()[
-            8:24]
+                       8:24]
         application_access_token = QuerySet(ApplicationAccessToken).filter(
             application_id=application.id).first()
         if application_access_token is None:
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application.py
```diff
@@ -965,7 +968,8 @@ def speech_to_text(self, instance, debug=True, with_valid=True):
             application = QuerySet(ApplicationVersion).filter(application_id=application_id).order_by(
                 '-create_time').first()
         if application.stt_model_enable:
-            model = get_model_instance_by_model_workspace_id(application.stt_model_id, application.workspace_id, **application.stt_model_params_setting)
+            model = get_model_instance_by_model_workspace_id(application.stt_model_id, application.workspace_id,
+                                                             **application.stt_model_params_setting)
             text = model.speech_to_text(instance.get('file'))
             return text
```
shaohuzhang1, line 975: The code you provided appears to be part of an application or service that handles various functionalities such as data export, publication, and speech-to-text conversion. Here's a checklist of common areas where improvements can be made:

### Code Improvements

1. **Exception Handling**: The current implementation doesn't include exception handling. Adding try-except blocks will help catch and deal with any errors during database queries or external calls.

2. **String Encoding**: When hashing a string using `hashlib`, it is important to ensure consistent behavior across platforms. Consider encoding the UUID universally as UTF-8.

3. **Loop Optimization**: The use of comprehension lists and `reduce` might not be necessary for this specific case since you only need to extend one list based on certain conditions. A simpler approach should suffice.

4. **Parameter Initialization**: Ensure that all dictionaries used within loops (like `application_stt_model_params_setting`) are initialized before use to avoid potential key errors.

5. **Security Concerns**: Make sure sensitive operations like generating access tokens involve secure practices, especially when dealing with authentication and authorization.

6. **Code DRY Principle**: There are repeated sections like checking if a token exists and saving it. Consider abstracting these into separate methods or functions to maintain clean and modular code.

### Optimizations

- **Database Query Efficiency**: Minimize database round trips by fetching related objects in bulk rather than individual records.
  
- **Data Validation**: Add input validation checks at the beginning of your methods to ensure that inputs meet expected criteria before processing.

Here’s an example of how some of these improvements could look:

```python
import hashlib
import uuid
from functools import reduce

    try:
        application = QuerySet(Application).filter(id=application_id).first()
        tool_id_list = [node['properties']['node_data'].get('tool_lib_id') 
                         for node in reduce(lambda x, y: [*x, *y], 
                                            ([n for n in nodes if n['type'] == 'tool-lib-node']] 
                                             + [n.get('properties', {}).get('node_data', {}).get('loop_body', {}).get('nodes', [])] 
                                             for nodes in application.work_flow.get('nodes', [])), [])
        
        tool_list = []
        if len(tool_id_list) > 0:
            # Process tools ...
    
    except Exception as e:
        # Log error ...
        print(f"Error exporting data: {e}")

# Similar improvement for other functions...
```

These changes improve robustness, security, and efficiency while maintaining clarity and readability.

PR #4187, opened 2025-10-14
fix: Typos error

shaohuzhang1:
fix: Typos error

### DIFF
--- a/ui/src/views/system/resource-authorization/constant.ts
+++ b/ui/src/views/system/resource-authorization/constant.ts
@@ -23,13 +23,13 @@ const permissionOptions = [
 ]
 
  
-const getPermissionOptions = (isFodler = false, isRootFolder = false) => {
-  if (isFodler && isRootFolder) {
+const getPermissionOptions = (isFolder = false, isRootFolder = false) => {
+  if (isFolder && isRootFolder) {
     return permissionOptions.filter(
       item => item.value === AuthorizationEnum.VIEW || item.value === AuthorizationEnum.MANAGE
     )
   }
-  if (isFodler) {
+  if (isFolder) {
     return permissionOptions
   }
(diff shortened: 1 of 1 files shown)


ui/src/views/system/resource-authorization/constant.ts
```diff
@@ -23,13 +23,13 @@ const permissionOptions = [
...
+  if (isFolder && isRootFolder) {
     return permissionOptions.filter(
       item => item.value === AuthorizationEnum.VIEW || item.value === AuthorizationEnum.MANAGE
     )
   }
-  if (isFodler) {
+  if (isFolder) {
     return permissionOptions
   }
   if (hasPermission([EditionConst.IS_EE, EditionConst.IS_PE], 'OR')) {
```
shaohuzhang1, line 35: The provided code looks generally correct for handling different permission options based on whether an entity is a folder, root folder, or does not have specific permissions. However, there are some minor issues that could be addressed:

1. **Misnaming**: Variable naming consistency is important for readability and maintainability. Using `isFolder` instead of both `isFodler` would make the function signature more uniform.

2. **Comments**: While comments can help clarify the logic, they should be concise and relevant to the code. In this case, the comment could be refined.

Here's the revised version with these suggestions:

```javascript
const permissionOptions = [
]

const getPermissionOptions = (isFolder = false, isRootFolder = false) => {
  if (isFolder && isRootFolder) {
    return permissionOptions.filter(item => item.value === AuthorizationEnum.VIEW || item.value === AuthorizationEnum.MANAGE)
  }
  if (isFolder) {
    return permissionOptions
  }
  if (hasPermission([EditionConst.IS_EE, EditionConst.IS_PE], 'OR')) { 
    // Additional checks here
  }
}
```

These changes ensure better coding practices, making it easier to understand the purpose and usage of the function.

PR #4193, opened 2025-10-15
fix: Get tool error

shaohuzhang1:
fix: Get tool error

### DIFF
--- a/apps/tools/serializers/tool.py
+++ b/apps/tools/serializers/tool.py
@@ -246,7 +246,7 @@ class ToolSerializer(serializers.Serializer):
             if workspace_id is not None:
                 folder_query_set = folder_query_set.filter(workspace_id=workspace_id)
                 default_query_set = default_query_set.filter(workspace_id=workspace_id)
-            if folder_id is not None:
+            if folder_id is not None and folder_id != workspace_id:
                 folder_query_set = folder_query_set.filter(parent=folder_id)
                 default_query_set = default_query_set.filter(folder_id=folder_id)
             if name is not None:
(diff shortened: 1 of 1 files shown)


apps/tools/serializers/tool.py
```diff
@@ -267,7 +267,6 @@ def get_query_set(self, workspace_manage, is_x_pack_ee):
                 tool_query_set = tool_query_set.filter(tool_type=tool_type)
 
             query_set_dict = {
-                'folder_query_set': folder_query_set,
                 'tool_query_set': tool_query_set,
                 'default_query_set': default_query_set,
             }
```
shaohuzhang1, line 272: The provided code seems mostly correct and well-structured for filtering a queryset based on various criteria like `workspace_id`, `folder_id`, `name`, etc. However, there are a few minor improvements that can be made:

1. **Consistent Naming Convention**: The names of variables (`folder_query_set`, `tool_query_set`, `default_query_set`) seem to follow a consistent naming scheme, which is good.

2. **Remove Redundancy in Logic**: Although not strictly necessary due to the logic's correctness, it might be clearer to remove the redundant condition where `folder_id` equals `workspace_id`. This check does not affect the filtered results if you're checking both conditions together:

```python
if folder_id is not None and folder_id != workspace_id:
```

This line means that the `folder_id` will only be included in the filter if it's different from the `workspace_id`.

3. **Ensure Proper Quoting Consistency**: Make sure all string literals use appropriate quoting (single or double) consistently within the same block to avoid confusion and potential bugs due to mixing types.

Otherwise, the code looks clean and functional according to the current knowledge cutoff date. It provides flexibility in querying documents based on specific conditions related to their folders and tools.

PR #4203, opened 2025-10-16
feat: Create a composite unique key to remove duplicate data

shaohuzhang1:
feat: Create a composite unique key to remove duplicate data


apps/system_manage/migrations/0004_alter_systemsetting_type_and_more.py
```diff
@@ -0,0 +1,37 @@
...
+    ]
+
+    operations = [
+       migrations.RunPython(remove_duplicates,
+                            ),
+        migrations.AlterUniqueTogether(
+            name='workspaceuserresourcepermission',
+            unique_together={('workspace_id', 'user', 'auth_target_type', 'target')},
+        ),
+    ]
```
shaohuzhang1, line 37: The provided code checks for duplicate entries in the `WorkspaceUserResourcePermission` model based on specific fields: `workspace_id`, `user`, `auth_target_type`, and `target`. Then deletes all but one occurrence of each duplicate. Here are some points to consider:

1. **Duplicate Check**: The query uses the `RowNumber()` function from `django.db.models.functions.RowNumber()`. This is correct for filtering out duplicates.

2. **Field Partitioning**: The partitioning criteria are set by `F('workspace_id')`, `F('user')`, `F('auth_target_type')`, and `F('target')`, which ensures that duplicates within these groups are identified.

3. **Order By and Filter**: The ordering is done by `_time.desc()` to ensure the latest entry remains, while duplicates with earlier creation times are filtered out.

4. **Run Python Operation**: The `migrations.RunPython` operation correctly calls the `remove_duplicates` function during migration.

5. **Alter Unique Together Constraint**: After removing duplicates, the unique-together constraint is added back with the same fields.

6. **No Optimization Suggestions**: Based on the current implementation, there aren't many areas where optimization can be applied directly. However, consider creating an index on the combination of `workspace_user_resource_permission_model._meta.fields` to improve performance if the table grows significantly.

Overall, the code appears well-structured and should work as intended for deduplicating entries according to specified rules.

apps/system_manage/models/workspace_user_permission.py
```diff
@@ -57,3 +57,4 @@ class WorkspaceUserResourcePermission(models.Model):
 
     class Meta:
         db_table = "workspace_user_resource_permission"
+        unique_together = ('workspace_id', 'user', 'auth_target_type', 'target')
```
shaohuzhang1, line 60: The provided code snippet is mostly correct for a Django model to manage permissions related to workspace user resources. However, there are a few suggestions for improvement:

1. **Docstring**: It's a good practice to add a docstring explaining the purpose of each attribute.

2. **Meta.ordering** (Optional): If you want certain orderings, you can specify them here without needing a separate `order_by` queryset method in views or elsewhere.

Here's improved version with the suggested changes:

```python
from django.db import models

class WorkspaceUserResourcePermission(models.Model):
    # Attributes
    workspace = models.ForeignKey('Workspaces', on_delete=models.CASCADE)
    user = models.ForeignKey('WorkspaceUsers', on_delete=models.CASCADE)
    auth_target_type = models.CharField(max_length=100)  # Assuming this should be specific like 'permission_group'
    target = models.CharField(max_length=255)

    # Metadata
    class Meta:
        db_table = "workspace_user_resource_permission"
        unique_together = ('workspace_id', 'user', 'auth_target_type', 'target')
        
        # Optional ordering
        # ordering = ['-created_at']  # For example, sort by creation time
        
    def __str__(self):
        return f"{self.user} {self.auth_target_type}.{self.target}"
```

### Explanation:

1. **Unique Togetherness**: Ensures that no two records in the database have the same combination of `workspace_id`, `user`, `auth_target_type`, and `target`.

2. **Ordering (Optional)**: Uncommenting `ordering = ['-created_at']` adds an optional option to display these entries in descending order based on their creation date (`created_at`). This is useful for listing recent permissions or other chronological data.

3. **Str Representation**: The `__str__()` method provides a readable string representation of the object, which can improve debugability.

If you plan to include additional fields or methods in your `WorkspaceUserResourcePermission` model, make sure they align with the intended use case and adhere to best practices from Django's ORM design principles.

PR #4229, opened 2025-10-22
feat: Update django migration

shaohuzhang1:
feat: Update django migration

### DIFF
--- a/apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
+++ b/apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
@@ -22,7 +22,7 @@ def get_workspace_user_resource_permission_list(auth_target_type, workspace_user
     return reduce(lambda x, y: [*x, *y], [
         [WorkspaceUserResourcePermission(target=f.id, workspace_id=f.workspace_id, user_id=wurm.user_id,
                                          auth_target_type=auth_target_type, auth_type="RESOURCE_PERMISSION_GROUP",
-                                         permission_list=['VIEW']) for wurm in
+                                         permission_list=['VIEW','MANAGE'] if wurm.user_id == f.user_id else ['VIEW']) for wurm in
          workspace_user_role_mapping_model_workspace_dict.get(f.workspace_id, [])] for f in
         QuerySet(folder_model).all()], [])
 


apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
```diff
@@ -22,7 +22,7 @@ def get_workspace_user_resource_permission_list(auth_target_type, workspace_user
     return reduce(lambda x, y: [*x, *y], [
         [WorkspaceUserResourcePermission(target=f.id, workspace_id=f.workspace_id, user_id=wurm.user_id,
                                          auth_target_type=auth_target_type, auth_type="RESOURCE_PERMISSION_GROUP",
-                                         permission_list=['VIEW']) for wurm in
+                                         permission_list=['VIEW','MANAGE'] if wurm.user_id == f.user_id else ['VIEW']) for wurm in
          workspace_user_role_mapping_model_workspace_dict.get(f.workspace_id, [])] for f in
         QuerySet(folder_model).all()], [])
```
shaohuzhang1, line 28: The code provided contains a few issues:

1. In the first iteration of the list comprehension (the one without a conditional), all `workspace_user_role_mapping_model_workspace_dict.get(f.workspace_id, [])` will be non-empty regardless of whether there's any matching records. This might lead to unnecessary processing or errors if some keys are not present.

2. The condition inside the second `if` statement is only checked once per workspace (`f`). If `wurm.user_id` changes between iterations of this loop for the same `f`, it won't affect the final outcome.

3. The logic can be simplified and optimized using set operations to avoid repeating checks within each iteration. However, given that your current approach is already performing well, these changes may introduce new complexity if performance becomes an issue at runtime.

Here are the specific improvements suggested:
```python
from functools import accumulate

def get_workspace_user_resource_permission_list(auth_target_type, workspace_user_role_mapping_model_workspace_dict):
    # Simplify and optimize: use dictionaries where possible
    user_permissions = defaultdict(set)
    
    # Build user permissions dict with filtered results directly
    for folder_f in QuerySet(folder_model).all():
        workspace_users = (
            WorkspaceUserRoleMappingModel.objects.filter(workspace_user__user_id=UserModel._default_manager.get(pk=folder_f.folder.owner)).values('id', 'user_id')
        )
        
        for f in workspace_users:
            if f['user_id'] == folder_f.folder.owner.pk:
                continue  # Skip own account as no additional permissions needed
            
            permission_group_name = (
                FolderUserRoleModel.objects.get(pk=f['id']).group.name.lower()
            )
            
            permission_set = {
                "VIEW": True,
                "MANAGE": bool(
                    folder_user_role_mapping_model_workspace_dict[permission_group_name]
                )
            }
            
            user_permissions[f['user_id']].update(permission_set)
    
    return list(accumulate(user_permissions.items(), lambda acc, item: [*acc, {'target': item[0], **item[1]}]))


@receiver(post_save, sender=UserModel)
def on_user_saved(instance, created=False, **kwargs) -> None:
    """
    Handles saving users and ensuring relevant permissions tables have correct relationships.
    """
    ...

```
Note: I've assumed `QuerySet`, `Reducer`, `FolderModel`, `WorkspaceUserRoleMappingModel`, `FolderUserRoleModel`, `UserModel`, and other entities exist based on context; replace them accordingly before implementation. Also note usage of `filter`, `get`, and dictionary comprehensions for efficiency. If performance becomes a concern later, consider implementing caching mechanisms.

PR #4233, opened 2025-10-22
fix: Variable splitting node expression not validated

shaohuzhang1:
fix: Variable splitting node expression not validated

### DIFF
--- a/ui/src/workflow/nodes/variable-splitting/component/VariableFieldDialog.vue
+++ b/ui/src/workflow/nodes/variable-splitting/component/VariableFieldDialog.vue
@@ -50,8 +50,7 @@
       <el-form-item
         :label="$t('views.applicationWorkflow.nodes.variableSplittingNode.expression.label')"
         :required="true"
-        prop="label"
-        :rules="rules.label"
+        prop="expression"
       >
         <el-input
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/variable-splitting/component/VariableFieldDialog.vue
```diff
@@ -50,8 +50,7 @@
       <el-form-item
         :label="$t('views.applicationWorkflow.nodes.variableSplittingNode.expression.label')"
         :required="true"
-        prop="label"
-        :rules="rules.label"
+        prop="expression"
       >
         <el-input
           v-model="form.expression"
```
shaohuzhang1, line 56: In the given code snippet, there is one main issue to be addressed:

1. The label property should match the prop used. Here's how you can correct it:

   ```diff
   +      :label="$t('views.applicationWorkflow.nodes.variableSplittingNode.expression.label')"
   
   ```
   Modify it to:
   ```diff
   -      :required="true"
+      :prop="expression"
 
       >
   ```

This change ensures that the corresponding rule (`rules.label`) will correctly validate the input against the `label` property on the form state when submitted.

No other major issues were found in this specific code snippet.

PR #4235, opened 2025-10-22
feat: Add SECRET_KEY environment variable settings

shaohuzhang1:
feat: Add SECRET_KEY environment variable settings

### DIFF
--- a/apps/maxkb/settings/base.py
+++ b/apps/maxkb/settings/base.py
@@ -22,7 +22,7 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 # See https://docs.djangoproject.com/en/4.2/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = 'django-insecure-zm^1_^i5)3gp^&0io6zg72&z!a*d=9kf9o2%uft+27l)+t(#3e'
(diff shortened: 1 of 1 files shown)


apps/maxkb/settings/base.py
```diff
@@ -22,7 +22,7 @@
 # See https://docs.djangoproject.com/en/4.2/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = 'django-insecure-zm^1_^i5)3gp^&0io6zg72&z!a*d=9kf9o2%uft+27l)+t(#3e'
+SECRET_KEY = CONFIG.get("SECRET_KEY") or 'django-insecure-zm^1_^i5)3gp^&0io6zg72&z!a*d=9kf9o2%uft+27l)+t(#3e'
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = CONFIG.get_debug()
```
shaohuzhang1, line 28: Your code contains:

- An unnecessary call to `CONFIG.get_debug()` when setting the SECRET_KEY, which might not be needed unless you have specific logic that depends on it.

Recommendation: Simplify the SECRET_KEY assignment by removing the conditional statement and just use `CONFIG.get()`. If DEBUG is needed elsewhere, ensure it is initialized appropriately before using it here.

PR #4237, opened 2025-10-22
fix: When collecting data from the multiple-choice box of the form collection node, it is recommended to support pressing enter to select the option

shaohuzhang1:
fix: When collecting data from the multiple-choice box of the form collection node, it is recommended to support pressing enter to select the option

### DIFF
--- a/ui/src/components/dynamics-form/items/select/MultiSelect.vue
+++ b/ui/src/components/dynamics-form/items/select/MultiSelect.vue
@@ -5,6 +5,7 @@
     filterable
     allow-create
     clearable
+    default-first-option
     :reserve-keyword="false"
     v-bind="$attrs"
     v-model="_modelValue"


ui/src/components/dynamics-form/items/select/MultiSelect.vue
```diff
@@ -5,6 +5,7 @@
     filterable
     allow-create
     clearable
+    default-first-option
     :reserve-keyword="false"
     v-bind="$attrs"
     v-model="_modelValue"
```
shaohuzhang1, line 11: The provided code snippet looks mostly correct but can be optimized slightly:

1. **Duplicate Default Value**: The `default-first-option` property is included twice at the end of the line, which might not be intended unless you have specific logic to handle this.

2. **Remove Duplicate Attributes**: Since all attributes are being passed via `v-bind="$attrs"`, including those that will overwrite existing properties (like `$attrs.default-first-option`), redundant assignments like `default-first-option:"true"` could conflict with what's already set via bindings.

Here's a more concise version:

```vue
<template>
  <!-- Your template content -->
</template>

<script>
export default {
  props: {
    filterable: Boolean,
    allowCreate: Boolean,
    clearable: Boolean
  },
  computed: {
    _modelValue() {
      // Define your computed or data properties here
    }
  },
  methods: {
    someMethod() {
      // Define your methods here
    }
  }
}
</script>

<style scoped>
/* Your styles go here */
</style>
```

### Optimizations and Corrections:
- Remove the duplicate `default-first-option="true"`.
- Ensure there are no conflicts between direct assignment (`v-model`) and binding (`v-bind`).
- Keep your script organized within `<script>`, `<computed>`, and `<methods>` blocks if necessary.

PR #4238, opened 2025-10-22
fix: Error parameter not popped up prompt during variable splitting and parameter extraction node verification

shaohuzhang1:
fix: Error parameter not popped up prompt during variable splitting and parameter extraction node verification

### DIFF
--- a/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
+++ b/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
@@ -46,7 +46,7 @@ def get_default_model_params_setting(model_id):
 
 
 def generate_properties(variable_list):
-    return {variable['field']: {'type': variable['parameter_type'], 'description': variable['desc'],
+    return {variable['field']: {'type': variable['parameter_type'], 'description': (variable.get('desc') or ""),
                                 'title': variable['label']} for variable in
             variable_list}
 
--- a/ui/src/workflow/nodes/parameter-extraction-node/index.vue
+++ b/ui/src/workflow/nodes/parameter-extraction-node/index.vue
@@ -79,7 +79,9 @@
         <el-form-item
           prop="variable_list"
           :rules="{
-            message: $t('views.applicationWorkflow.nodes.parameterExtractionNode.extractParameters.label'),
+            message: $t(
+              'views.applicationWorkflow.nodes.parameterExtractionNode.extractParameters.variableListPlaceholder',
+            ),
             trigger: 'blur',
             required: true,
           }"
@@ -184,7 +186,9 @@ const model_change = (model_id?: string) => {
 
 const VariableSplittingRef = ref()
 const validate = async () => {
-  return VariableSplittingRef.value.validate()
+  return VariableSplittingRef.value.validate().catch((err: any) => {
+    return Promise.reject({ node: props.nodeModel, errMessage: err })
+  })
 }
 
 onMounted(() => {
--- a/ui/src/workflow/nodes/variable-splitting/index.vue
+++ b/ui/src/workflow/nodes/variable-splitting/index.vue
@@ -39,7 +39,9 @@
         <el-form-item
           prop="variable_list"
           :rules="{
-            message: $t('views.applicationWorkflow.variable.placeholder'),
+            message: $t(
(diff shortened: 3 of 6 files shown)


ui/src/workflow/nodes/parameter-extraction-node/index.vue
```diff
@@ -184,7 +186,9 @@ const model_change = (model_id?: string) => {
 
 const VariableSplittingRef = ref()
 const validate = async () => {
-  return VariableSplittingRef.value.validate()
+  return VariableSplittingRef.value.validate().catch((err: any) => {
+    return Promise.reject({ node: props.nodeModel, errMessage: err })
+  })
 }
 
 onMounted(() => {
```
shaohuzhang1, line 194: There are no irregularities or significant issues with the provided code snippet. The code appears to be correctly structured and contains typical practices for form validation using El-form from Element UI in Vue.js.

### Suggested Optimization:

#### 1. **Error Handling in `validate` Function:**
   - Although there's already error handling in this part of the code using `.catch()`, it could benefit from adding a more clear message when an error occurs. This can help debugging and improve user experience if something goes wrong.

```python
try:
    await VariableSplittingRef.value.validate()
except ValidationErrors as e:
    throw new Error(`Validation failed for variable spli<|fim_suffix|>on will then catch and reject this promise with additional data about the validation errors, including details such as the affected node and the specific error messages. If successful, the validated result is returned normally.

File: ui/src/workflow/nodes/variable-splitting/index.vue [hunk 2]
```diff
@@ -81,7 +83,9 @@ const form_data = computed({
 })
 const VariableSplittingRef = ref()
 const validate = async () => {
-  return VariableSplittingRef.value.validate()
+  return VariableSplittingRef.value.validate().catch((err: any) => {
+    return Promise.reject({ node: props.nodeModel, errMessage: err })
+  })
 }
 onMounted(() => {
   set(props.nodeModel, 'validate', validate)
```
shaohuzhang1, line 91: There are two main issues in this code:

1. The `message` string in the validation rule is not localized correctly when using internationalization library `$t`.

2. There's an error handling block inside `validate()` where it catches rejection, but it incorrectly returns `{ node: props.nodeModel, errMessage: err }`. This might prevent further actions from being processed correctly.

**Suggested Changes**

```diff
const rules = computed(() => ({
-  variable_list: [
+  variable_list: {
     required: true,
     message: $t(
       'views.applicationWorkflow.nodes.variableSplittingNode.variableListPlaceholder',
     ),
-    trigger: 'blur'
+    },
+    trigger: ['blur', 'change'],
  }),
}))
```

Explanation of changes:
- Added braces to make the `message` property inline with other properties of the same level, which makes object syntax cleaner.
- Enhanced the trigger to include both blur and change events (`trigger: ['blur', 'change']`) to ensure that the form item checks if all required fields have been filled without losing focus after changing values.

Regarding the second issue of returning the promise directly instead of passing it through reject (which would require additional code elsewhere to handle), I think the correct approach is to either propagate errors properly so they can be handled higher up in your application logic, or to log them appropriately within the catch block if there's no possibility of handling them further. If you're confident these errors won't need to be propagated, leaving them caught should be sufficient for debugging purposes. But usually, it would best practice to fix what caused the initial failure rather than ignore it silently.

apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
```diff
@@ -46,7 +46,7 @@ def get_default_model_params_setting(model_id):
 
 
 def generate_properties(variable_list):
-    return {variable['field']: {'type': variable['parameter_type'], 'description': variable['desc'],
+    return {variable['field']: {'type': variable['parameter_type'], 'description': (variable.get('desc') or ""),
                                 'title': variable['label']} for variable in
             variable_list}
```
shaohuzhang1, line 52: The provided code appears to be generating property settings based on a list of variables with fields like `field`, `parameter_type`, `desc`, and `label`. The main issue is that if there's no description (`'desc'`) provided for a variable, it currently raises an error because you're trying to access `None` properties directly.

I suggest:

1. Replace the line: `'type': variable['type'], 'description': (variable.get('desc') or ""),`
2. With this one: `'type': variable['type'], 'description': str((variable.get('desc', '')),`

This modification ensures that if `variable['desc']` is `None`, it defaults to an empty string, avoiding any runtime errors.

PR #4239, opened 2025-10-22
fix: The parameter extraction node does not display input parameters in the execution details

shaohuzhang1:
fix: The parameter extraction node does not display input parameters in the execution details

### DIFF
--- a/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
+++ b/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
@@ -88,8 +88,10 @@ class BaseParameterExtractionNode(IParameterExtractionNode):
         for key, value in details.get('result').items():
             self.context[key] = value
         self.context['result'] = details.get('result')
+        self.context['request'] = details.get('request')
 
     def execute(self, input_variable, variable_list, model_params_setting, model_id, **kwargs) -> NodeResult:
+        self.context['request'] = input_variable
         if model_params_setting is None:
             model_params_setting = get_default_model_params_setting(model_id)
         workspace_id = self.workflow_manage.get_body().get('workspace_id')
@@ -106,6 +108,7 @@ class BaseParameterExtractionNode(IParameterExtractionNode):
             "index": index,
             'run_time': self.context.get('run_time'),
             'type': self.node.type,
+            'request': self.context.get('request'),
             'result': self.context.get('result'),
             'status': self.status,
             'err_message': self.err_message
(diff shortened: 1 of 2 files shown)


apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
```diff
@@ -106,6 +108,7 @@ def get_details(self, index: int, **kwargs):
             "index": index,
             'run_time': self.context.get('run_time'),
             'type': self.node.type,
+            'request': self.context.get('request'),
             'result': self.context.get('result'),
             'status': self.status,
             'err_message': self.err_message
```
shaohuzhang1, line 114: Yes, there are two issues. You forgot to add 'self' before calling methods on `workflow_manage`. The second issue is that you're trying to access keys from self.context that don't exist when they should be set.

Here's the corrected code:

    @@ -88,8 +88,10 @@ def save_context(self, details, workflow_manage):
         for key, value in details.get('result').items():
             self.context[key] = value
         self.context['result'] = details.get('result')
+        self.context['request'] = details.get('request', '')
 
     def execute(self, input_variable, variable_list, model_params_setting, model_id, **kwargs) -> NodeResult:
+        # Make sure context request doesn't error out when it shouldn't.
+        self.request = kwargs.get('request') if isinstance(input_variable, dict) and 'request' in input_variable else ''
             
         if model_params_setting is None:
             model_params_setting = get_default_model_params_setting(model_id)
@@ -106,6 +108,7 @@ def get_details(self, index: int, **kwargs):
             "index": index,
             'type': self.node.type,
+            'request': self.context.get('request'), # Corrected this line
             'result': self.context.get('result'),
             'status': self.status,
```

File: apps/application/flow/step_node/variable_splitting_node/impl/base_variable_splitting_node.py [hunk 2]
```diff
@@ -47,6 +49,7 @@ def get_details(self, index: int, **kwargs):
             "index": index,
             'run_time': self.context.get('run_time'),
             'type': self.node.type,
+            'request': self.context.get('request'),
             'result': self.context.get('result'),
             'status': self.status,
             'err_message': self.err_message
```
shaohuzhang1, line 55: 1. In the `save_context` method, adding `'request'` to the context dictionary ensures that all necessary data is stored.
2. It's a good practice to initialize all variables used within methods before they are accessed.
3. Consider using more descriptive variable names to improve code readability.
4. Adding logging could be beneficial during debugging.

PR #4250, opened 2025-10-24
fix: Xinference add

shaohuzhang1:
fix: Xinference add  --bug=1062934 --user=张展玮 【github#4242】添加xinference平台的语音识别模型失败 https://www.tapd.cn/62980211/s/1789683

### DIFF
--- a/apps/models_provider/impl/xinference_model_provider/credential/stt.py
+++ b/apps/models_provider/impl/xinference_model_provider/credential/stt.py
@@ -27,7 +27,7 @@ class XInferenceSTTModelCredential(BaseForm, BaseModelCredential):
                 else:
                     return False
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             if isinstance(e, AppApiException):
--- a/apps/models_provider/impl/xinference_model_provider/model/stt.py
+++ b/apps/models_provider/impl/xinference_model_provider/model/stt.py
@@ -17,6 +17,7 @@ class XInferenceSpeechToText(MaxKBBaseModel, BaseSpeechToText):
     api_base: str
     api_key: str
     model: str
+    params: dict
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)


apps/models_provider/impl/xinference_model_provider/credential/stt.py
```diff
@@ -27,7 +27,7 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
                 else:
                     return False
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             if isinstance(e, AppApiException):
```
shaohuzhang1, line 33: The provided code looks mostly correct but can be optimized in terms of handling keyword arguments. The line:

```python
try:

```

should be updated to correctly handle positional and keyword arguments passed to the `get_model` method. Here's an improved version:

```python
try:
    model = provider.get_model(model_type, model_name, **model_credential, **model_params)
except Exception as e:
        # Handle specific exceptions here
        pass
```

This change uses the unpacking operator (`**`) effectively to expand dictionaries into the function call, accommodating both positional and key-value pairs that might be used when calling `model.cred`. This approach ensures compatibility with different sets of input parameters without having to manually check which arguments are present.

Also, consider adding logging around exception captures to provide more context on what went wrong during the execution process.

apps/models_provider/impl/xinference_model_provider/model/stt.py
```diff
@@ -17,6 +17,7 @@ class XInferenceSpeechToText(MaxKBBaseModel, BaseSpeechToText):
     api_base: str
     api_key: str
     model: str
+    params: dict
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
```
shaohuzhang1, line 23: No significant issues found in the provided code. However, here are some minor improvements and optimizations you may consider:

1. **Docstring**: Adding a docstring to describe the parameters would be helpful.

2. **Validation**: Ensure `api_key` is correctly validated or secured (e.g., not exposed).

3. **Parameter Initialization**: Optionally, initialize default values or enforce non-empty strings for any required fields beyond `model`.

Here's an improved version with these considerations:

```python
from transformers import MaxKBBaseModel, BaseSpeechToText

class XInferenceSpeechToText(MaxKBBaseModel, BaseSpeechToText):
    """
    A speech-to-text model using X inference API.

    Parameters:
        api_base (str): The base URL of the X inference API.
        api_key (str): Your personal access token for authentication.
        model (str): Name of the speech-to-text model to use by X inference.
        params (dict): Additional parameters specific to the X inference API call.
                  Defaults to an empty dictionary if not specified.

    Attributes:
        api_base (str): The base URL of the X inference API set during initialization.
        api_key (str): Your personal access token for authentication.
        model (str): Name of the speech-to-text model used by X inference.
        params (dict): Additional parameters passed to the X inference service.
    """

    api_base: str
    api_key: str
    model: str = ""
    params: dict = {}

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

        # Validate input parameters here based on their type and constraints
        if not isinstance(api_base, str) or not api_base.strip():
            raise ValueError("Invalid api_base. It must be a non-empty string.")
        
        if not isinstance(api_key, str):
            raise ValueError("Invalid api_key. It must be a string.")

        if not isinstance(model, str) or not model.strip():
            logging.warning(f"Warning: Invalid model name '{model}'. Using '' as default.")
            self.model = ""

        if "params" in kwargs:
            if not isinstance(kwargs["params"], dict):
                raise ValueError("Invalid 'params'. Must be a dictionary.")

            self.params.update(kwargs.get("params", {}))
```

These modifications enhance clarity and robustness in handling inputs, ensuring that all necessary parameters are properly initialized and validated.

PR #4255, opened 2025-10-24
fix: Resource authorization Sub-resource authorization Parent directory cascading authorization

shaohuzhang1:
fix: Resource authorization Sub-resource authorization Parent directory cascading authorization

### DIFF
--- a/ui/src/views/system/resource-authorization/component/PermissionTable.vue
+++ b/ui/src/views/system/resource-authorization/component/PermissionTable.vue
@@ -319,7 +319,6 @@ const select = (val: any[], active: any) => {
         selectObj[active.id] = 0
       }
       if (selectObj[active.id] % 2 == 0) {
-        console.log(TreeToFlatten([active]))
         TreeToFlatten([active])
           .filter((item: any) => item.id != active.id)
           .forEach((item: any) => {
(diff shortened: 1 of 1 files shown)


ui/src/views/system/resource-authorization/component/PermissionTable.vue
```diff
@@ -371,6 +370,27 @@ function submitPermissions(value: string, row: any) {
...
+      }
+    }
+    return result
+  }
+  if (['VIEW', 'MANAGE', 'ROLE'].includes(value)) {
+    emitSubmitPermissions(props.data, [row.folder_id], obj)
+  }
   emit('submitPermissions', obj)
 }
 const provider_list = ref<Array<Provider>>([])
```
shaohuzhang1, line 396: The code looks well structured but has some optimizations and improvements that could enhance performance and readability.

1. **Optimization**:
   - In the `emitSubmitPermissions` function, you can use array methods like `.map()` and recursion directly without converting to a flat array with unnecessary steps. Consider using nested loops for more efficient manipulation of children arrays.

2. **Avoid Global State**:
   - The variables `obj`, `permissionValue`, and `rowFolderId` are defined within different scopes (`submitPermissions` vs. `props`) but used across both. This might lead to unexpected side effects if they aren't properly scoped or initialized.

3. **Variable Naming**:
   - Names such as `TreeToFlatten` and `emitterSubmitPermissions` don't clearly state what they represent. It's better to have descriptive names.

4. **Return Statement in `submitPermissions`**:
   - The `return null;` statement is not needed after emitting the event because it doesn't affect the flow. Removing it can make the code cleaner.

Here’s an improved version based on these points:

```javascript
function submitPermissions(value: string, row: any) {
  const permissionObject: { [id: number]: number } = {}
  if (value !== 'NOT_AUTH') {
    Object.assign(permissionObject, {[active.id]: 0})
  }

  const filteredItems = TreeToFlatten([active])
    .filter(item => item.id !== active.id)

  let resultArr = []

  if (['VIEW', 'MANAGE', 'ROLE'].includes(value)) {
    const emitSubmitPermissions = (treeData: any[], ids: Array<string>, result: Array<any>): void => {
      for (const node of treeData) {
        const isMatch = node.permission === 'NOT_AUTH' && ids.includes(node.id)
        if (isMatch) {
          ids.push(node.folder_id)
          result.push({target_id: node.id, permission: 'VIEW'})
        }
        if (node.children?.length > 0) {
          emitSubmitPermissions(node.children, ids.slice(), [...result]);
        }
      }
    };

    emitSubmitPermissions(props.data, [row.folder_id], []);
  }

  emit('submitPermissions', permissionObject);
}
```

### Key Changes:
- Used `Object.assign` instead of initializing with `{}` and assignment.
- Renamed functions and parameters for clarity.
- Removed unneeded variables.
- Improved loop structure for `emitSubmitPermissions`.
- Ensured the event emission logic remains simple and clear.

PR #4260, opened 2025-10-27
fix: Folder and Resource Selection Interaction Abnormality

shaohuzhang1:
fix: Folder and Resource Selection Interaction Abnormality

### DIFF
--- a/ui/src/views/system/resource-authorization/component/PermissionTable.vue
+++ b/ui/src/views/system/resource-authorization/component/PermissionTable.vue
@@ -322,9 +322,10 @@ const select = (val: any[], active: any) => {
         TreeToFlatten([active])
           .filter((item: any) => item.id != active.id)
           .forEach((item: any) => {
-            multipleTableRef.value?.toggleRowSelection(item, true)
+            if (multipleSelection.value.some((select) => item.id == select.id)) {
+              multipleTableRef.value?.toggleRowSelection(item, true)
+            }
           })
-
         multipleSelection.value = multipleTableRef.value.getSelectionRows()
       } else {
         multipleSelection.value = val


ui/src/views/system/resource-authorization/component/PermissionTable.vue
```diff
@@ -322,9 +322,10 @@ const select = (val: any[], active: any) => {
...
           .forEach((item: any) => {
-            multipleTableRef.value?.toggleRowSelection(item, true)
+            if (multipleSelection.value.some((select) => item.id == select.id)) {
+              multipleTableRef.value?.toggleRowSelection(item, true)
+            }
           })
-
         multipleSelection.value = multipleTableRef.value.getSelectionRows()
       } else {
         multipleSelection.value = val
```
shaohuzhang1, line 331: The code you've provided is mostly correct with some minor optimizations and improvements:

1. **Simplified Condition**: The condition inside the `if` statement can be simplified to directly check if an item exists in `multipleSelection`. This avoids iterating over the entire array twice.

Here's your updated code with this optimization:

```javascript
const select = (val: any[], active: any) => {
  TreeToFlatten([active])
    .filter((item: any) => item.id !== active.id)
    .forEach((item: any) => {
      const isSelected = multipleSelection.value.some(s => s.id === item.id);
      
      // Toggle row selection only if it is not already selected
      if (!isSelected && multipleTableRef.value)
        multipleTableRef.value.toggleRowSelection(item, true);
    });

  multipleSelection.value = multipleTableRef.value.getSelectionRows();
};
```

### Key Improvements:
- **Single Iteration Check**: Reduced the need for filtering out items that cannot exist in `multipleSelection`.
- **Direct Selection Check**: Use `some()` to check if an item is already selected without needing to iterate further through `multipleSelection`.

These changes make the function more efficient while maintaining its functionality.

PR #4271, opened 2025-10-27
fix: Tool permission

shaohuzhang1:
fix: Locales<br>fix: Tool permission

### DIFF
--- a/ui/src/permission/tool/workspace.ts
+++ b/ui/src/permission/tool/workspace.ts
@@ -126,8 +126,8 @@ const workspace = {
       [
         new ComplexPermission([RoleConst.USER],[PermissionConst.TOOL.getToolWorkspaceResourcePermission(source_id)],[],'AND'),
         RoleConst.WORKSPACE_MANAGE.getWorkspaceRole,
-        PermissionConst.TOOL_EXPORT.getToolWorkspaceResourcePermission(source_id),
-        PermissionConst.TOOL_EXPORT.getWorkspacePermissionWorkspaceManageRole
+        PermissionConst.TOOL_EDIT.getToolWorkspaceResourcePermission(source_id),
+        PermissionConst.TOOL_EDIT.getWorkspacePermissionWorkspaceManageRole
       ],
       'OR'
     ),
--- a/apps/locales/en_US/LC_MESSAGES/django.po
+++ b/apps/locales/en_US/LC_MESSAGES/django.po
@@ -8761,3 +8761,6 @@ msgstr ""
 
 msgid "Tag value already exists"
 msgstr ""
+
+msgid "Non-existent id"
+msgstr ""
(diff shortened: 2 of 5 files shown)


ui/src/permission/tool/workspace.ts
```diff
@@ -126,8 +126,8 @@ const workspace = {
       [
         new ComplexPermission([RoleConst.USER],[PermissionConst.TOOL.getToolWorkspaceResourcePermission(source_id)],[],'AND'),
         RoleConst.WORKSPACE_MANAGE.getWorkspaceRole,
-        PermissionConst.TOOL_EXPORT.getToolWorkspaceResourcePermission(source_id),
-        PermissionConst.TOOL_EXPORT.getWorkspacePermissionWorkspaceManageRole
+        PermissionConst.TOOL_EDIT.getToolWorkspaceResourcePermission(source_id),
+        PermissionConst.TOOL_EDIT.getWorkspacePermissionWorkspaceManageRole
       ],
       'OR'
     ),
```
shaohuzhang1, line 133: The provided TypeScript code snippet appears to be defining permission checks in a complex workflow system. Here are some observations:

1. **Permission Changes**:
   The original line has two different export permissions being added:
   ```typescript
   PermissionConst.TOOL_EXPORT.getToolWorkspaceResourcePermission(source_id)
   ```
   This was changed to use an edit instead of an export:
   ```typescript
   PermissionConst.TOOL_EDIT.getToolWorkspaceResourcePermission(source_id)
   ```

2. **Security Concerns**:
   Using `export` can potentially expose sensitive or critical data to un authorized users, especially if source_id is not properly secured.

3. **Readability**:
   The addition of duplicate `workspace.manage-role` roles could cause confusion and redundancy. It might be better to consolidate these into fewer roles if possible.

4. **Code Structure**: 
   There aren't any syntax errors but it's good practice to ensure the structure remains consistent and efficient.

**Optimization Suggestions**:
   1. **Consolidate Permissions**: If there aren't specific reasons to keep both an export and edit role, consider removing one based on the intended function.
   
   2. **Secure Source ID Usage**: Ensure that `source_id` is sanitized and appropriately checked during runtime before passing it to methods like getting tool permissions.

Here’s a revised version with those considerations:
```typescript
const workspace: {
    allowedRolesAndPermissions: Array<{
        conditions: ({ UserRole, WorkspaceRole }: typeof Role) => boolean[],
    }>,
} = {
    [
        new ComplexPermission([Role.User], [Permission.EditToolWorkspaceResourcePermission(source_id)], [], 'AND'),
        Role.WorkspaceManage,
        Permission.EditToolWorkspaceResourcePermission(source_id),
        Permission.EditToolWorkspacePermission(Role.Manage)
    ],
};
```
Note: Adjusted method names according to updated references, assuming they exist within the context of your codebase.

PR #4274, opened 2025-10-28
fix: Resource authorization: "Select All" is ineffective

shaohuzhang1:

### DIFF
--- a/ui/src/views/system/resource-authorization/component/PermissionTable.vue
+++ b/ui/src/views/system/resource-authorization/component/PermissionTable.vue
@@ -56,13 +56,14 @@
         class="mt-16"
         :data="filteredData"
         @select="select"
+        @select-all="selectAll"
         :maxTableHeight="260"
         :row-key="(row: any) => row.id"
         style="min-width: 600px"
         :expand-row-keys="defaultExpandKeys"
         show-overflow-tooltip
       >
-        <el-table-column type="selection" width="55" :reserve-selection="true" />
+        <el-table-column type="selection" width="55" :reserve-selection="true"> </el-table-column>
         <el-table-column prop="name" :label="$t('common.name')">
           <template #default="{ row }">
             <span style="vertical-align: sub">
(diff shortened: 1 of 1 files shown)


ui/src/views/system/resource-authorization/component/PermissionTable.vue
```diff
@@ -312,6 +313,9 @@ const filteredData = computed(() => {
 
 const multipleSelection = ref<any[]>([])
 const selectObj: any = {}
+const selectAll = (selection: any[]) => {
+  multipleSelection.value = selection
+}
 const select = (val: any[], active: any) => {
   if (active.resource_type === 'folder') {
     if (!val.some((item) => item.id == active.id)) {
```
shaohuzhang1, line 321: The code looks mostly correct, but there are a few minor improvements and suggestions:

1. **Suggested Changes**:
   - Replace the `@select-all` event with a more conventional name like `onSelectAll`. ThisChangeEvent makes it clear what this method does.

   - Use `nextTick()` before modifying data in lifecycle hooks that depend on DOM rendering to ensure all changes have been applied. In your `selectAll` method, you can add `await nextTick()` after setting `multipleSelection`.

Here's how the updated code could look:

```javascript
<template>
  <div>
    <!-- Component template remains unchanged -->
  </div>
</template>

<script setup lang="ts">
import { ref, onMounted, computed, nextTick } from 'vue' // Use import statement instead of require
import type { Provider } from '@/api/type/model'
import { SourceTypeEnum } from '@/enums/common'

// Existing code logic remains intact

const filteredData = computed(() => {
  return data.filter(item => !showTrash || item.is_trash !== true);
});

const multipleSelection = ref<any[]>([]);
const selectObj: any = {};
// Update onSelectAll event handler
const onSelectAll = (selection: any[]) => {
  await nextTick(); // Ensure DOM updates
  multipleSelection.value = selection;
};
// Updated select function
const select = (val: any[], active: any) => {
  if (active.resource_type === 'folder') {
    if (!val.some((item) => item.id == active.id)) {
      val.push(active);
    }
  } else {
    if (!val.includes(active.row)) { // Corrected comparison logic
      val.push(Object.assign({}, active));
      
    }

// ... Rest of the update ...

```

These changes should make the code cleaner and potentially enhance performance depending on the complexity of your application.

PR #4276, opened 2025-10-28
fix: When extracting parameters, input parameters that are not strings result in the node being unable to execute

shaohuzhang1:
fix: When extracting parameters, input parameters that are not strings result in the node being unable to execute

### DIFF
--- a/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
+++ b/apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
@@ -91,6 +91,7 @@ class BaseParameterExtractionNode(IParameterExtractionNode):
         self.context['request'] = details.get('request')
 
     def execute(self, input_variable, variable_list, model_params_setting, model_id, **kwargs) -> NodeResult:
+        input_variable = str(input_variable)
         self.context['request'] = input_variable
         if model_params_setting is None:
             model_params_setting = get_default_model_params_setting(model_id)


apps/application/flow/step_node/parameter_extraction_node/impl/base_parameter_extraction_node.py
```diff
@@ -91,6 +91,7 @@ def save_context(self, details, workflow_manage):
         self.context['request'] = details.get('request')
 
     def execute(self, input_variable, variable_list, model_params_setting, model_id, **kwargs) -> NodeResult:
+        input_variable = str(input_variable)
         self.context['request'] = input_variable
         if model_params_setting is None:
             model_params_setting = get_default_model_params_setting(model_id)
```
shaohuzhang1, line 97: The provided code has a few issues that need to be addressed:

1. **Type Casting Issue**: The line `input_variable = str(input_variable)` casts the input parameter regardless of its original type. This can lead to unexpected behavior if the input is not a string, such as numbers or other data types.

   **Suggestion**: Consider using implicit conversion instead of explicit casting if possible, but avoid this if you want to ensure consistency with the variable's expected type before use. If necessary, introduce a more targeted approach based on your application requirements.

2. **Code Style Mismatch**:
   - Line 86 contains an extra space after `-`. In Python, consecutive whitespace characters are considered a single space.
   
3. **Function Documentation**: Ensure that all functions have appropriate docstrings (Python documentation). While the function doesn't require much in terms of explanation given its simplicity, it's a good practice to maintain clarity.

Here's the optimized version of the code without addressing the type-casting issue for now:

```python
class MyClass:
    def __init__(self):
        self.context = {}

        self.context['request'] = details.get('request')

    def execute(self, input_variable, variable_list, model_params_setting, model_id, **kwargs) -> 'NodeResult':
        if model_params_setting is None:
            model_params_setting = get_default_model_params_setting(model_id)
```

Addressing the type-casting issue requires more context about how and where this method is used within your larger application to determine the most appropriate handling strategy.

PR #4282, opened 2025-10-29
fix: When the workflow does not respond, there is an error in the historical chat record data

shaohuzhang1:
fix: When the workflow does not respond, there is an error in the historical chat record data

### DIFF
--- a/apps/application/models/application_chat.py
+++ b/apps/application/models/application_chat.py
@@ -9,7 +9,7 @@
 import uuid_utils.compat as uuid
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext as _
 from langchain_core.messages import HumanMessage, AIMessage
 
 from application.models import Application


apps/application/models/application_chat.py
```diff
@@ -9,7 +9,7 @@
 import uuid_utils.compat as uuid
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext as _
 from langchain_core.messages import HumanMessage, AIMessage
 
 from application.models import Application
```
shaohuzhang1, line 15: The provided code appears generally correct with one minor issue to note:

1. **Import Statements**: The `gettext` function is used instead of `gettext_lazy`. While this works fine for simple applications where translation isn't a significant concern, it might not be ideal if the application needs better performance in larger scale or more complex scenarios.

   Instead, `gettext_lazy` should be preferred over `gettext` when translating because it allows for lazy loading of translations and can improve memory usage by avoiding unnecessary localization until translation strings are accessed.

**Optimization Suggestion**: Use `gettext_lazy`.

Here's how you can fix it:

```diff
@@ -9,7 +9,7 @@
 import uuid_utils.compat as uuid
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _

 from application.models import Application
```

This change will ensure that the `gettext` function is used correctly for string interpolation and internationalization purposes in your Django application.

PR #4285, opened 2025-10-29
fix: In the variable aggregation strategy that returns the first non-null value of each group, empty strings, empty lists, and empty objects are not treated as null values

shaohuzhang1:
fix: In the variable aggregation strategy that returns the first non-null value of each group, empty strings, empty lists, and empty objects are not treated as null values

### DIFF
--- a/apps/application/flow/step_node/variable_aggregation_node/impl/base_variable_aggregation_node.py
+++ b/apps/application/flow/step_node/variable_aggregation_node/impl/base_variable_aggregation_node.py
@@ -25,7 +25,7 @@ class BaseVariableAggregationNode(IVariableAggregation):
             v = self.workflow_manage.get_reference_field(
                 variable.get('variable')[0],
                 variable.get('variable')[1:])
-            if v is not None:
+            if v is not None and not(isinstance(v, (str,list,dict)) and len(v) == 0) :
                 return v
         return None
 


apps/application/flow/step_node/variable_aggregation_node/impl/base_variable_aggregation_node.py
```diff
@@ -25,7 +25,7 @@ def get_first_non_null(self, variable_list):
             v = self.workflow_manage.get_reference_field(
                 variable.get('variable')[0],
                 variable.get('variable')[1:])
-            if v is not None:
+            if v is not None and not(isinstance(v, (str,list,dict)) and len(v) == 0) :
                 return v
         return None
```
shaohuzhang1, line 31: The provided code has a few minor issues and improvements that can be made:

1. **Redundant Condition**: The current condition `if isinstance(v, (str,list,dict)) and len(v) == 0` is redundant because an empty string (`''`) or other falsy values won't evaluate to `True`. Therefore, you can simplify this part of the check to just `v is not None`.

2. **Line Length**: There's one line too long, which might make it harder to read. You can break the line using parentheses or a backslash.

3. **Whitespace Improvements**: Ensure consistent indentation to maintain readability.

Here's the improved version:

```python
def get_first_non_null(self, variable_list):
        ref_var = self.workflow_manage.get_reference_field(
            variable.get('variable')[0],
            variable.get('variable')[1:]
        )
        if ref_var is not None:
            return ref_var
    return None
```

Note: This assumes that `ref_var` returned from `get_reference_field()` cannot be considered "null" in the sense defined by checking its truthiness after stripping whitespace if needed (though strings are inherently non-empty in Python). If that assumption changes, additional checks should be implemented accordingly.

PR #4291, opened 2025-10-30
refactor: update ZhipuChatModel to use BaseChatOpenAI and improve token counting

shaohuzhang1:
refactor: update ZhipuChatModel to use BaseChatOpenAI and improve token counting  --bug=1061305 --user=刘瑞斌 【应用】ai对话启用工具后部分模型（智谱）不统计tokens https://www.tapd.cn/62980211/s/1791683

### DIFF
--- a/apps/models_provider/impl/zhipu_model_provider/model/llm.py
+++ b/apps/models_provider/impl/zhipu_model_provider/model/llm.py
@@ -7,27 +7,21 @@
     @desc:
 """
 
-import json
-from collections.abc import Iterator
-from typing import Any, Dict, List, Optional
+from typing import Dict, List
 
-from langchain_community.chat_models import ChatZhipuAI
-from langchain_community.chat_models.zhipuai import _truncate_params, _get_jwt_token, connect_sse, \
-    _convert_delta_to_message_chunk
-from langchain_core.callbacks import (
-    CallbackManagerForLLMRun,
-)
-from langchain_core.messages import (
-    AIMessageChunk,
-    BaseMessage
-)
-from langchain_core.outputs import ChatGenerationChunk
+from langchain_core.messages import BaseMessage, get_buffer_string
 
+from common.config.tokenizer_manage_config import TokenizerManage
 from models_provider.base_model_provider import MaxKBBaseModel
+from models_provider.impl.base_chat_open_ai import BaseChatOpenAI
 
 
-class ZhipuChatModel(MaxKBBaseModel, ChatZhipuAI):
-    optional_params: dict
+def custom_get_token_ids(text: str):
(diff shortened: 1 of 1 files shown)


apps/models_provider/impl/zhipu_model_provider/model/llm.py
```diff
@@ -39,69 +33,23 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
...
-                    yield chunk
-                    if run_manager:
-                        run_manager.on_llm_new_token(chunk.text, chunk=chunk)
-                    if finish_reason is not None:
-                        break
+        try:
+            return super().get_num_tokens(text)
+        except Exception as e:
+            tokenizer = TokenizerManage.get_tokenizer()
+            return len(tokenizer.encode(text))
```
shaohuzhang1, line 55: No significant changes have been identified in the provided code snippet, but here are some minor improvements or clarifications:

- Remove `self.optional_params` assignment as it's not used anywhere in the class.
  
- Adjust the usage of token counting methods to use consistent approach between classes.

Here's a revised version of the code with these corrections applied:

```python
# Import necessary libraries and classes
import json

from typing import Dict, List

from langchain_core.messages import BaseMessage


class ZhipuChatModel(MaxKBBaseModel, BaseChatOpenAI):

    @staticmethod
    def is_cache_model():
        pass
    
    def __init__(self, *args, max_length=None, **kwargs):
        self.max_length = max_length

    def _token_count(self, text: str) -> int:
            # Use the default method provided by BaseChatOpenAI if available
            try:
                return super().get_num_tokens(text)
            except Exception as e:
                tokens = tokenizer.encode(text)
                return min(len(tokens), self.max_length) if max_length else len(tokens)

    def generate_responses(self, messages: List[BaseMessage]) -> List[str]:
        total_tokens = 0
        responses = []
        for message in messages:
            token_count = self._token_count(message.content)
            total_tokens += token_count
            
            # Perform your logic here based on whether you want to split or concatenate input data
        
        # Return all generated responses or handle further processing

```

### Key Changes:
1. **Initialization**: Added `max_length` parameter during initialization and set it to an instance variable.
   
2. **Token Count Calculation**: Modified `_token_count` method to first call the superclass’s method (`super().get_num_tokens(text)`), which should be more reliable than manually encoding texts via TokenizerManage (since it might have additional optimizations).

3. **Replaced Custom Method**: Removed the custom `custom_get_token_ids` method since we can now directly get a list of IDs using the same technique.

This change will ensure that the token count reflects the actual input length rather than potentially over-counting due to the use of multiple encoders. This may make future modifications easier without changing the core logic too much.

PR #4295, opened 2025-10-30
fix: Migration

shaohuzhang1:
fix: Migration

### DIFF
--- a/apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
+++ b/apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
@@ -64,6 +64,9 @@ def auth_folder(apps, schema_editor):
 class Migration(migrations.Migration):
     dependencies = [
         ('system_manage', '0002_refresh_collation_reindex'),
+        ('tools', '0001_initial'),
+        ('application', '0001_initial'),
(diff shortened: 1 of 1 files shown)


apps/system_manage/migrations/0003_alter_workspaceuserresourcepermission_target.py
```diff
@@ -64,6 +64,9 @@ def auth_folder(apps, schema_editor):
 class Migration(migrations.Migration):
     dependencies = [
         ('system_manage', '0002_refresh_collation_reindex'),
+        ('tools', '0001_initial'),
+        ('application', '0001_initial'),
+        ('knowledge', '0001_initial'),
     ]
 
     operations = [
```
shaohuzhang1, line 72: Yes, there appears to be an error in this migration file. The dependencies list should not include apps that are not required by the models being modified. Specifically:

1. There is no `system_manage.models` mentioned.
2. Although 'tool' app was previously added, it seems like an unnecessary dependency.

The correct line should be without the entries for 'tools' and 'knowledge':

```python
dependencies = [
    ('system_manage', '0002_refresh_collation_reindex'),
]
```

Also, ensure that none of these dependencies have already been satisfied earlier in the migrations history (in previous versions), or you might cause circular references if they depend on each other.

PR #4297, opened 2025-10-31
fix: XF stt model

shaohuzhang1:
fix: XF stt model

### DIFF
--- a/apps/models_provider/impl/xinference_model_provider/model/stt.py
+++ b/apps/models_provider/impl/xinference_model_provider/model/stt.py
@@ -41,6 +41,7 @@ class XInferenceSpeechToText(MaxKBBaseModel, BaseSpeechToText):
             model=model_name,
             api_base=model_credential.get('api_base'),
             api_key=model_credential.get('api_key'),
+            params=model_kwargs,
             **optional_params,
         )
 


apps/models_provider/impl/xinference_model_provider/model/stt.py
```diff
@@ -41,6 +41,7 @@ def new_instance(model_type, model_name, model_credential: Dict[str, object], **
             model=model_name,
             api_base=model_credential.get('api_base'),
             api_key=model_credential.get('api_key'),
+            params=model_kwargs,
             **optional_params,
         )
```
shaohuzhang1, line 47: The provided code snippet appears to be a method for creating an instance of a model based on certain parameters and keyword arguments. However, there are a few areas where the code can be improved:

1. **Type Hinting**: Ensure that type hints are complete and accurate. The `Dict[str, object]` is not very descriptive. It would be better if you could specify more concrete types for the keys and values.

2. **Optional Parameters**: Make sure all optional parameters (`api_base`, `api_key`) have default values defined. This will prevent errors when calling the function without specifying these parameters unless required.

3. **Suggestion for Adding Params Keyword Argument**: If you intended to include additional configuration options for the model instantiation, consider adding a specific parameter name (e.g., `config`) instead of using an arbitrary key `model_kwargs`. This makes it clearer what kind of data is expected here.

Here's a revised version of the code with these suggestions incorporated:

```python
def new_instance(
    model_type: str,
    model_name: str,
    model_credential: Dict[str, Any],
    api_base: Optional[str] = None,  # Added default value
    api_key: Optional[str] = None,  # Added default value
    config: Optional[Dict[str, Any]] = None,  # Suggested new parameter
    **optional_params,
):
```

**Explanation**:
- **Type Hints**: Updated type hinting to use `str` for `model_type` and `model_name` and `Dict[str, Any]` for `model_credential` and other dictionaries.
- **Default Values**: Provided default values for `api_base` and `api_key`.
- **Additional Parameter**: Introduced a new parameter `config` to handle potential additional configuration settings.
This revision aims to enhance clarity, maintainability, and future-proofing the function by ensuring type consistency and providing clear usage guidelines through default parameters.

PR #4298, opened 2025-10-31
fix: Unable to export dialogue log data due to error

shaohuzhang1:
fix: Unable to export dialogue log data due to error

### DIFF
--- a/apps/application/serializers/application_chat.py
+++ b/apps/application/serializers/application_chat.py
@@ -151,7 +151,7 @@ class ApplicationChatQuerySerializers(serializers.Serializer):
     @staticmethod
     def to_row(row: Dict):
         details = row.get('details') or {}
-        padding_problem_text = ' '.join(node.get("answer", "") for key, node in details.items() if
+        padding_problem_text = ' '.join((node.get("answer", "") or "") for key, node in details.items() if
                                         node.get("type") == 'question-node')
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_chat.py
```diff
@@ -151,7 +151,7 @@ def paragraph_list_to_string(paragraph_list):
     @staticmethod
     def to_row(row: Dict):
         details = row.get('details') or {}
-        padding_problem_text = ' '.join(node.get("answer", "") for key, node in details.items() if
+        padding_problem_text = ' '.join((node.get("answer", "") or "") for key, node in details.items() if
                                         node.get("type") == 'question-node')
         search_dataset_node_list = [(key, node) for key, node in details.items() if
                                     node.get("type") == 'search-dataset-node' or node.get(
```
shaohuzhang1, line 157: The provided code has a small issue with string formatting. The use of `" ".join(...)` is correct, but using `or ""` will result in an extra space between strings when there are non-empty answers. To avoid this, you could modify the line inside the join to:

```python
... ("" if node.get("answer") else "") for ...
```

This way, if there's no answer (`None'), it skips adding extra whitespace.

Additionally, while there are no significant performance improvements or obvious issues in the basic functionality of the method, consider using list comprehensions and generator expressions more efficiently. However, the current implementation seems straightforward enough, so these might be minor optimizations.

PR #4302, opened 2025-11-03
fix: The maximum limit for application dialogue has been modified to 10000000 times

shaohuzhang1:
fix: The maximum limit for application dialogue has been modified to 10000000 times

### DIFF
--- a/apps/application/serializers/application_access_token.py
+++ b/apps/application/serializers/application_access_token.py
@@ -24,7 +24,7 @@ class AccessTokenEditSerializer(serializers.Serializer):
     access_token_reset = serializers.BooleanField(required=False,
                                                   label=_("Reset Token"))
     is_active = serializers.BooleanField(required=False, label=_("Is it enabled"))
-    access_num = serializers.IntegerField(required=False, max_value=10000,
+    access_num = serializers.IntegerField(required=False, max_value=10000000,
                                           min_value=0,
                                           label=_("Number of visits"))
     white_active = serializers.BooleanField(required=False,
--- a/ui/src/views/application-overview/xpack-component/XPackLimitDrawer.vue
+++ b/ui/src/views/application-overview/xpack-component/XPackLimitDrawer.vue
@@ -16,7 +16,7 @@
           v-model="form.access_num"
           :min="0"
           :step="1"
-          :max="10000"
+          :max="10000000"
           :value-on-clear="0"
           controls-position="right"
           style="width: 268px"


apps/application/serializers/application_access_token.py
```diff
@@ -24,7 +24,7 @@ class AccessTokenEditSerializer(serializers.Serializer):
     access_token_reset = serializers.BooleanField(required=False,
                                                   label=_("Reset Token"))
     is_active = serializers.BooleanField(required=False, label=_("Is it enabled"))
-    access_num = serializers.IntegerField(required=False, max_value=10000,
+    access_num = serializers.IntegerField(required=False, max_value=10000000,
                                           min_value=0,
                                           label=_("Number of visits"))
     white_active = serializers.BooleanField(required=False,
```
shaohuzhang1, line 30: The `access_num` field's `max_value` is set to 1,000, which may not be the intended constraint. Typically, an access number should be higher or at least manageable depending on usage requirements (e.g., one million accesses per token). Here's a revised version:

1. **Increase Maximum Value**: The maximum value can be increased based on expected usage.
2. **Documentation Update**: A note about this adjustment should be included in the Django REST Framework documentation.

```python
class AccessTokenEditSerializer(serializers.Serializer):
    """
        A serializer for editing access tokens with additional fields such as reset, activation status, and visit count.
    """

    # ...
```

**Revision Summary:** Increase `max_value` from 1,000 to a more suitable upper bound like 10,000,000 or even higher if needed. This change ensures the field's correctness according to typical use cases without risking database constraints exceeding limits.

ui/src/views/application-overview/xpack-component/XPackLimitDrawer.vue
```diff
@@ -16,7 +16,7 @@
           v-model="form.access_num"
           :min="0"
           :step="1"
-          :max="10000"
+          :max="10000000"
           :value-on-clear="0"
           controls-position="right"
           style="width: 268px"
```
shaohuzhang1, line 22: In the provided code snippet, there is one change to optimize performance and reduce user input errors:

```diff
-          :max="10000"
+          :max="10000000"
```

**Modification Explanation**: The `max` prop from 10,000 to 10,000,000 increases the range of values that users can enter into the input field. This allows for more flexible data entry without exceeding typical limits (e.g., for numerical IDs). Adjusting this value ensures that potential future requirements or constraints are considered. 

This change does not introduce irregularities or significant issues, but it enhances the flexibility of the form component by accommodating larger inputs, which could be beneficial depending on specific use cases.

PR #4305, opened 2025-11-03
chore: update document deletion logic to handle paragraph IDs and related mappings

shaohuzhang1:


apps/knowledge/serializers/document.py
```diff
@@ -1217,9 +1218,12 @@ def batch_delete(self, instance: Dict, with_valid=True):
...
             QuerySet(DocumentTag).filter(document_id__in=document_id_list).delete()
-            delete_problems_and_mappings(document_id_list)
+            paragraph_ids = QuerySet(Paragraph).filter(document_id__in=document_id_list).values_list("id", flat=True)
+            # 删除问题关系
+            delete_problems_and_mappings(paragraph_ids)
+            # 删除段落
+            QuerySet(Paragraph).filter(document_id__in=document_id_list).delete()
             # 删除向量库
             delete_embedding_by_document_list(document_id_list)
             return True
```
shaohuzhang1, line 1229: There is an issue in the `batch_delete` method where multiple paragraph IDs related to each document need to be processed instead of just using single document ID for problem deletion. Here's how you can fix it:

```python
# ... (previous unchanged code)

def batch_delete(instance: Dict, with_valid=True):
    document_id_list = [item.get('id') for item in instance]
    
    # Assuming 'instance' contains metadata about documents
    meta_data_querysets = [
        Document.objects.filter(id__in=document_id_list).values("meta")
    ]
    
    QuerySet(File).filter(id__in=source_file_ids).delete()
    QuerySet(Document).filter(id__in=document_id_list).delete()
    
    paragraph_ids_dict = {}
    for doc_id in document_id_list:
        paragraphs_queryset = Paragraph.objects.filter(document_id=doc_id) \
                                           .annotate(problem_count=models.Count('problem')) \
                                           .order_by('-problem_count')
        
        if paragraphs_queryset.count() > 0:
            paragraph_id_set = set(paragraphs_queryset.annotate(
                problem_id=models.F('problemtypes__pk')).exclude(problemtypes=None)
                                            ).values_list('id', flat=True)
            
            paragraph_ids_dict[doc_id] = list(set(paragraph_id_set))
            delete_problems_and_mappings(list(map(int, paragraph_id_set)))
            QuerySet(Paragraph).filter(document_id=doc_id).delete()
    
    delete_embedding_by_document_list(document_id_list)
    return True

```

### Optimizations/Suggestions:
1. **Avoid Deleting Multiple Times**: The code currently deletes segments three times. This redundancy can be eliminated by only calling `QuerySet.delete()` once per queryset after all necessary filtering is done.
  
2. **Batch Processing**: Using bulk operations whenever possible is more efficient than making many individual database queries.
   
3. **Error Handling**: Consider adding error handling around network requests, file I/O, etc., as mentioned in a previous conversation.

4. **Performance Improvements**: Optimize SQL queries such as those involving annotations and join operations. For example, consider caching query results when they do not change frequently. 

These changes should improve performance and correctness of your methods.

PR #4324, opened 2025-11-06
fix: Rearrange model dependency

shaohuzhang1:
fix: Rearrange model dependency

### DIFF
--- a/apps/models_provider/impl/local_model_provider/model/reranker/model.py
+++ b/apps/models_provider/impl/local_model_provider/model/reranker/model.py
@@ -11,7 +11,6 @@ from typing import Sequence, Optional, Dict, Any
 
 from langchain_core.callbacks import Callbacks
 from langchain_core.documents import Document, BaseDocumentCompressor
-from transformers import AutoModelForSequenceClassification, AutoTokenizer
 
 from models_provider.base_model_provider import MaxKBBaseModel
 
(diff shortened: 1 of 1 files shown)


apps/models_provider/impl/local_model_provider/model/reranker/model.py
```diff
@@ -25,6 +24,7 @@ class LocalReranker(MaxKBBaseModel, BaseDocumentCompressor):
 
     def __init__(self, model_name, cache_dir=None, **model_kwargs):
         super().__init__()
+        from transformers import AutoModelForSequenceClassification, AutoTokenizer
         self.model = model_name
         self.cache_dir = cache_dir
         self.model_kwargs = model_kwargs
```
shaohuzhang1, line 30: The provided code checks out with no immediate errors or issues related to syntax, logic, or imports. The initialization of `AutoModelForSequenceClassification` and `AutoTokenizer` is correctly placed after the necessary libraries have been imported.

**Optimization Suggestion**: Ensure that the paths used in file operations (like reading data) are correct relative to where the script is being run, as this can lead to `FileNotFoundError`.

Additionally, consider handling exceptions such as when attempting to load models to avoid silent failures:

```python
from transformers import AutoModelForSequenceClassification, AutoTokenizer

class LocalReranker(MaxKBBaseModel, BaseDocumentCompressor):
    __name__ = "LocalReranker"

    def __init__(self, model_name : str, cache_dir: Optional[str] = None, **model_kwargs):
        super().__init__()
        
        try:
            from transformers import AutoModelForSequenceClassification, AutoTokenizer
        except ImportError as e:
            raise SystemError(f"Failed to import required transformers library(s): {str(e)}")

        self.model_name = model_name
        self.cache_dir = cache_dir if cache_dir else "/path/to/default/cache/dir"
        self.model_kwargs = model_kwargs
        
        # Initialize or load the model here
```

This modification adds basic error checking for missing dependencies and specifies a default path for caching models. Adjust the path according to how you set up your environment.

PR #4328, opened 2025-11-06
fix: dependency package

shaohuzhang1:
fix: dependency package

### DIFF
--- a/apps/maxkb/wsgi/web.py
+++ b/apps/maxkb/wsgi/web.py
@@ -6,14 +6,34 @@
     @date：2025/11/5 15:14
     @desc:
 """
+import builtins
 import os
+import sys
 
 from django.core.wsgi import get_wsgi_application
+
+
+class TorchBlocker:
+    def __init__(self):
+        self.original_import = builtins.__import__
+
+    def __call__(self, name, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


apps/maxkb/wsgi/web.py
```diff
@@ -29,9 +49,10 @@ def post_scheduler_handler():
...
 
+
 # 启动后处理函数
 post_handler()
 
 # 仅在scheduler中启动定时任务，dev local_model celery 不需要
 if os.environ.get('ENABLE_SCHEDULER') == '1':
-    post_scheduler_handler()
\ No newline at end of file
+    post_scheduler_handler()
```
shaohuzhang1, line 58: No significant issues were found in the provided code. The changes made to intercept and disable imports of popular machine learning libraries like `torch` could potentially be beneficial in environments where these should not be used due to security concerns, performance penalties, or dependency conflicts with other applications running on the same system.

However, please note that this change may require adjustments depending on whether you want certain parts of your Django app (e.g., using PyTorch modules directly during development) to work without restrictions imposed by this interceptor. You will also need ensure compatibility across different versions and configurations of software dependencies. Finally, consider documenting the impact of this change within your project documentation for users who might use it.

PR #4329, opened 2025-11-07
feat: Using the HuggingFace tokenizer

shaohuzhang1:
feat: Using the HuggingFace tokenizer


apps/common/config/tokenizer_manage_config.py
```diff
@@ -6,19 +6,27 @@
...
-                cache_dir="/opt/maxkb-app/model/tokenizer",
-                local_files_only=True,
-                resume_download=False,
-                force_download=False)
-        return TokenizerManage.tokenizer
+        from tokenizers import Tokenizer
+        # 创建Tokenizer
+        s = os.path.join(BASE_DIR.parent, 'tokenizer', 'bert-base-cased', 'tokenizer.json')
+        TokenizerManage.tokenizer = Tokenizer.from_file(s)
+        return MKTokenizer(TokenizerManage.tokenizer)
```
shaohuzhang1, line 32: ### Code Review and Suggestions

1. **Imports**:
   - Ensure all necessary imports are at the top for clarity.

2. **Base Directory**:
   - The use of `os.path.join` along with `Path()` simplifies path handling and ensures cross-platform compatibility.

3. **MKTokenizer Class**:
   - The `tokenize` method can be simplified to just returning `tokenizer.encode(text)` without wrapping it inside another list.

4. **TokenizerManage Class**:
   - Simplify the creation process by directly loading the tokenizer's parameters into `Tokenizer`.
   - Create an instance of `MKTokenizer` using the loaded tokenizer within the `get_tokenizer` method.
   - Remove unnecessary comments that may clutter the code.

5. **File Paths**:
   - Ensure that paths relative to `__file__` are correctly calculated. The current absolute path (`BASE_DIR`) should point to the directory containing this file, not its parent or grandparent directories twice.

6. **Static Method Considerations**:
   - Since the tokenizer might change during runtime (e.g., reloading based on user input or configuration), consider making this non-static or moving to a different pattern if changes are expected frequently.

### Revised Code:

```python
@@ -6,15 +6,21 @@
 @date：2024/4/28 10:17
 @desc:
 """

+import os
+from pathlib import Path
+
+BASE_DIR = Path(__file__).resolve().parent
+
+
+class MKTokenizer:
+    def __init__(self, tokenizer):
+        self.tokenizer = tokenizer
        
+    def encode(self, text):
+        return self.tokenizer.encode(text)


class TokenizerManage:
    tokenizer = None
    
    @staticmethod
    def get_tokenizer():
+        from transformers import BertTokenizer  
        # Load tokenizer params from file
-        path = "/opt/maxkb-app/model/tokenizer/bert-base-cased"
+        s = os.path.join(BASE_DIR.parent, 'tokenizer', 'bert-base-cased', 'tokenizer.json')
        
        TokenizerManage.tokenizer = BertTokenizer.from_file(s)
        return MKTokenizer(TokenizerManage.tokenizer)

```

By applying these improvements, the code will be cleaner, more maintainable, and adhere more closely to best practices in Python programming paradigms.

apps/models_provider/impl/tencent_cloud_model_provider/model/llm.py
```diff
@@ -6,9 +6,7 @@
     @date：2024/4/18 15:28
     @desc:
 """
-from typing import List, Dict
-
-from langchain_core.messages import BaseMessage, get_buffer_string
+from typing import Dict
 
 from common.config.tokenizer_manage_config import TokenizerManage
 from models_provider.base_model_provider import MaxKBBaseModel
```
shaohuzhang1, line 12: The provided Python code has a minor issue that you may want to address:

1. **Redundant Import**: You're importing `List`, which is imported by default if using `typing`. Therefore, the import of `from typing import List` can be removed.

Here's an updated version with the redundant import removed:

```python
@@ -6,12 +6,7 @@
     @date：2024/4/18 15:28
     @desc:
 """
-import typing
-from typing import List, Dict

from models_provider.base_model_provider import MaxKBBaseModel

# Remove the redundant import of 'List'
```

This change makes your code cleaner and eliminates unnecessary redundancy. If there are no other issues, this will suffice.

PR #4337, opened 2025-11-10
fix: When having read permission for the application, the application settings page can be viewed

shaohuzhang1:
fix: When having read permission for the application, the application settings page can be viewed

### DIFF
--- a/ui/src/router/modules/application-detail.ts
+++ b/ui/src/router/modules/application-detail.ts
@@ -82,13 +82,13 @@ const ApplicationDetailRouter = {
           () => {
             const to: any = get_next_route()
             if (to.params.from == 'resource-management') { } else {
-              return PermissionConst.APPLICATION_EDIT.getWorkspacePermissionWorkspaceManageRole()
+              return PermissionConst.APPLICATION_READ.getWorkspacePermissionWorkspaceManageRole()
             }
           },
           () => {
             const to: any = get_next_route()
             if (to.params.from == 'resource-management') { } else {
-              return PermissionConst.APPLICATION_EDIT.getApplicationWorkspaceResourcePermission(
+              return PermissionConst.APPLICATION_READ.getApplicationWorkspaceResourcePermission(
                 to ? to.params.id : '',
               )
             }
@@ -99,7 +99,7 @@ const ApplicationDetailRouter = {
           },
           () => {
             const to: any = get_next_route()
-            if (to.path.includes('resource-management')) { return PermissionConst.RESOURCE_APPLICATION_EDIT }
+            if (to.path.includes('resource-management')) { return PermissionConst.RESOURCE_APPLICATION_READ }
           },
         ]
       },


ui/src/router/modules/application-detail.ts
```diff
@@ -99,7 +99,7 @@ const ApplicationDetailRouter = {
           },
           () => {
             const to: any = get_next_route()
-            if (to.path.includes('resource-management')) { return PermissionConst.RESOURCE_APPLICATION_EDIT }
+            if (to.path.includes('resource-management')) { return PermissionConst.RESOURCE_APPLICATION_READ }
           },
         ]
       },
```
shaohuzhang1, line 105: The provided code has several potential issues:

1. **Function `get_next_route` Usage**:
   - The function `get_next_route()` is not defined in the given snippet. Without its implementation, the logic for determining permissions will not be functional.

2. **String Matching in Conditional Logic**:
   - In both cases where `to.params.from == 'resource-management'`, nothing significant is done in either return statement. This could cause unexpected behavior if you intend to conditionally apply different policies within these routes.
   
3. **Permissions Return Values**:
   - While it appears that `PermissionConst.APPLICATION_READ` should work similarly to `PermissionConst.APPLICATION_EDIT`, using `PermissionConst.RESOURCE_APPLICATION_READ` instead of `PermissionConst.APPLICATION_READ` when handling resource management routes might inadvertently allow more access than intended due to differences between application read and resource edit permissions.

4. **Path Handling**:
   - When checking paths with `if (to.path.includes('resource-management'))`, this approach directly checks against partial strings (`includes`). It’s generally better to compare full URLs or specific path segments to ensure accuracy and prevent false positives.

### Suggestions:

- Ensure that `get_next_route()` returns an object containing necessary route information if the dependency is needed elsewhere in your application.
- Evaluate whether conditional logic based on `from` params is actually required. If all resources fall under "application," perhaps simplifying the check would suffice: `if (!to.path.startsWith('/resource-management'))`. This reduces complexity without changing permissions unnecessarily.
- Double-check the definitions/assignments of `PermissionConst` constants to maintain consistency across your project.
  
To address some of these concerns, here's a revised version of one of the affected sections assuming proper dependencies and correct usage contexts:

```javascript
const ApplicationDetailRouter = {
  [RouteType.Application]: [
    {
      method: RouteMethod.Get,
      controllerName: 'ApplicationController',
      methodName: 'GetApplicationById',
      authorizationRuleFn: (
        _authInfo: AuthResultInterface | null,
        context: ControllerContext<Request<any>, Response>
      ) => {
        const to: any = context.req.url // Assuming req.url provides enough info
        if (to && to.includes("resource-management")) {
          return PermissionConst.RESOURCE_APPLICATION_READ;
        }
        return PermissionConst.APPLICATION_READ; // Default permission for everything else
      }
    },
    // Repeat similar pattern for other methods/routes as applicable...
  ],
}
```

Make sure that this revision aligns correctly with your actual architecture and data flow, including how request objects are structured.

PR #4362, opened 2025-11-14
perf: Memory optimization

shaohuzhang1:
perf: Memory optimization

### DIFF
--- a/apps/application/flow/i_step_node.py
+++ b/apps/application/flow/i_step_node.py
@@ -94,6 +94,7 @@ class WorkFlowPostHandler:
                 application_public_access_client.access_num = application_public_access_client.access_num + 1
                 application_public_access_client.intraday_access_num = application_public_access_client.intraday_access_num + 1
                 application_public_access_client.save()
+        self.chat_info = None
 
 
 class NodeResult:
--- a/apps/application/serializers/common.py
+++ b/apps/application/serializers/common.py
@@ -6,6 +6,7 @@
     @date：2025/6/9 13:42
     @desc:
 """
+import json
 from typing import List
 
 from django.core.cache import cache
@@ -19,6 +20,7 @@ from application.models import Application, ChatRecord, Chat, ApplicationVersion
 from application.serializers.application_chat import ChatCountSerializer
 from common.constants.cache_version import Cache_Version
 from common.database_model_manage.database_model_manage import DatabaseModelManage
+from common.encoder.encoder import SystemEncoder
 from common.exception.app_exception import ChatException
 from knowledge.models import Document
 from models_provider.models import Model
@@ -234,7 +236,7 @@ class ChatInfo:
             'knowledge_id_list': self.knowledge_id_list,
             'exclude_document_id_list': self.exclude_document_id_list,
             'application_id': self.application_id,
-            'chat_record_list': [self.chat_record_to_map(c) for c in self.chat_record_list],
+            'chat_record_list': [self.chat_record_to_map(c) for c in self.chat_record_list][-20:],
             'debug': self.debug
         }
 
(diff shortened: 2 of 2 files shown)


apps/application/serializers/common.py
```diff
@@ -276,15 +278,18 @@ def set_cache(self):
...
+        return c
 
     @staticmethod
     def get_cache(chat_id):
-        chat_info_dict = cache.get(Cache_Version.CHAT.get_key(key=chat_id), version=Cache_Version.CHAT_INFO.get_version())
+        chat_info_dict = cache.get(Cache_Version.CHAT.get_key(key=chat_id),
+                                   version=Cache_Version.CHAT_INFO.get_version())
         if chat_info_dict:
             return ChatInfo.map_to_chat_info(chat_info_dict)
         return None
```
shaohuzhang1, line 295: There are no major issues with the provided code. However, a few optimizations can be made:

1. **String Interpolation**: Use f-string formatting in Python's `format()` method to avoid concatenation of strings repeatedly.

2. **Method Call Optimization**: If `map_to_chat_info` is called multiple times within another function or class, consider caching its result instead of recomputing it each time.

Here is an optimized version of the relevant part:

```python
class YourClassName:
    @staticmethod
    def map_to_chat_info(chat_info_dict):
        # Using fstring for string interpolation
        c = ChatInfo(
            chat_id=chat_info_dict['chat_id'],
            chat_user_id=chat_info_dict['chat_user_id'],
            chat_user_type=chat_info_dict['chat_user_type'],
            knowledge_id_list=chat_info_dict['knowledge_id_list'],
            exclude_document_id_list=chat_info_dict['exclude_document_id_list'],
            application_id=chat_info_dict['application_id']
        )

        # Assign directly using dict unpacking
        c.chat_record_list = [
            ChatInfo.map_to_chat_record(c_r)
            for cr in chat_info_dict['chat_record_list']
        ]

        return c

    @staticmethod
    def get_cache(chat_id):
        chat_info_dict = cache.get(Cache_Version.CHAT.get_key(key=chat_id), version=Cache_Version.CHAT_INFO.get_version())

        if chat_info_dict:
            c = ChatInfo(*list(chat_info_dict.values()))
            c.chat_record_list = [
                ChatInfo(map_to_chat_record(cr))
                for cr in chat_info_dict['chat_record_list']
            ]
            return c

        return None
```

**Additional Suggestions**:
- Ensure that the variables (`cache`, `Model`) have been imported properly at the top of the file.
- The use of list comprehension ensures readability and efficiency.
- Consider handling exceptions where necessary, as they might affect the flow of your application.

These changes should make the code slightly more efficient and easier to maintain.

apps/application/flow/i_step_node.py
```diff
@@ -94,6 +94,7 @@ def handler(self, workflow):
                 application_public_access_client.access_num = application_public_access_client.access_num + 1
                 application_public_access_client.intraday_access_num = application_public_access_client.intraday_access_num + 1
                 application_public_access_client.save()
+        self.chat_info = None
 
 
 class NodeResult:
```
shaohuzhang1, line 100: The provided Python code snippet seems to be part of a logging system with classes `Handler` and `NodeResult`. However, there are several potential concerns and areas for improvement:

1. **Empty `self.chat_info` Assignment**: The last line of the `handler` method sets `self.chat_info` to `None`, but this assignment does not serve any purpose if it's followed by other operations that depend on having a value. Consider removing or commenting out the empty assignment before using `self.chat_info`.

2. **Potential Infinite Loop**: Assuming you have logic in `handler` that might cause an infinite loop due to certain conditions (untested), ensure such loops are appropriately handled.

3. **Database Transactions**: If database interactions are involved, consider enclosing them within transaction blocks to ensure atomicity, particularly if changes could fail halfway through execution.

4. **Unused Class**: If the `class NodeResult:` is only defined at the end of the file, remove it since it doesn't appear to contribute anything useful here.

Here’s a suggested revision:

```
def handler(self, workflow):
    while True: 
        application_public_access_client = ApplicationPublicAccessClient.objects.get(id=some_id)
        
        # Check if access limits are exceeded or need reset
        if check_limits(application_public_access_client): 
            application_public_access_client.access_num = application_public_access_client.access_num + 1
            application_public_access_client.intraday_access_num = application_public_access_client.intraday_access_num + 1
            application_public_access_client.save()

# Helper function to check access limits (implement based on business requirements)
def check_limits(instance):
    access_threshold = get_access_threshold()  
    return instance.intraday_access_num < access_threshold

# Optionally add a mechanism to restart/exit the flow when limits are reached
while running_loop:
    try:
        handle_next_task(workflow)  
        sleep(5)   # Example delay between checks
    except Exception as e:
```

PR #4365, opened 2025-11-17
fix: About read permission

shaohuzhang1:
fix: About read permission

### DIFF
--- a/ui/src/layout/layout-header/avatar/index.vue
+++ b/ui/src/layout/layout-header/avatar/index.vue
@@ -132,7 +132,17 @@
         </el-dropdown-item>
         <el-dropdown-item
           @click="openAbout"
-          v-if="hasPermission([RoleConst.ADMIN, PermissionConst.ABOUT_READ], 'OR')"
+          v-if="
+              hasPermission(
+                new ComplexPermission(
+                  [RoleConst.ADMIN, RoleConst.USER, RoleConst.WORKSPACE_MANAGE],
+                  [PermissionConst.ABOUT_READ],
+                  [],
+                  'OR',
+                ),
+                'OR',
+              )
+            "
         >
           {{ $t('layout.about.title') }}
         </el-dropdown-item>


ui/src/layout/layout-header/avatar/index.vue
```diff
@@ -132,7 +132,17 @@
...
+                  [PermissionConst.ABOUT_READ],
+                  [],
+                  'OR',
+                ),
+                'OR',
+              )
+            "
         >
           {{ $t('layout.about.title') }}
         </el-dropdown-item>
```
shaohuzhang1, line 148: Yes, there is an issue with this code snippet. The second `v-if` condition uses a variable that is not defined:

```javascript
hasPermission(new ComplexPermission([...], [...]), 'OR')
```

The `ComplexPermission` constructor takes four parameters: permissions for users/roles (`[...].concat([ RoleConst.WORKSPACE_MANAGE])`). However, inside the constructor call, you are only concatenating `[ RoleConst.WORKSPACE_MANAGE]`. This means that the permission array will be `[ RoleConst.WORKSPACE_MANAGE ]`, which does not include any roles or permissions.

If no role/permission exists in this array, calling `hasPermission(...)` may result in unexpected behavior.

To correct this, ensure that all required arguments are passed correctly to `new ComplexPermission(...)`: either both arrays (for users/roles and permissions) or one of them empty.

Here's how you can fix it:

**Incorrect:**
```javascript
new ComplexPermission([RoleConst.ADMIN, RoleConst.USER, RoleConst.WORKSPACE_MANAGE], [PermissionConst.ABOUT_READ])
```
**Corrected:**
```javascript
new ComplexPermission([
  ...[RoleConst ADMIN, RoleConst.USER]
  RoleConst.WORKSPACE_MANAGE,
],[
  PermissionConst.BLOG_DELETE,
  PermissionConst.ANNOUNCE_COMMENT_CREATE,
])
```

Ensure that every role or permission is added to at least one of those two arrays when using `ComplexPermission`.

Also, keep these considerations in mind when creating instances of complex permission classes to avoid runtime errors due to incorrect argument lengths.

PR #4368, opened 2025-11-18
fix: Reason content

shaohuzhang1:
fix: Reason content

### DIFF
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -416,10 +416,10 @@ class BaseChatStep(IChatStep):
             reasoning_result_end = reasoning.get_end_reasoning_content()
             content = reasoning_result.get('content') + reasoning_result_end.get('content')
             if 'reasoning_content' in chat_result.response_metadata:
-                reasoning_content = chat_result.response_metadata.get('reasoning_content', '')
+                reasoning_content = (chat_result.response_metadata.get('reasoning_content', '') or '')
             else:
-                reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get(
-                    'reasoning_content')
+                reasoning_content = (reasoning_result.get('reasoning_content') or "") + (reasoning_result_end.get(
+                    'reasoning_content') or "")
             post_response_handler.handler(chat_id, chat_record_id, paragraph_list, problem_text,
                                           content, manage, self, padding_problem_text,
                                           reasoning_content=reasoning_content)
--- a/apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
+++ b/apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
@@ -108,9 +108,9 @@ def write_context(node_variable: Dict, workflow_variable: Dict, node: INode, wor
     content = reasoning_result.get('content') + reasoning_result_end.get('content')
     meta = {**response.response_metadata, **response.additional_kwargs}
     if 'reasoning_content' in meta:
-        reasoning_content = meta.get('reasoning_content', '')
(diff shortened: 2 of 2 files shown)


apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
```diff
@@ -200,6 +200,14 @@ def execute(self, message_list: List[BaseMessage],
...
+        # 删除临时生成的MCP代码文件
+        if self.context.get('execute_ids'):
+            executor = ToolExecutor(CONFIG.get('SANDBOX'))
+            # 清理工具代码文件，延时删除，避免文件被占用
+            for tool_id in self.context.get('execute_ids'):
+                code_path = f'{executor.sandbox_path}/execute/{tool_id}.py'
+                if os.path.exists(code_path):
+                    os.remove(code_path)
         return {
             'step_type': 'chat_step',
```
shaohuzhang1, line 212: The provided Python function `execute_block` has a minor issue with string concatenation. Instead of using an OR operator (`or`) which could lead to unexpected results when handling empty strings, it should be combined more straightforwardly. Here's the revised version:

```python
def execute_block(self, message_list: List[BaseMessage],
                     chat_result):
    # Existing logic...
    
        reasoning_content = chat_result.response_metadata.get('reasoning_content', '')
    else:
        reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get(
            'reasoning_content')
        
    # Revised line
    output_string = ("{}{}{}".format(*[
        item or ""
        for item in [
            reasoning_result.get('content'),
            reasoning_result_end.get('content'),
            " " if any(item) else "",  # Add space only if either part is not empty
            reasoning_content
        ]
]))
    
                                 content, manage, self, padding_problem_text,
                                 reasoning_content=output_string)
```

In this revision, I used tuple unpacking within f-string formatting. This method ensures that all values are evaluated and concatenated correctly without unnecessary checks for emptiness. The spaces are added conditionally based on whether either the main content or reasoning content is present.

apps/application/flow/step_node/ai_chat_step_node/impl/base_chat_node.py
```diff
@@ -108,9 +108,9 @@ def write_context(node_variable: Dict, workflow_variable: Dict, node: INode, wor
     content = reasoning_result.get('content') + reasoning_result_end.get('content')
     meta = {**response.response_metadata, **response.additional_kwargs}
     if 'reasoning_content' in meta:
-        reasoning_content = meta.get('reasoning_content', '')
+        reasoning_content = (meta.get('reasoning_content', '') or '')
     else:
-        reasoning_content = reasoning_result.get('reasoning_content') + reasoning_result_end.get('reasoning_content')
+        reasoning_content = (reasoning_result.get('reasoning_content') or '') + (reasoning_result_end.get('reasoning_content') or '')
     _write_context(node_variable, workflow_variable, node, workflow, content, reasoning_content)
```
shaohuzhang1, line 116: The provided Python function is mostly correct but contains two minor issues:

1. The line `content = reasoning_result.get('content') + reasoning_result_end.get('content')` can be simplified using logical OR (`or`) to avoid multiple conditions.

2. Similarly, the line where both `reasoning_content` and its fallbacks are concatenated should also use `or`.

Here's an optimized version of the code with these corrections:
```python
def write_context(node_variable: Dict, workflow_variable: Dict, node: INode,
                 workflow: WorkflowData, response):
    # Sum reasoning content from responses or their empty strings
    reasoning_content = (response.reasoning_content or '') \
                      + (response.response_metadata.get('reasoning_content', '') or '')

    content = reasoninig_result.get('content') + reasoning_resultEnd.get('content')

    _write_context(node_variable, workflow_variable, node, workflow, content, reasoning_content)

```

These changes ensure that if either `response.reasoning_content` or the one specified in metadata is None, it will default to an empty string before concatenation. This avoids potential runtime errors when attempting non-string values.

PR #4372, opened 2025-11-20
feat: Knowledge write

shaohuzhang1:
feat: Knowledge write

### DIFF
--- a/apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
+++ b/apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
@@ -34,6 +34,7 @@ class ParagraphInstanceSerializer(serializers.Serializer):
 class KnowledgeWriteParamSerializer(serializers.Serializer):
     name = serializers.CharField(required=True, label=_('document name'), max_length=128, min_length=1,
                                  source=_('document name'))
+    meta = serializers.DictField(required=False)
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
```diff
@@ -34,6 +34,7 @@ class ParagraphInstanceSerializer(serializers.Serializer):
 class KnowledgeWriteParamSerializer(serializers.Serializer):
     name = serializers.CharField(required=True, label=_('document name'), max_length=128, min_length=1,
                                  source=_('document name'))
+    meta = serializers.DictField(required=False)
     paragraphs = ParagraphInstanceSerializer(required=False, many=True, allow_null=True)
```
shaohuzhang1, line 40: There are no significant irregularities in the provided `KnowledgeWriteParamSerializer` code. However, there is an issue with how the field names are being set.

In Django REST Framework (DRF) Serializer fields defined using the standard DRF attributes like `label`, `max_length`, and `min_length` should be used directly without setting the `source` attribute if you don't intend to change their default value during serialization. In this case, it seems all these properties are intended to match the Python variable names (`name`), so using them directly or omitting `source=_('document name')` would make more sense.

Additionally, consider adding validators or additional metadata handling if needed beyond basic type validation and constraints.

PR #4374, opened 2025-11-21
fix: Allow problem be blank

shaohuzhang1:
fix: Allow problem be blank

### DIFF
--- a/apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
+++ b/apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
@@ -26,7 +26,7 @@ class ParagraphInstanceSerializer(serializers.Serializer):
                                     allow_blank=True)
     title = serializers.CharField(required=False, max_length=256, label=_('section title'), allow_null=True,
                                   allow_blank=True)
-    problem_list = serializers.ListField(required=False, child=serializers.CharField(required=True))
+    problem_list = serializers.ListField(required=False, child=serializers.CharField(required=False, allow_blank=True))
     is_active = serializers.BooleanField(required=False, label=_('Is active'))
     chunks = serializers.ListField(required=False, child=serializers.CharField(required=True))
 
@@ -90,7 +90,7 @@ def get_paragraph_problem_model(knowledge_id: str, document_id: str, instance: D
 def get_paragraph_model(document_model, paragraph_list: List):
     knowledge_id = document_model.knowledge_id
     paragraph_model_dict_list = [
-        get_paragraph_problem_model(knowledge_id,document_model.id,paragraph)
+        get_paragraph_problem_model(knowledge_id, document_model.id, paragraph)
         for paragraph in paragraph_list
     ]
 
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
```diff
@@ -212,4 +209,4 @@ def get_details(self, index: int, **kwargs):
             'write_content': self.context.get("write_content"),
             'status': self.status,
             'err_message': self.err_message
-        }
\ No newline at end of file
+        }
```
shaohuzhang1, line 212: There seem to be a couple of potential issues in your code that could be addressed:

1. **Problem List Field**: The `problem_list` field should allow blank entries if it's supposed to represent multiple problems or tasks associated with a paragraph.

   ```python
   problem_list = serializers.ListField(
       required=False, 
       child=serializers.CharField(required=False, allow_blank=True)
   )
   ```

2. **Empty Return Statement in get_paragraph_problem_model Function**: There might be an unhandled case that results in no data being returned from `get_paragraph_problem_model`.

3. **Return Statements in Class Methods**:
    - Some methods may not have been executed correctly due to return statements that do not align with their expected use (e.g., within loops).

Here’s how you can address these points with corrections:

### Code Corrections

```python
def get_paragraph_problem_model(knowledge_id: str, document_id: str, instance: Dict):
    # Placeholder implementation; adjust based on actual requirements
    paragraphs = []
    for paragraph in instance['paragraphs']:
        print(f"Creating paragraph model for {knowledge_id}, {document_id}")
        paragraphs.append({
            "title": paragraph.get('title'),
            "problem_list": [
                task.strip() for task in paragraph.get('task', '').split(',') if task.strip()
            ],
            # Add additional fields as needed
        })
    
    return paragraphs

# Adjust other methods similarly if necessary

class ParagraphInstanceSerializer(serializers.Serializer):
    ...
    problem_list = serializers.ListField(
        required=False, 
        child=serializers.CharField(required=False, allow_blank=True)
    )

# Ensure all methods call super().__init__() and handle exceptions appropriately

def execute(self, documents, **kwargs) -> NodeResult:
    try:
        ...
        result_content = {
               'write_content': write_content_list,
               # Add more keys/values as appropriate
           }
        
        self.add_to_write_context(result_content)  # Assuming this method exists somewhere
        
        return NodeResult(result_content, {})  # This assumes NodeResult takes two parameters (success/data/error)

    except Exception as e:
        self.err_message(str(e))
        return NodeResult({}, {"error": str(e)})
```

### Additional Suggestion

1. **Exception Handling**: Implement robust exception handling throughout your code to catch unexpected errors and ensure proper cleanup when something goes wrong.

2. **Validation Logic**: Consider adding validation logic before saving changes to ensure that the input meets expectations.

3. **Code Structure**: Simplify complex operations in functions like `execute` or `save_by_doc_ids_if_needed` by breaking down these responsibilities into smaller, focused functions.

By addressing these areas, your code becomes more robust, maintainable, and likely performs better under real-world conditions.

PR #4380, opened 2025-11-24
feat: Chunks stored

shaohuzhang1:
feat: Chunks stored


apps/knowledge/migrations/0006_paragraph_chunks.py
```diff
@@ -0,0 +1,19 @@
...
+        ('knowledge', '0005_knowledgeaction'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='paragraph',
+            name='chunks',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(), default=list, size=None, verbose_name='块'),
+        ),
+    ]
```
shaohuzhang1, line 19: The code is generally correct but can be optimized in a few places:

1. **Default Value Type**: The `default` parameter should use an immutable type like `tuple()` instead of `list()`. This ensures that changes to the list don't affect other instances.

   ```python
   field=django.contrib.postgres.fields.ArrayField(
       base_field=models.CharField(),
       default=tuple,
       size=None,
       verbose_name='块',
   ),
   ```

2. **Verbose Name Capitalization**: Django's convention is to capitalize the first letter when using verbose names. So, it should be capitalized as " chunks" (with a space).

   ```python
   verbose_name='块',
   ```

3. **Docstring for Models**: It would be beneficial to add docstrings to the `BaseModel` or `models.Model` class where necessary, and perhaps include comments explaining specific configurations if they are not obvious from the code itself.

Here's the revised version with these optimizations:

```python
@@ -0,0 +1,26 @@
+# Generated by Django 5.2.8 on 2025-11-24 07:09
+
+import django.contrib.postgres.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('knowledge', '0005_knowledgeaction'),  #: Dependency tuple containing migration information, such as app label, operation, etc.
+                                          # Usage example: apps.models.OperationDependency(app='appname', operation='model_operation')
+                                      ]
+
+    operations = [
+        migrations.AddColumn (
+            model_name='paragraph',           #: Specifies the Django model being modified. For each addition or deletion of fields, you must specify this unique table name.
+                                          # Usage example: models.AddColumn(model=models.Post, column='body_text', type=db.TextField)
+                                  ),          
+                                           
+        	migrations.AddField(
+            model_name='paragraph',
+            name='chunks',
+            #: Specifies the name of the new or existing attribute within your model class. In case of altering an existing one, ensure that its data types match correctly.
+            type=django.contrib.postgresql.fields.ArrayField(base_field=models.CharField()), 
+                                       # Defines the type of value stored in the database as well as additional attributes that control how arrays behave.
+                               			default=tuple,                   # Sets the defalut value for this attribute.
+                                   									size=None              	# Optional argument specifying maximum number of elements array may contain at any time; sets to None (unlimited).
+                                                                                        	 						 	         			         	
+                                    										verbose_name='内容',    					# Provides human readable representation of the field name used primarily during serialization.
+                                                                                                                                                                          			 										 
+                                   ),
+                                            							                                      
+    
```
Make sure all dependencies are properly formatted according to the documentation provided and replace placeholder values with actual details relevant to your application.

apps/knowledge/vector/base_vector.py
```diff
@@ -104,7 +105,8 @@ def batch_save(self, data_list: List[Dict], embedding: Embeddings, is_the_task_i
                 break
 
     @abstractmethod
-    def _save(self, text, source_type: SourceType, knowledge_id: str, document_id: str, paragraph_id: str, source_id: str,
+    def _save(self, text, source_type: SourceType, knowledge_id: str, document_id: str, paragraph_id: str,
+              source_id: str,
               is_active: bool,
               embedding: Embeddings):
         pass
```
shaohuzhang1, line 112: The code looks generally well-organized and follows best practices. However, there are a few minor improvements that can make it more robust and maintainable:

1. **Error Handling for Input Types**: The `SourceType.PARAGRAPH.value` comparison should be done using an equality operator (`==`) instead of strict type conversion with `str()`, which may cause errors.

2. **Optimization in `_save` Method**: The abstract method signature should not specify all parameters explicitly, as this is typically handled by the subclass implementation. This allows subclasses to have flexibility in their parameter lists.

Here's the slightly improved version of the code:

```python
@@ -63,7 +63,8 @@ def save_pre_handler(self):
                 BaseVectorStore.vector_exists = True
         return True

-    def save(self, text, source_type: SourceType, knowledge_id: str, document_id: str,
+    def save(self, text, source_type: SourceType, knowledge_id: str, document_id: str,
              paragraph_id: str, source_id: str,
              is_active: bool,
              embedding: Embeddings) -> bool:

-        if str(source_type) != SourceType.PARAGRAPH.value:
+        if source_type != SourceType.PARAGRAPH:
             return False

         chunks = data.get('chunks', [])
         new_chunks_list = []
```

### Summary of Improvements:
- Used `==` for comparing string representations to ensure correct logic.
- Removed explicit specification of all parameters from the abstract `_save` method signature, as this will be implemented by the subclass.

apps/application/flow/step_node/knowledge_write_node/impl/base_knowledge_write_node.py
```diff
@@ -67,14 +68,14 @@ def link_file(source_file_id, document_id):
...
         id=uuid.uuid7(),
         document_id=document_id,
         content=instance.get("content"),
         knowledge_id=knowledge_id,
-        title=instance.get("title") if 'title' in instance else ''
+        title=instance.get("title") if 'title' in instance else '',
+        chunks = instance.get('chunks') if 'chunks' in instance else text_to_chunk(instance.get("content")),
     )
 
     problem_paragraph_object_list = [ProblemParagraphObject(
```
shaohuzhang1, line 81: There are a few potential issues and optimizations in the provided code:

1. **Importing `text_to_chunk`**: The call to `text_to_chunk(instance.get("content"))` is at the wrong place within the `get_paragraph_problem_model` function. It should be assigned directly to the `paragraph.chunks` attribute instead of inside the dictionary before returning or storing it.

2. **Handling Empty Titles**: While it's generally good practice to have default values for optional fields like titles, there might be cases where an empty title could cause problems (e.g., validation errors). Adding more error handling around `instance.get("title")` can make the code more robust.

3. **Unnecessary Code Block**: Remove any unnecessary indentation in the `if 'title' in instance:` block since there are no further statements that will execute under this condition.

Here is a revised version of the code with these improvements:

```python
@transaction.atomic
def update_knowledge(knowledge_data):
    """
    更新知识库，同时添加或修改问题章节、段落和相关文件
    :param knowledge_data:
    :return:
    """
    
    # 解析传入的知识内容
    document_title = knowledge_data.get('document_title')
    paragraphs_data = knowledge_data.get('paragraphs', [])
    files_data = knowledge_data.get('files', [])

    # 创建文档
    document = Document.objects.update_or_create(
        name=document_title,
        # 其他必要字段
    )

    updated_doc_id = document.id

    # 添加文章问题（并批量写入数据库）
    paragraph_queryset = []

    for para_data in paragraphs_data:
        paragraph_dict_ = dict(document=document_id)
        
        if 'title' in para_data:
            paragraph_dict_.update(title=para_data['title'])
        
        if 'chunks' in para_data:
            paragraph_dict_.update(chunks=text_to_chunk(para_data['content']))

        paragraph = Paragraph(**paragraph_dict_)
        paragraph_queryset.append(paragraph)

    # 批量插入句子段落到数据库
    bulk_insert_result = bulk_create_in_batches(ProblemParagraphObject.objects.all().filter(knowledge__id=updated_doc_id))
    print(f"批量创建句段结果: {bulk_insert_result}")

# 使用上述函数示例调用：
# data = {
#     "document_title": "新测试文档",
#     "paragraphs": [
#         {"content": "段落1内容", "title": "段落标题1"},
#         {"content": "段落2内容"}
#     ],
#     "files": ["file1.pdf"]
# }
# update_knowledge(data)
```

By making these changes, you ensure proper handling of the `chunks` field when creating paragraph objects and improve the structure of your code by removing extraneous nesting levels.

PR #4382, opened 2025-11-25
feat: Filter user by role

shaohuzhang1:
feat: Filter user by role

### DIFF
--- a/apps/system_manage/serializers/user_resource_permission.py
+++ b/apps/system_manage/serializers/user_resource_permission.py
@@ -9,6 +9,7 @@
 import json
 import os
 
+from django.contrib.postgres.fields import ArrayField
 from django.core.cache import cache
 from django.db import models
 from django.db.models import QuerySet, Q, TextField
@@ -343,10 +344,13 @@ class ResourceUserPermissionSerializer(serializers.Serializer):
             "role": models.CharField(),
             "role_setting.type": models.CharField(),
             "user_role_relation.workspace_id": models.CharField(),
+            'tmp.type_list': ArrayField(models.CharField()),
+            'tmp.role_name_list_str': models.CharField()
 
         }))
         nick_name = instance.get('nick_name')
         username = instance.get('username')
+        role_name = instance.get('role')
         permission = instance.get('permission')
         query_p_list = [None if p == "NOT_AUTH" else p for p in permission]
 
@@ -375,15 +379,31 @@ class ResourceUserPermissionSerializer(serializers.Serializer):
                 **{"u.id__in": QuerySet(workspace_user_role_mapping_model).filter(
                     workspace_id=self.data.get('workspace_id')).values("user_id")})
         if is_x_pack_ee:
-            user_query_set = user_query_set.filter(
-                **{'role_setting.type': "USER", 'user_role_relation.workspace_id': self.data.get('workspace_id')})
+            user_query_set = user_query_set.filter(**{
+                "tmp.type_list__contains": ["USER"]
+            })
+            role_name_and_type_query_set = QuerySet(model=get_dynamics_model({
+            'user_role_relation.workspace_id': models.CharField(),
+        })).filter(**{
+                "user_role_relation.workspace_id": self.data.get('workspace_id'),
+            })
+            if role_name:
+                user_query_set = user_query_set.filter(
+                    **{'tmp.role_name_list_str__icontains': str(role_name)}
+                )
+
+            return {
+                'workspace_user_resource_permission_query_set': workspace_user_resource_permission_query_set,
+                'user_query_set': user_query_set,
+                'role_name_and_type_query_set': role_name_and_type_query_set
+            }
         else:
             user_query_set = user_query_set.filter(
                 **{'role': "USER"})
-        return {
-            'workspace_user_resource_permission_query_set': workspace_user_resource_permission_query_set,
-            'user_query_set': user_query_set
-        }
+            return {
+                'workspace_user_resource_permission_query_set': workspace_user_resource_permission_query_set,
+                'user_query_set': user_query_set
+            }
 
     def list(self, instance, with_valid=True):
         if with_valid:
--- a/apps/system_manage/views/user_resource_permission.py
+++ b/apps/system_manage/views/user_resource_permission.py
@@ -196,6 +196,7 @@ class WorkspaceResourceUserPermissionView(APIView):
             return result.success(ResourceUserPermissionSerializer(
                 data={'workspace_id': workspace_id, "target": target, 'auth_target_type': resource, }
             ).page({'username': request.query_params.get("username"),
+                    'role': request.query_params.get("role"),
                     'nick_name': request.query_params.get("nick_name"),
                     'permission': request.query_params.getlist("permission[]")}, current_page, page_size,
                    ))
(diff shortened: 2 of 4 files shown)


apps/system_manage/serializers/user_resource_permission.py
```diff
@@ -375,15 +379,31 @@ def get_queryset(self, instance, is_x_pack_ee: bool):
...
-            'workspace_user_resource_permission_query_set': workspace_user_resource_permission_query_set,
-            'user_query_set': user_query_set
-        }
+            return {
+                'workspace_user_resource_permission_query_set': workspace_user_resource_permission_query_set,
+                'user_query_set': user_query_set
+            }
 
     def list(self, instance, with_valid=True):
         if with_valid:
```
shaohuzhang1, line 409: The provided code has the following issues and optimizations that can be addressed:

Issue 1: Unnecessary field inclusion in queryset
The line `**{"u.id__in": QuerySet(workspace_user_role_mapping_model).filter(...)}`
can be simplified to just use `workspace_user_role_mapping_model.objects.filter(...)`, because Django's ORM will automatically include the fields specified by the manager.

```py
user_query_set = user_query_set.filter(
    user_role_relation__workspace_id=self.data.get('workspace_id'))
```

Note: Make sure you have an appropriate foreign key relationship defined between `workspace_user_role_mapping` and another model where `user_id` is stored.

Optimization 2: Simplified queryset filters
Some of the nested list comprehensions can be removed by directly filtering using Django's ORM methods. For example:

```py
if role_name:
    role_name_filter = Q(user_role_relation__role=role_name)
else:
    role_name_filter = None

user_query_set = user_query_set \
    .filter(workspace_user_role_relation__workspace_id=self.data.get('workspace_id')) \
    .apply_if_condition(lambda q, f: q.filter(f))
```

Or simply:

```py
user_query_set = user_query_set \
    .filter(
        workspace_user_role_relation__workspace_id=self.data.get('workspace_id'),
        (Q(user_role_relation__role=role_name) | Q(tmp__type_list=["USER"]))
    )
```

Also consider whether it makes sense for all cases to filter on `Q(tmp__type_list=["USER"])`. You might need different logic depending on your specific requirements.

Other minor improvements:
- Use more consistent naming convention for variables like `instance`.
- Consider renaming functions if they're not clear what they do, e.g., `get_queryset` could be renamed something related to accessing permissions.
- Review error handling for edge cases like when certain keys are missing or invalid data types.

This should help clean up the code structure while maintaining its functionality.

ui/src/components/resource-authorization-drawer/index.vue
```diff
@@ -338,11 +333,12 @@ const searchType = ref('nick_name')
...
+  role: '',
   permission: undefined,
 })
 
 const search_type_change = () => {
-  searchForm.value = { nick_name: '', username: '', permission: undefined }
+  searchForm.value = { nick_name: '', username: '', role: '', permission: undefined }
 }
 
 const paginationConfig = reactive({
```
shaohuzhang1, line 344: The code looks generally clean and well-structured. However, there are a few issues to consider:

### 1. Conditional Rendering for `Role` Column
There is an unnecessary conditional rendering block in the table column:
```html
<!-- <el-table-column prop="role_name" :label="$t('views.role.member.role')" width="210"> -->
<el-table-column v-if="hasPermission([EditionConst.IS_EE,EditionConst.IS_PE],'OR')" prop="role_name" :label="$t('views.role.member.role')" width="210">
```
This will render a blank column if `hasPermission([EditionConst.IS_EE, EditionConst.IS_PE], "OR")` returns `false`.

### 2. Missing Import Statements
Ensure that you have imported all necessary constants at the top of your file:
```javascript
import { VersionConst, PermissionsTypesEnum } from './path/to/constants';
// ...
```

### 3. Code Duplication
In the `<el-radio-group>` template, there might be duplication or redundancy. You can simplify it by using computed properties.

#### Example Simplification:
Assuming some logic needs different handling based on permissions, you could refactor the radio group content into separate functions for better readability:
```html
<template>
  <div>
    <el-radio-group v-model="selectedOption">
      <el-radio label="option1">{{ $t("option1Label") }}</el-radio>
      <el-radio label="option2">{{ $t("option2Label") }}</el-radio>
      <!-- Add more options as needed -->
    </el-radio-group>

    <!-- Handle specific actions based on selected option -->
    <h2>{{ handleAction(selectedOption) }}</h2>
  </div>
</template>

<script>
export default {
  data() {
    return {
      selectedOption: 'default',
    };
  },
  methods: {
    handleAction(option) {
      switch (option) {
        case 'option1':
          return this.someOperationForOption();
        case 'option2':
          return this.otherOperationForOption();
        default:
          return this.defaultOperation();
      }
    },

    someOperationForOption() {
      // Logic for when option1 is selected
    },

    otherOperationForOption() {
      // Logic for when option2 is selected
    },

    defaultOperation() {
      // Default operation logic
    },
  },
};
</script>
```

### Optimization Suggestions
- **Use Refs**: Ensure you are reusing refs properly instead of creating new ones inside loops or conditions. This helps maintain state consistency across components.
  
- **Refactor Repeated Code**: If there's common functionality, extract it into reusable methods or functions to avoid repetition.

By addressing these points, the application should perform more smoothly and effectively while keeping its interface clean and efficient.

apps/system_manage/views/user_resource_permission.py
```diff
@@ -196,6 +196,7 @@ def get(self, request: Request, workspace_id: str, target: str, resource: str, c
             return result.success(ResourceUserPermissionSerializer(
                 data={'workspace_id': workspace_id, "target": target, 'auth_target_type': resource, }
             ).page({'username': request.query_params.get("username"),
+                    'role': request.query_params.get("role"),
                     'nick_name': request.query_params.get("nick_name"),
                     'permission': request.query_params.getlist("permission[]")}, current_page, page_size,
                    ))
```
shaohuzhang1, line 202: The provided code snippet seems to be from a Django REST framework view that handles GET requests for retrieving resource user permissions. There are a few issues and improvements that can be made:

1. **Parameter Type Mismatch**: The `request.query_params` method is intended to return strings (and lists of strings). However, you're passing these values directly into the serializer without checking their types. This could potentially lead to errors if one parameter isn't a string.

   ```python
   ResourceUserPermissionSerializer(
       data={
           'workspace_id': workspace_id,
           "target": target,
           'auth_target_type': resource,
           'username':
               request.query_params.get("username"),  # Should be str or None
           'role': request.query_params.get("role"),    # Should be str orNone
           'nick_name':
               request.query_params.get("nick_name"),  # Should be str or None
           'permission': [
                item.strip() for item in request.query_params.getlist("permission[]")],  # Ensure items are stripped and not just split
       }
   )
   ```

2. **Whitespace Errors**: The spaces at the beginning of lines like `'resource', "target": target, 'auth_target_type'` are unnecessary and could cause indentation mismatches in Python versions where strict indentation matters.

3. **String Handling**: When working with query parameters, it's good practice to strip whitespace around them to avoid unexpected results when handling inputs like `'role= Admin'`.

4. **Use Dict Comprehension**: Using `dict comprehension` instead of list operations (`[`, `,]`) on query parameters might make the code cleaner and more readable, especially when dealing with multiple keys.

Here’s the refined version of the function:

```python
def get(self, request: Request, workspace_id: str, target: str, resource: str, current_page=None, page_size=None):
    filter_conditions = {
        "username": request.query_params.get("username").strip(),
        "role": request.query_params.get("role").strip(),
        "nick_name": request.query_params.get("nick_name").strip(),
        'permission': [item.strip() for item in request.query_params.getlist("permission[]")]  
    }

    return result.success(ResourceUserPermissionSerializer(
        queryset=self.filter_queryset(Resource.objects.filter(**filter_conditions)),
        context={"current_page": current_page, "page_size": page_size},
        many=True
)).data
```

### Key Changes:
- **Dict Comprehension**: Used dictionary comprehension to build the `filter_conditions`.
- **Stripping Whitespace**: Stripped any leading/trailing whitespace from each condition before adding it to the filters.
- **Simplified Serializer Call**: Called `many=True` explicitly in the serializer call.

These changes should ensure that your API endpoint handles query parameters correctly and consistently.

PR #4383, opened 2025-11-25
feat: Alter fork depth

shaohuzhang1:
feat: Alter fork depth

### DIFF
--- a/apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
+++ b/apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
@@ -61,7 +61,7 @@ class BaseDataSourceWebNode(IDataSourceWebNode):
         collect_handler, document_list = get_collect_handler()
 
         try:
-            ForkManage(source_url,selector.split(" ") if selector is not None else []).fork(1,set(),collect_handler)
+            ForkManage(source_url,selector.split(" ") if selector is not None else []).fork(3,set(),collect_handler)
 
             return  NodeResult({'document_list': document_list},
                                self.workflow_manage.params.get('knowledge_base') or {})


apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
```diff
@@ -61,7 +61,7 @@ def execute(self, **kwargs) -> NodeResult:
         collect_handler, document_list = get_collect_handler()
 
         try:
-            ForkManage(source_url,selector.split(" ") if selector is not None else []).fork(1,set(),collect_handler)
+            ForkManage(source_url,selector.split(" ") if selector is not None else []).fork(3,set(),collect_handler)
 
             return  NodeResult({'document_list': document_list},
                                self.workflow_manage.params.get('knowledge_base') or {})
```
shaohuzhang1, line 67: The code provided makes several improvements:

1. **Increased Depth Parameter**: The `ForkManage` method's first argument, which specifies the depth of traversal, was changed from 1 to 3 (from `"source_url"`). This might be necessary if you intend to perform deeper crawling.

2. **Consolidated Set for Collectors**: Instead of explicitly passing an empty set as one of the arguments, it can be omitted since sets without elements are considered valid and do not affect performance significantly with Python sets.

Here's the updated version with these changes:

```python
@@ -61,8 +61,7 @@ def execute(self, **kwargs) -> NodeResult:
     collect_handler, document_list = get_collect_handler()

     try:
-        ForkManage(source_url, selector.split(" ") if selector is not None else [], collect_handler=collect_handler, max_depth=3).
-        fork()
+        ForkManage(source_url, selector.split(" ") if selector is not None else [], collect_handler=collect_handler, max_depth=3).fork()

     return NodeResult({'document_list': document_list},
                     self.workflow_manage.params.get('knowledge_base') or {})
```

**Changes Made**:
- Changed the `max_depth` parameter to 3.
- Omitted the unnecessary explicit setting of a third default argument (`set()`), which does not impact functionality.

PR #4400, opened 2025-11-28
fix: build type-check error

shaohuzhang1:
fix: build type-check error

### DIFF
--- a/ui/src/components/workflow-dropdown-menu/index.vue
+++ b/ui/src/components/workflow-dropdown-menu/index.vue
@@ -6,7 +6,7 @@ import { inject } from 'vue'
 import { WorkflowMode } from '@/enums/application'
 import ApplicationDropdownMenu from '@/components/workflow-dropdown-menu/application/index.vue'
 import KnowledgeDropdownMenu from '@/components/workflow-dropdown-menu/knowledge/index.vue'
-const workflow_mode = inject('workflowMode') || WorkflowMode.Application
+const workflow_mode: WorkflowMode = inject('workflowMode') || WorkflowMode.Application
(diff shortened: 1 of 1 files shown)


ui/src/components/workflow-dropdown-menu/index.vue
```diff
@@ -6,7 +6,7 @@ import { inject } from 'vue'
 import { WorkflowMode } from '@/enums/application'
 import ApplicationDropdownMenu from '@/components/workflow-dropdown-menu/application/index.vue'
 import KnowledgeDropdownMenu from '@/components/workflow-dropdown-menu/knowledge/index.vue'
-const workflow_mode = inject('workflowMode') || WorkflowMode.Application
+const workflow_mode: WorkflowMode = inject('workflowMode') || WorkflowMode.Application
 defineProps({
   show: {
     type: Boolean,
```
shaohuzhang1, line 12: - **Line 6**: You can remove the `|| WorkflowMode.Application` as TypeScript will infer the type of `inject('workflowMode')`.
- **Line 7**: Ensure that `$v-model.show`, `$v-model.workflow_mode`, and other reactive properties used in templates have valid bindings to ensure Vue's reactivity system works correctly.
- **Line 9**: Since you are using `defineProps` with types, the line should be updated accordingly based on what values your component expects. For example:

```javascript
defineProps<{
  mode?: WorkflowMode;
  show?: boolean; // Optional property if desired
}>();
```

This is assuming that `mode` might also be provided through props.

Overall, focusing on proper TypeScript inference and ensuring correct data binding within the component template would generally suffice for this piece of code.

PR #4411, opened 2025-12-02
fix: After modifying the dialogue prefix, the embedded page cannot have a conversation

shaohuzhang1:
fix: After modifying the dialogue prefix, the embedded page cannot have a conversation

### DIFF
--- a/apps/chat/serializers/chat_embed_serializers.py
+++ b/apps/chat/serializers/chat_embed_serializers.py
@@ -75,6 +75,7 @@ class ChatEmbedSerializer(serializers.Serializer):
                  'white_active': 'true' if application_access_token.white_active else 'false',
                  'is_draggable': is_draggable,
                  'float_icon': float_icon,
+                 'prefix': CONFIG.get_chat_path(),
                  'query': query,
                  'show_guide': show_guide,
                  'x_type': float_location.get('x', {}).get('type', 'right'),
--- a/apps/chat/template/embed.js
+++ b/apps/chat/template/embed.js
@@ -26,9 +26,9 @@ const chatButtonHtml=
 
 
 
-const getChatContainerHtml=(protocol,host,token,query)=>{
+const getChatContainerHtml=(protocol,host,token,query,prefix)=>{
  return `<div id="maxkb-chat-container">
-<iframe id="maxkb-chat" allow="microphone" src=${protocol}://${host}/chat/${token}?mode=embed${query}></iframe>
+<iframe id="maxkb-chat" allow="microphone" src=${protocol}://${host}${prefix}/${token}?mode=embed${query}></iframe>
 <div class="maxkb-operate"><div class="maxkb-closeviewport maxkb-viewportnone"><svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20" fill="none">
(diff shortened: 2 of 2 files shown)


apps/chat/template/embed.js
```diff
@@ -62,7 +62,7 @@ const initChat=(root)=>{
   // 添加对话icon
   root.insertAdjacentHTML("beforeend",chatButtonHtml)
   // 添加对话框
-  root.insertAdjacentHTML('beforeend',getChatContainerHtml('{{protocol}}','{{host}}','{{token}}','{{query}}'))
+  root.insertAdjacentHTML('beforeend',getChatContainerHtml('{{protocol}}','{{host}}','{{token}}','{{query}}','{{prefix}}'))
   // 按钮元素
   const chat_button=root.querySelector('.maxkb-chat-button')
   const chat_button_img=root.querySelector('.maxkb-chat-button > img')
```
shaohuzhang1, line 68: The provided code has a few suggestions and considerations:

1. **Prefix Parameter**: The code now includes an optional `prefix` parameter to handle different paths within the domain, which is good approach.

2. **Template Literals**: It's recommended to use template literals for string interpolation where possible, but you're already using them correctly.

3. **Dynamic Content Handling**: Ensure that the dynamic content (like protocol, host, token, query) provided by your application doesn't contain sensitive information like tokens or passwords directly in the HTML string. Consider passing these through JavaScript variables instead of hardcoding them.

4. **Function Calls with Template Strings**: When calling functions with multiple parameters using template strings, ensure consistency in spacing around commas.

Here’s a revised version of the code with comments and minor improvements:

```javascript
const chatButtonHtml = `
<div id="maxkb-chat-container">
  <iframe id="maxkb-chat" allow="microphone"
          src="${protocol}:${host}${prefix}/chat/${token}?mode=embed${encodeURIComponent(query)}"></iframe>
  <div class="maxkb-operate">
    <div class="maxkb-closeviewport maxkb-viewportnone">
      <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20" fill="none">
        <!-- SVG path details -->
      </svg>
    </div>
  </div>
</div>
`;

function initChat(root) {
  // Add chat button
  root.insertAdjacentHTML("beforeend", chatButtonHtml);

  // Add chat container
  const fullPath = `${protocol}:${host}${prefix}`;
  root.insertAdjacentHTML(
    'beforeend',
    getChatContainerHtml(fullPath, token, { query }, prefix)
  );

  // Button elements
  const chat_button = root.querySelector('.maxkb-chat-button');
  const chat_button_img = root.querySelector('.maxkb-chat-button > img');
}

// Function signature should match expectations
function getChatContainerHtml(protocol, host, params, prefix) {
  return `
  <div id="maxkb-chat-container">
    <iframe id="maxkb-chat" allow="microphone"
            src="${protocol}:${host}${prefix}/chat/${params.token}?mode=embed&${new URLSearchParams(params.query).toString()}"></iframe>
    <div class="maxkb-operate">
      <div class="maxkb-closeviewport maxkb-viewportnone">
        <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20" fill="none">
          <!-- SVG path details -->
        </svg>
      </div>
    </div>
  </div>`;
}
```

### Key Points:
- Dynamically build the iframe source using template literals.
- Use `URLSearchParams` to encode the query parameters without adding spaces around commas.
- Pass `fullDomain`, `token`, `{ query }`, and `prefix` as separate arguments to avoid hardcoding unnecessary values in templates.

apps/chat/serializers/chat_embed_serializers.py
```diff
@@ -75,6 +75,7 @@ def get_embed(self, with_valid=True, params=None):
                  'white_active': 'true' if application_access_token.white_active else 'false',
                  'is_draggable': is_draggable,
                  'float_icon': float_icon,
+                 'prefix': CONFIG.get_chat_path(),
                  'query': query,
                  'show_guide': show_guide,
                  'x_type': float_location.get('x', {}).get('type', 'right'),
```
shaohuzhang1, line 81: The provided code snippet includes an addition to the `get_embed` method in the `EmbedBuilder` class. Specifically, it adds a new parameter `prefix` which assigns the result of `CONFIG.get_chat_path()` to this new key `'prefix'`. Additionally, there's no change made to existing parameters.

### Potential Issues:
1. **Unused Parameter**: The variable `with_valid` is declared but not used within the method.
   ```python
   'white_active': 'true' if application_access_token.white_active else 'false'
   ```

2. **Lack of Type Specification**: It’s good practice to explicitly specify types of function arguments and return values for better readability and maintainability.

3. **Uninitialized Variable Reference** (`application_access_token.white_active`) - Ensure `application_access_token` has been previously defined or properly initialized before use.

4. **Magic Number for 'right'**:
   ```python
   x_type = location.get('x', {}).get('type', 'right')
   ```
   Instead of using hardcoded strings (like "'right'" without quotes), consider defining constants for these kinds of configurations:

```python
RIGHT_ALIGN = "right"
LEFT_ALIGN = "left"
```
Then update your logic accordingly:

```python
x_type = location.get("x", {}).get("type", RIGHT_ALIGN)  # Assuming right is the default alignment
```

### Suggested Optimizations / Improvements:
- Remove unused imports/declarations that might not be needed elsewhere.
- Add type annotations to functions parameters & return values, e.g.,
  
  ```python
  def get_embed(
      self,
      with_valid: bool = True,
      params: Optional[dict] = None
  ) -> dict:
      ...
      return {
          ...
      }
  ```

- Use more descriptive names instead of single-character placeholders like `"q"` for `'query'`.

Overall, the code seems mostly clean and functional, but incorporating the above mentioned improvements would make it both clearer and robust.

PR #4422, opened 2025-12-03
feat: Locales

shaohuzhang1:
feat: Locales

### DIFF
--- a/apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
+++ b/apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
@@ -19,8 +19,10 @@ from common.utils.logger import maxkb_logger
 
 
 class BaseDataSourceWebNodeForm(BaseForm):
-    source_url = forms.TextInputField(_('Web source url'), required=True)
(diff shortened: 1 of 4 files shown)


apps/application/flow/step_node/data_source_web_node/impl/base_data_source_web_node.py
```diff
@@ -61,24 +63,22 @@ def execute(self, **kwargs) -> NodeResult:
...
         return {
             'name': self.node.properties.get('stepName'),
             "index": index,
             'run_time': self.context.get('run_time'),
             'type': self.node.type,
-            'input_params': {"source_url": self.context.get("source_url"),"selector": self.context.get('selector')},
+            'input_params': {"source_url": self.context.get("source_url"), "selector": self.context.get('selector')},
             'output_params': self.context.get('document_list'),
             'knowledge_base': self.workflow_params.get('knowledge_base'),
             'status': self.status,
```
shaohuzhang1, line 84: No immediate changes needed for this file. The code looks clean, well-indented, and follows Python style guidelines.

However, one minor suggestion is to avoid using single-character variable names like `handler` and `result`. Using meaningful names makes the code more readable and maintainable. For example:

```python
def get_data_collect_handler():
```

And

```python
ForkManage(source_url, selector.split(" ") if selector is not None else []).fork(3, set(), data_collect_handler)
```

This change will improve clarity in the code.

PR #4423, opened 2025-12-03
fix: Knowledge write node height

shaohuzhang1:
fix: Knowledge write node height

### DIFF
--- a/ui/src/workflow/common/data.ts
+++ b/ui/src/workflow/common/data.ts
@@ -130,8 +130,8 @@ export const knowledgeWriteNode = {
   type: WorkflowType.KnowledgeWriteNode,
   text: t('views.workflow.nodes.knowledgeWriteNode.text'),
   label: t('views.workflow.nodes.knowledgeWriteNode.label'),
+  height: 252,
   properties: {
-    height: 252,
     stepName: t('views.workflow.nodes.knowledgeWriteNode.label'),
     config: {
       fields: [],


ui/src/workflow/common/data.ts
```diff
@@ -130,8 +130,8 @@ export const knowledgeWriteNode = {
   type: WorkflowType.KnowledgeWriteNode,
   text: t('views.workflow.nodes.knowledgeWriteNode.text'),
   label: t('views.workflow.nodes.knowledgeWriteNode.label'),
+  height: 252,
   properties: {
-    height: 252,
     stepName: t('views.workflow.nodes.knowledgeWriteNode.label'),
     config: {
       fields: [],
```
shaohuzhang1, line 137: The code appears to be a function definition within an object called `knowledgeWriteNode`. There is a slight issue with the `height` property being set twice, which is unnecessary and potentially redundant.

Here's a revised version of the code:

```javascript
export const knowledgeWriteNode = {
  type: WorkflowType.KnowledgeWriteNode,
  text: t('views.workflow.nodes.knowledgeWriteNode.text'),
  label: t('views.workflow.nodes.knowledgeWriteNode.label'),
+  height: 252,
  properties: {
    // Remove the duplicate 'height' field here if it's not intentionally repeated elsewhere in your codebase
    stepName: t('views.workflow.nodes.knowledgeWriteNode.label'),
    config: {
      fields: [],
    },
};
```

This change ensures that each key-value pair in the `properties` object is unique and reduces redundancy. If you have other instances where the `height` might also appear incorrectly in the same context, those should also be addressed accordingly.

PR #4424, opened 2025-12-03
fix: Node height

shaohuzhang1:
fix: Node height

### DIFF
--- a/ui/src/workflow/common/data.ts
+++ b/ui/src/workflow/common/data.ts
@@ -130,7 +130,7 @@ export const knowledgeWriteNode = {
   type: WorkflowType.KnowledgeWriteNode,
   text: t('views.workflow.nodes.knowledgeWriteNode.text'),
   label: t('views.workflow.nodes.knowledgeWriteNode.label'),
-  height: 252,
+  height: 100,
   properties: {
     stepName: t('views.workflow.nodes.knowledgeWriteNode.label'),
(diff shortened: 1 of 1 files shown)


ui/src/workflow/common/data.ts
```diff
@@ -130,7 +130,7 @@ export const knowledgeWriteNode = {
   type: WorkflowType.KnowledgeWriteNode,
   text: t('views.workflow.nodes.knowledgeWriteNode.text'),
   label: t('views.workflow.nodes.knowledgeWriteNode.label'),
-  height: 252,
+  height: 100,
   properties: {
     stepName: t('views.workflow.nodes.knowledgeWriteNode.label'),
     config: {
```
shaohuzhang1, line 136: The code you provided has two minor changes:

1. The `height` property of the `knowledgeWriteNode` is reduced from 252 to 100 pixels.
2. The value for the `widthPropertyKey` in the `config` section was changed to `'stepName'`.

These changes may impact the visual layout of the workflow node on some interfaces. If these changes were intended, they align with reducing the size and focusing on the step name. However, without additional context about the specific requirements or constraints of your workflow application, it's difficult to determine whether either change is optimal. 

If you have more details or if there are particular areas of concern (e.g., responsiveness, compatibility across different devices), feel free to share them so I can provide a more tailored response.

PR #4425, opened 2025-12-03
fix: Workflow save before publish

shaohuzhang1:
fix: Node height<br>fix: Workflow save before publish

### DIFF
--- a/ui/src/views/knowledge-workflow/index.vue
+++ b/ui/src/views/knowledge-workflow/index.vue
@@ -315,7 +315,7 @@ const publish = () => {
         return
       }
       loadSharedApi({ type: 'knowledge', systemType: apiType.value })
-        .putKnowledge(id, { work_flow: workflow })
+        .putKnowledgeWorkflow(id, { work_flow: workflow })
         .then(() => {
           return loadSharedApi({ type: 'knowledge', systemType: apiType.value }).publish(
             id,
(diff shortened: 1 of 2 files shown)


ui/src/views/knowledge-workflow/index.vue
```diff
@@ -315,7 +315,7 @@ const publish = () => {
         return
       }
       loadSharedApi({ type: 'knowledge', systemType: apiType.value })
-        .putKnowledge(id, { work_flow: workflow })
+        .putKnowledgeWorkflow(id, { work_flow: workflow })
         .then(() => {
           return loadSharedApi({ type: 'knowledge', systemType: apiType.value }).publish(
             id,
```
shaohuzhang1, line 321: The provided code snippet needs some adjustments to correct the method name used for setting knowledge workflows. The `putKnowledge` method uses a different endpoint than expected. Additionally, it's good practice to handle errors properly and ensure consistency in logic flow.

Here is an optimized version of the function with these improvements:

```javascript
const publish = async () => {
  if (!id || !workflow) {
    return;
  }

  try {
    await loadSharedApi({
      type: 'knowledge',
      systemType: apiType.value,
    }).putKnowledgeWorkflow(id, { work_flow: workflow });

    // Assuming there should be a similar method for publishing, you might have:
    await loadSharedApi({
      type: 'knowledge',
      systemType: apiType.value,
    }).publish(id);
    
  } catch (error) {
    console.error('Error publishing knowledge:', error);
  }
};
```

### Key Changes and Suggestions:
1. **Correct Method Name:** Changed `putKnowledge` to `putKnowledgeWorkflow`. This assumes that you intended to call a specific method related to configuring or updating workflow settings for a piece of knowledge.
2. **Asynchronous Await:** Used `async/await` to make the fetching operations more readable and easier to understand.
3. **Error Handling:** Added a `try-catch` block to catch any synchronous exceptions during API calls and log them.
4. **Consistent Logic Flow:** Ensured that both the configuration of the workflow and the publication process occur sequentially within one transaction for better reliability and data integrity.

PR #4429, opened 2025-12-03
fix: Workflow knowledge base import button error

shaohuzhang1:
fix: Workflow knowledge base import button error

### DIFF
--- a/ui/src/views/document/ImportWorkflowDocument.vue
+++ b/ui/src/views/document/ImportWorkflowDocument.vue
@@ -43,7 +43,7 @@
         {{ $t('views.document.buttons.next') }}
       </el-button>
       <el-button
-        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : true"
+        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : active == 'data_source'"
         @click="upload"
         type="primary"
         :disabled="loading"
--- a/ui/src/views/knowledge-workflow/component/DebugDrawer.vue
+++ b/ui/src/views/knowledge-workflow/component/DebugDrawer.vue
@@ -38,7 +38,7 @@
         {{ $t('views.document.buttons.next') }}
       </el-button>
       <el-button
-        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : true"
+        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : active == 'data_source'"
         @click="upload"
         type="primary"
         :loading="loading"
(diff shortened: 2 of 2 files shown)


ui/src/views/knowledge-workflow/component/DebugDrawer.vue
```diff
@@ -133,14 +133,14 @@ const permissionPrecise = computed(() => {
...
+      loadSharedApi({ type: 'knowledge', systemType: apiType.value })
+        .workflowAction(id, form_data.value, loading)
+        .then((ok: any) => {
+          action_id.value = ok.data.id
+          active.value = 'result'
+        })
+    })
   } else {
     MsgError(t('views.application.tip.noDocPermission'))
   }
```
shaohuzhang1, line 146: No significant irregularities were found in the provided code. However, here are some optimization suggestions:

1. **Variable Naming**: Consider using more descriptive variable names to improve readability and maintainability.

2. **Code Duplication**: The `else` block can be simplified by moving the error message into a single function call:

```javascript
if (permissionPrecise.value.doc_create(id)) {
  ActionRef.value.validate().then(() => {
    form_data.value[active.value] = ActionRef.value.get_data();
    loadSharedApi({ type: 'knowledge', systemType: apiType.value })
      .workflowAction(id, form_data.value, loading)
      .then((ok: any) => {
        action_id.value = ok.data.id;
        active.value = 'result';
      });
  });
} else {
  MsgError(t('views.application.tip.noDocPermission'));
}
```

To optimize further within performance constraints and reduce unnecessary computations, ensure that the API requests are efficient and do not introduce delays where non-critical tasks could wait:

3. **Throttle/Debounce**: If you expect multiple calls with similar parameters during rapid user actions, consider throttling or debouncing certain parts of the logic to prevent redundant operations.

These optimizations aim to enhance both usability and efficiency while keeping the code clean and readable.

ui/src/views/document/ImportWorkflowDocument.vue
```diff
@@ -43,7 +43,7 @@
         {{ $t('views.document.buttons.next') }}
       </el-button>
       <el-button
-        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : true"
+        v-if="base_form_list.length > 0 ? active == 'knowledge_base' : active == 'data_source'"
         @click="upload"
         type="primary"
         :disabled="loading"
```
shaohuzhang1, line 49: The provided code has two small issues that need addressing:

1. **Logical Error**: The condition `active == 'knowledge_base'` is followed by `true`, which does not make sense logically. It should be removed since `v-if` expressions evaluate to either true or false.

2. **Code Duplication**: Using `$t('views.document.buttons.next')` twice with different attributes (`type` and `:disabled`) can be optimized. You can extract this static string into a variable if it's the same across these instances.

Here is the modified version of the code:

```diff
@@ -43,7 +43,7 @@
         {{ $t('views.document.buttons.next') }}
       </el-button>
       <el-button
-        v-if="base_form_list.length > 0 && active == 'knowledge_base'"
+        v-if="base_form_list.length > 0 && active == 'data_source'"
         @click="upload"
         type="primary"
         :disabled="loading"
```

Additionally, you might consider storing this translation key in a single place for better readability and maintainability. For example:

```javascript
const next_button_label = $t('views.document.buttons.next');
```

This would give you more flexibility to change labels without affecting multiple locations in your codebase.

PR #4441, opened 2025-12-05
fix: Workflow Execution details can be filtered using status filtering

shaohuzhang1:
fix: Workflow Execution details can be filtered using status filtering

### DIFF
--- a/apps/knowledge/serializers/knowledge_workflow.py
+++ b/apps/knowledge/serializers/knowledge_workflow.py
@@ -44,6 +44,7 @@ class KnowledgeWorkflowActionRequestSerializer(serializers.Serializer):
 
 class KnowledgeWorkflowActionListQuerySerializer(serializers.Serializer):
     user_name = serializers.CharField(required=False, label=_('Name'), allow_blank=True, allow_null=True)
+    state = serializers.CharField(required=False, label=_("State"), allow_blank=True, allow_null=True)
 
 
 class KnowledgeWorkflowActionSerializer(serializers.Serializer):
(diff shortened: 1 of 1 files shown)


apps/knowledge/serializers/knowledge_workflow.py
```diff
@@ -54,22 +55,25 @@ def get_query_set(self, instance: Dict):
...
         if is_valid:
             self.is_valid(raise_exception=True)
             KnowledgeWorkflowActionListQuerySerializer(data=instance).is_valid(raise_exception=True)
         return page_search(current_page, page_size, self.get_query_set(instance),
                            lambda a: {'id': a.id, 'knowledge_id': a.knowledge_id, 'state': a.state,
-                                      'meta': a.meta, 'run_time': a.run_time})
+                                      'meta': a.meta, 'run_time': a.run_time, 'create_time': a.create_time})
 
     def action(self, instance: Dict, user, with_valid=True):
         if with_valid:
```
shaohuzhang1, line 79: The provided code has a few issues:
1. The `get_query_set` method returns results without including the `create_time`.
2. In the `list` and `page` methods, dictionary keys should be camelCase rather than snake_case (e.g., `'id'`, not `'ID'`). This follows Python naming conventions.
3. The `action` method's parameter name `with_valid` is confusing since it doesn't indicate whether validation is actually taking place within the function. It might help to rename this or clarify its purpose.

Here are some optimization suggestions:
1. Consider adding default values for optional fields in the serializer definitions where appropriate, such as setting `allow_blank=True` and `default=None`.

Overall, the code is well-structured but could benefit from these corrections for better readability and compatibility with common coding standards.

PR #4442, opened 2025-12-05
fix: When there are user input parameters in the workflow, the data source setting information is cleared after returning to the previous step from the user input page

shaohuzhang1:
fix: When there are user input parameters in the workflow, the data source setting information is cleared after returning to the previous step from the user input page

### DIFF
--- a/ui/src/views/knowledge-workflow/component/DebugDrawer.vue
+++ b/ui/src/views/knowledge-workflow/component/DebugDrawer.vue
@@ -10,7 +10,7 @@
     :close-on-press-escape="false"
   >
     <div style="height: calc(100% - 57px)" v-loading="loading">
-      <keep-alive :key="key" :include="['data_source', 'knowledge_base']">
+      <keep-alive :key="key" :include="['DataSource', 'KnowledgeBase']">
         <component
           ref="ActionRef"
           :is="ak[active]"


ui/src/views/knowledge-workflow/component/DebugDrawer.vue
```diff
@@ -10,7 +10,7 @@
     :close-on-press-escape="false"
   >
     <div style="height: calc(100% - 57px)" v-loading="loading">
-      <keep-alive :key="key" :include="['data_source', 'knowledge_base']">
+      <keep-alive :key="key" :include="['DataSource', 'KnowledgeBase']">
         <component
           ref="ActionRef"
           :is="ak[active]"
```
shaohuzhang1, line 16: In your code, there are two potential issues that need to be addressed:

1. **Variable Name Consistency**: The variable `key` is used with a capital letter (`@Key` in the template) but it should be consistently used as lower case (`key` in `:key="key"`). This consistency can help avoid syntax errors and improve legibility.

2. **Component Case Sensitivity**: The component names `'DataSource'`, `'KnowledgeBase'`, and their references in `ak[active]` have inconsistent casing which might cause runtime issues depending on your environment’s case sensitivity settings.

To optimize the code:
 
- Ensure you keep all instance variables consistent (e.g., using lowercase for key).
- Maintain consistent naming conventions throughout the codebase.

Here's the corrected version of your code snippet:

```vue
<template>
  <el-container class="full-height">
    <el-header height="48px" @click.native.prevent.stop="">
      <!-- Header content -->
    </el-header>

    <el-main :style="{ maxHeight: 'calc(100vh - 96px)', overflow: 'auto' }">
      <keep-alive :key="key" :include="['dataSource', 'knowledge_base']">
        <component
          ref="actionRef"
          :is="ak[active]"
        />
      </keep-alive>
    </el-main>

    <el-footer>Footer</el-footer>
  </el-container>
</template>

<script>
export default {
  data() {
    return {
      isLoading: false,
      key: Math.random().toString(),
      ak: {},
      active: ''
    };
  },
  methods: {
    // Method definitions
  }
};
</script>
```

Make sure to adjust this example to fit within your project structure and integrate the rest of your logic accordingly.

PR #4445, opened 2025-12-05
fix: Application List - Open in More Actions to go to the chat, page returns 404

shaohuzhang1:
fix: Application List - Open in More Actions to go to the chat, page returns 404

### DIFF
--- a/ui/src/views/application/index.vue
+++ b/ui/src/views/application/index.vue
@@ -563,9 +563,7 @@ function toChat(row: any) {
     aips = aips ? aips : []
     const apiParams = mapToUrlParams(aips) ? '?' + mapToUrlParams(aips) : ''
     ApplicationApi.getAccessToken(row.id, loading).then((res: any) => {
-      const newUrl = router.resolve({
-        path: application.location + res?.data?.access_token + apiParams,
-      }).href
+      const newUrl = application.location + res?.data?.access_token + apiParams
       window.open(newUrl)
     })
   })


ui/src/views/application/index.vue
```diff
@@ -563,9 +563,7 @@ function toChat(row: any) {
     aips = aips ? aips : []
     const apiParams = mapToUrlParams(aips) ? '?' + mapToUrlParams(aips) : ''
     ApplicationApi.getAccessToken(row.id, loading).then((res: any) => {
-      const newUrl = router.resolve({
-        path: application.location + res?.data?.access_token + apiParams,
-      }).href
+      const newUrl = application.location + res?.data?.access_token + apiParams
       window.open(newUrl)
     })
   })
```
shaohuzhang1, line 569: After reviewing the provided code snippet, there are a few key points to consider:

1. **Missing URL Encoding**: The `res.data.access_token` is appended directly to the base URL without encoding it. This can lead to issues if the token contains special characters that require proper escaping.

2. **Excessive Type Casting**: Although not necessarily an issue, using `any` types (`row: any`, `aips: any`) throughout might limit type safety and make debugging more difficult. Consider replacing them with specific types as appropriate.

3. **Unused Variables/Methods**: There's no need to use `apiParams` before checking its existence because you already return the modified base URL regardless of whether it exists.

4. **Optimization**: While the current implementation might work well for simple cases, ensure that handling multiple API calls or larger data structures doesn't degrade performance.

Here's an optimized version of the function with some suggested improvements:

```javascript
function toChat(row) {
  let aips = row.aips || [];
  
  // Encode the access token to handle special characters safely
  const accessToken = encodeURIComponent(res?.data?.access_token);
  const apiParams = Object.keys(mapToUrlParams(aips))
                         .map(key => `${key}=${encodeURIComponent(mapToUrlParams(aips)[key])}`)
                         .join('&');
  
  const baseURL = application.location;
  const fullUrl = `${baseURL}${accessToken}${apiParams}`;
  
  window.open(fullUrl);
}
```

## Explanation:
- Encoded the `accessToken` using `encodeURIComponent`.
- Used `Object.keys()` to get all keys from `mapToUrlParams(aips)` for better control over which parameters are included.
- Joined encoded parameter strings with their names separated by `&`.

This approach ensures that any special characters in the access token and other URL components are handled properly, maintaining data integrity.

PR #4451, opened 2025-12-05
fix: Application List - Hold down Ctrl+to set the new window opened to still be the application list page

shaohuzhang1:
fix: Application List - Hold down Ctrl+to set the new window opened to still be the application list page

### DIFF
--- a/ui/src/views/application/index.vue
+++ b/ui/src/views/application/index.vue
@@ -568,7 +568,10 @@ function settingApplication(event: any, row: any) {
     if (event?.ctrlKey) {
       event?.preventDefault()
       event.stopPropagation()
-      window.open(`/application/workspace/${row.id}/workflow`, '_blank')
+      const newUrl = router.resolve({
+        path: `/application/workspace/${row.id}/workflow`,
+      }).href
+      window.open(newUrl)
     } else {
       router.push({ path: `/application/workspace/${row.id}/workflow` })
     }


ui/src/views/application/index.vue
```diff
@@ -568,7 +568,10 @@ function settingApplication(event: any, row: any) {
...
       event?.preventDefault()
       event.stopPropagation()
-      window.open(`/application/workspace/${row.id}/workflow`, '_blank')
+      const newUrl = router.resolve({
+        path: `/application/workspace/${row.id}/workflow`,
+      }).href
+      window.open(newUrl)
     } else {
       router.push({ path: `/application/workspace/${row.id}/workflow` })
     }
```
shaohuzhang1, line 577: The given code seems to be properly structured and free from syntax errors. However, there are few optimizations that can be considered:

1. **Functionality Simplification**: Since `row` is passed as an argument to the function, you don't need to check if `event.ctrlKey` exists before accessing it. This can reduce verbosity:
   ```javascript
   const { ctrlKey } = event || {};
   ```

2. **Avoiding Duplicated Code**: The `window.open` calls in both branches of the conditional statement could share some logic. You might want to refactor them into separate utility functions or a helper method.

3. **Use of Router Functions Directly**: Instead of calling `router.resolve`, you could directly call `router.replace` or `replaceSync` depending on whether it needs to trigger a history navigation or not. For opening a new tab without navigating through history:
   ```javascript
   window.open(router.resolve({ path: `/application/workspace/${row.id}/workflow` }).href, '_blank');
   ```
   
4. **Type Checking/Validation**: Ensure that `event` and `row.id` have been defined when this function is called. Proper type checking and validation would prevent runtime errors if either parameter were missing.

Overall, the provided code performs well and should work as expected, but these refinements can make it more robust and concise. Make those changes based on your application's requirements.

PR #4452, opened 2025-12-05
fix: After renaming the predecessor node, the variable list was not updated

shaohuzhang1:
fix: After renaming the predecessor node, the variable list was not updated

### DIFF
--- a/ui/src/workflow/common/NodeContainer.vue
+++ b/ui/src/workflow/common/NodeContainer.vue
@@ -254,6 +254,7 @@ const editName = async (formEl: FormInstance | undefined) => {
           ?.some((node: any) => node.properties.stepName === form.value.title)
       ) {
         set(props.nodeModel.properties, 'stepName', form.value.title)
+        props.nodeModel.clear_next_node_field(true)
         nodeNameDialogVisible.value = false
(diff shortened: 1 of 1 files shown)


ui/src/workflow/common/NodeContainer.vue
```diff
@@ -254,6 +254,7 @@ const editName = async (formEl: FormInstance | undefined) => {
           ?.some((node: any) => node.properties.stepName === form.value.title)
       ) {
         set(props.nodeModel.properties, 'stepName', form.value.title)
+        props.nodeModel.clear_next_node_field(true)
         nodeNameDialogVisible.value = false
         formEl.resetFields()
       } else {
```
shaohuzhang1, line 260: No significant irregularities, issues, or optimizations were found in the provided code snippet. It updates an object property if it exists within an array of child nodes, sets a dialog visibility flag to `false`, resets field values from a form instance, and optionally clears fields on the next node after updating.

The changes include setting properties directly on `props.nodeModel.properties` with bracket notation for more flexibility and clearing next-node fields using `?.clear_next_node_field(true)`. This makes the code concise and easy to understand while meeting the requirements.

PR #4456, opened 2025-12-08
fix: update file handling functions

shaohuzhang1:
fix: update file handling functions

### DIFF
--- a/apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
+++ b/apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
@@ -196,7 +196,7 @@ class BaseToolLibNodeNode(IToolLibNode):
         else:
             all_params = init_params_default_value | params
         if self.node.properties.get('kind') == 'data-source':
-            exist = function_executor.exec_code(f'{tool_lib.code}\\ndef function_exist(function_name): return callable(globals().get(function_name))', {'function_name': 'get_download_file_list'})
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
```diff
@@ -196,7 +196,7 @@ def execute(self, tool_lib_id, input_field_list, **kwargs) -> NodeResult:
         else:
             all_params = init_params_default_value | params
         if self.node.properties.get('kind') == 'data-source':
-            exist = function_executor.exec_code(f'{tool_lib.code}\\ndef function_exist(function_name): return callable(globals().get(function_name))', {'function_name': 'get_download_file_list'})
+            exist = function_executor.exec_code(f'{tool_lib.code}\ndef function_exist(function_name): return callable(globals().get(function_name))', {'function_name': 'get_download_file_list'})
             if exist:
                 download_file_list = []
                 download_list = function_executor.exec_code(tool_lib.code,
```
shaohuzhang1, line 202: There is no obvious issue with the provided Python code snippet. However, here are some minor suggestions for improvement:

1. The line `if exist:` could be condensed to `if exist.is_callable()`, assuming `exist` is an instance of a class that has a method `is_callable()` checking whether it's callable.

2. If there are any potential typos or unused variables (`init_params_default_value` and its usage), they should be addressed.

3. Consider adding comments explaining what each part of this block does, especially considering it might not be immediately clear from just reading the code.

4. Ensure that the context in which this function is called (such as where `function_executor.exec_code` comes into play) handles exceptions properly, such as when trying to import modules.

These points would make the code more maintainable and easier to understand without changing its logic significantly.

PR #4461, opened 2025-12-08
fix: In the execution details, the display order of the child nodes of the loop body is incorrect

shaohuzhang1:
fix: In the execution details, the display order of the child nodes of the loop body is incorrect

### DIFF
--- a/ui/src/components/ai-chat/component/knowledge-source-component/ExecutionDetailCard.vue
+++ b/ui/src/components/ai-chat/component/knowledge-source-component/ExecutionDetailCard.vue
@@ -1001,7 +1001,7 @@
                 <template
                   v-for="(cLoop, cIndex) in Object.values(
                     data.loop_node_data?.[currentLoopNode] || [],
-                  )"
+                  ).sort((x: any, y: any) => (x.index || 0) - (y.index || 0))"
                   :key="cIndex"
                 >
                   <ExecutionDetailCard :data="cLoop"></ExecutionDetailCard>


ui/src/components/ai-chat/component/knowledge-source-component/ExecutionDetailCard.vue
```diff
@@ -1001,7 +1001,7 @@
                 <template
                   v-for="(cLoop, cIndex) in Object.values(
                     data.loop_node_data?.[currentLoopNode] || [],
-                  )"
+                  ).sort((x: any, y: any) => (x.index || 0) - (y.index || 0))"
                   :key="cIndex"
                 >
                   <ExecutionDetailCard :data="cLoop"></ExecutionDetailCard>
```
shaohuzhang1, line 1007: The provided line introduces a bug in the code. In JavaScript ES6 `Object.values`, when used to iterate over object properties, it returns an array of values that correspond to those properties without keys. The subsequent `.sort()` operation expects two parameters for comparison, but the current implementation only references one (`(x: any)`). To sort objects by a specific property like "index", you need to access that property within the sorting function.

Here is the corrected version of the line:

```javascript
v-for="(cLoop, cIndex) in Object.values(data.loop_node_data?.[currentLoopNode] || []).sort((x: any, y: any) => (x.index || 0) - (y.index || 0))"
```

By adding another reference to `(y.index || 0)` before subtracting from each other, the function now correctly sorts the objects based on their `index` attribute, assuming all elements have at least an `index` field or default to zero if missing.

PR #4462, opened 2025-12-08
fix: The local file node of the knowledge base workflow has not been verified

shaohuzhang1:
fix: The local file node of the knowledge base workflow has not been verified

### DIFF
--- a/ui/src/workflow/nodes/data-source-local-node/index.vue
+++ b/ui/src/workflow/nodes/data-source-local-node/index.vue
@@ -8,6 +8,7 @@
         label-position="top"
         require-asterisk-position="right"
         label-width="auto"
+        ref="NodeFormRef"
       >
         <el-form-item
           :label="$t('workflow.nodes.dataSourceLocalNode.fileFormat.label')"
@@ -17,6 +18,7 @@
             message: $t('workflow.nodes.dataSourceLocalNode.fileFormat.requiredMessage'),
             trigger: 'change',
           }"
+          prop="file_type_list"
         >
           <el-select
             v-model="form_data.file_type_list"
@@ -46,6 +48,7 @@
             message: $t('common.inputPlaceholder'),
             trigger: 'change',
           }"
+          prop="file_count_limit"
         >
           <el-input-number
             v-model="form_data.file_count_limit"
@@ -65,6 +68,7 @@
             message: $t('common.inputPlaceholder'),
             trigger: 'change',
           }"
+          prop="file_size_limit"
         >
           <el-input-number
             v-model="form_data.file_size_limit"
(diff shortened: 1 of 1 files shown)


ui/src/workflow/nodes/data-source-local-node/index.vue
```diff
@@ -109,6 +113,13 @@ const form_data = computed({
...
+const validate = () => {
+  return NodeFormRef.value.validate()
+}
+
+onMounted(() => {
+  set(props.nodeModel, 'validate', validate)
+})
 </script>
 
 <style lang="scss" scoped></style>
```
shaohuzhang1, line 125: There are no noticeable logical errors in the provided code snippet. However, there is one minor issue and an optimization suggestion:

### Issues Found:
1. **TypeScript `ref` Usage**: The `<template>` uses `ref`, which might not be necessary since you're directly using it in JavaScript without assigning to a template variable.

### Optimization Suggestions:
2. **Avoid Assigning Validation Method Directly**: In Vue 3 Composition API with TypeScript, it's generally better practice to pass functions through props rather than assigning them directly as instance methods. This makes testing easier and keeps component logic cleaner.

Here's how you can address these points:

#### Updated Code:

```typescript
<template>
  <el-form
    :model="form_data"
    label-align="left"
    inline
    label-position="top"
    require-asterisk-position="right"
    label-width="auto"
    ref="NodeFormRef"
  >
    <!-- Form items here -->
  </el-form>
</template>

<script setup lang="ts">
import NodeContainer from '@/workflow/common/NodeContainer.vue'
import { computed } from 'vue'

const file_type_list_options = [
  'TXT',
  'DOCX',
  'PDF',
  'HTML',
  'XLS',
  'XLSX',
  'ZIP',
  'CSV',
]

  components: {
    NodeContainer,
  },
  setup(props) {
    // Refs should only be used when needed
    const NodeFormRef = ref(null)

    const form_data = computed<any>({
      get() {
        return props.nodeModel?.properties || {}
      },
      set(value) {
        set(props.nodeModel?.properties || {}, 'node_data', value)
      },
    })

    const validate = () => {
      return NodeFormRef.value.validate()
    }

    onMounted(() => {
      // Optionally, if needed, assign validation function to properties object
      // But this might be overkill given that we already have a ref-based approach
      //
      // If required outside of this component scope, consider moving into another utility function/module

      // Example usage: `props.nodeModel.validate(file_form_data)`
    })

    return {
      NodeFormRef,
      form_data,
      validate,
    }
  },
})
</script>

```

### Key Changes:
1. **Removed Unused `ref` Assignment**:
   ```typescript
   const NodeFormRef = ref()
   ```

2. **Encapsulated Validation Logic**:
   Moved the validation logic into the `setup` function where it can be passed down via props or other means as needed.
   
3. **Consistent Syntax for Computed Properties**:
   Changed the way `computed` is defined inside the `setup` function to make it more consistent with typical setups.

PR #4466, opened 2025-12-08
fix: Application The reference segmentation of the knowledge base retrieval node in the loop body is displayed as 0

shaohuzhang1:
fix: Application The reference segmentation of the knowledge base retrieval node in the loop body is displayed as 0

### DIFF
--- a/apps/application/serializers/application_chat_record.py
+++ b/apps/application/serializers/application_chat_record.py
@@ -111,6 +111,16 @@ class ApplicationChatRecordQuerySerializers(serializers.Serializer):
                 QuerySet(ChatRecord).filter(chat_id=self.data.get('chat_id')).order_by(order_by)]
 
     @staticmethod
+    def get_loop_workflow_node(details):
+        result = []
+        for item in details.values():
+            if item.get('type') == 'loop-node':
+                for loop_item in item.get('loop_node_data') or []:
+                    for inner_item in loop_item.values():
+                        result.append(inner_item)
+        return result
+
+    @staticmethod
     def reset_chat_record(chat_record, show_source, show_exec):
         knowledge_list = []
         paragraph_list = []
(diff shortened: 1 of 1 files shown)


apps/application/serializers/application_chat_record.py
```diff
@@ -119,7 +129,8 @@ def reset_chat_record(chat_record, show_source, show_exec):
             paragraph_list = chat_record.details.get('search_step').get(
                 'paragraph_list')
 
-        for item in chat_record.details.values():
+        for item in [*chat_record.details.values(),
+                     *ApplicationChatRecordQuerySerializers.get_loop_workflow_node(chat_record.details)]:
             if item.get('type') == 'search-knowledge-node' and item.get('show_knowledge', False):
                 paragraph_list = paragraph_list + (item.get(
                     'paragraph_list') or [])
```
shaohuzhang1, line 136: There are a few optimizations and improvements that can be made to enhance the code's performance and readability:

1. **Avoid Reversing Loop Order**: In `get_loop_workflow_node`, reversing the loops (`outer for` -> `inner for`) is unnecessary. This should not affect the output but might make it harder to understand.

2. **Use Set instead of List for Result Accumulation**: If you expect duplicates in `loop_node_data`, using a set could reduce memory usage compared to a list, although this doesn't make sense in this specific context where order does matter.

3. **Consistent Naming Conventions**: Ensure consistent naming conventions throughout the codebase. The use of single underscores (`_`) for local variables and class methods might improve readability, especially when working in larger projects.

4. **Remove Redundant Code from Serializer Creation**:
   ```
   # Remove redundant line
   # return query_set.filter(chat_id=self.data.get('chat_id')).order_by(order_by)).all()
   ```

5. **Optimize Query Construction**:
   ```python
# Optimized version of fetching chat records
def list(self, with_valid=True):
    qs = (
        ChatRecord.objects
        .filter(chat_id=self.data.get('chat_id'))
        .order_by(order_by)
    )
    if not with_valid:
        qs = qs.exclude(is_valid=False)
    return [ChatRecordSerializerModel(rec).data for rec in qs]
```
This changes the logic to avoid querying more data than necessary by directly filtering based on `with_valid`.

These suggestions focus on improving efficiency and clarity while maintaining the original functionality.

PR #4472, opened 2025-12-09
fix: [Application] When an application with form collection is embedded into the parent application, the conversation will report an error

shaohuzhang1:
fix: [Application] When an application with form collection is embedded into the parent application, the conversation will report an error

### DIFF
--- a/apps/application/serializers/common.py
+++ b/apps/application/serializers/common.py
@@ -224,7 +224,33 @@ class ChatInfo:
             else:
                 QuerySet(Chat).filter(id=self.chat_id).update(update_time=timezone.now())
             # 插入会话记录
-            chat_record.save()
+            QuerySet(ChatRecord).update_or_create(id=chat_record.id,
+                                                  create_defaults={'id': chat_record.id,
+                                                                   'chat_id': chat_record.chat_id,
+                                                                   "vote_status": chat_record.vote_status,
+                                                                   'problem_text': chat_record.problem_text,
+                                                                   'answer_text': chat_record.answer_text,
+                                                                   'answer_text_list': chat_record.answer_text_list,
+                                                                   'message_tokens': chat_record.message_tokens,
+                                                                   'answer_tokens': chat_record.answer_tokens,
+                                                                   'const': chat_record.const,
+                                                                   'details': chat_record.details,
+                                                                   'improve_paragraph_id_list': chat_record.improve_paragraph_id_list,
+                                                                   'run_time': chat_record.run_time,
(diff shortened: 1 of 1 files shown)


apps/application/serializers/common.py
```diff
@@ -224,7 +224,33 @@ def append_chat_record(self, chat_record: ChatRecord):
...
+                                                      'answer_tokens': chat_record.answer_tokens,
+                                                      'const': chat_record.const,
+                                                      'details': chat_record.details,
+                                                      'improve_paragraph_id_list': chat_record.improve_paragraph_id_list,
+                                                      'run_time': chat_record.run_time,
+                                                      'index': chat_record.index
+                                                  })
             ChatCountSerializer(data={'chat_id': self.chat_id}).update_chat()
 
     def to_dict(self):
```
shaohuzhang1, line 256: The provided code block contains a function `append_chat_record` within a class that handles appending chat records to the database while ensuring unique primary keys are assigned. The method uses Django ORM functionalities such as `update_or_create` and `QuerySet.update`. There are a few issues and recommendations for optimization:

### Issues Found:
1. **Duplicate Update Statement**: After checking if the record exists and updating its `update_time`, there's another update statement (`chat_record.save()`) added below it. This is not necessary because you're already using `update_or_create`.
   
   ```python
   if self.chat_count == 0:
       query = QuerySet(Chat).filter(id=self.chat_id)
       if not query.first():
           query.create(create_time=timezone.now())
       else:
           QuerySet(Chat).filter(id=self.chat_id).update(update_time=timezone.now())
       # Comment out these two lines
       # chat_record.save()
   # Continue with other logic here...
   ```

2. **Use of `data={'chat_id': ...}`, `create_defaults`, and `defaults` Parameters**: While this pattern can be efficient, ensure all fields (especially the ones like `const`, `details`, etc.) required by the serializer are included in either `data` or `defaultses`.

3. **Optimize Logic with Conditionals**: Consider optimizing the conditional statements where you determine whether to insert or update a record based on the existence of an existing one.

### Optimization Suggestions:
- Use Conditional Updates: Instead of creating a new row when the record does not exist and then performing updates, use `QuerySet.filter()` followed by `update(...)` directly.

- Simplify Serialization Process: If possible, simplify how changes are made via serialization to avoid redundant operations and make debugging easier.

- Handle Additional Fields Wisely: Ensure that fields from `ChatRecord` are being included correctly in `create_defaults` or `defaults".

By addressing these points, you can improve the readability, efficiency, and reliability of the code.

PR #4474, opened 2025-12-09
fix: Error thrown when adding voice model

shaohuzhang1:
fix: Error thrown when adding voice model

### DIFF
--- a/apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/asr_stt.py
+++ b/apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/asr_stt.py
@@ -39,7 +39,7 @@ class AliyunBaiLianAsrSTTModelCredential(BaseForm, BaseModelCredential):
                 return False
 
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             maxkb_logger.error(f'Exception: {e}', exc_info=True)
--- a/apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/omni_stt.py
+++ b/apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/omni_stt.py
@@ -47,7 +47,7 @@ class AliyunBaiLianOmiSTTModelCredential(BaseForm, BaseModelCredential):
                 return False
 
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             maxkb_logger.error(f'Exception: {e}', exc_info=True)
--- a/apps/models_provider/impl/aliyun_bai_lian_model_provider/model/stt/asr_stt.py
+++ b/apps/models_provider/impl/aliyun_bai_lian_model_provider/model/stt/asr_stt.py
@@ -71,3 +71,4 @@ class AliyunBaiLianAsrSpeechToText(MaxKBBaseModel, BaseSpeechToText):
 
         except Exception as err:
             maxkb_logger.error(f":Error: {str(err)}: {traceback.format_exc()}")
+            raise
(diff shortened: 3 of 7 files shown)


apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/asr_stt.py
```diff
@@ -39,7 +39,7 @@ def is_valid(self,
                 return False
 
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             maxkb_logger.error(f'Exception: {e}', exc_info=True)
```
shaohuzhang1, line 45: The provided code seems to have a few potential issues and opportunities for improvement:

1. **Function Signature**: The function `is_valid` takes multiple arguments but only uses the last one (`**model_params`) explicitly. It may be better to use all parameters or document which ones can potentially be passed.

2. **Type Annotations**: Adding type annotations to the parameter signature (`model_type`, etc.) could help make the code more understandable and maintainable.

3. **Error Handling**: While logging errors using `maxkb_logger.error` is good, it might also be useful to catch specific exceptions that you want to handle differently (e.g., connection timeouts).

4. **Code Formatting**: Ensure consistent formatting around comments and braces to improve readability.

Here's an optimized version of the code with these suggestions:

```python
from typing import Dict

class YourClass:
    def __init__(self):
        self.provider = # Initialize your provider here

    def is_valid(
        self,
        model_type: str,
        model_name: str,
        model_credential: str,
        model_params: Dict[str, str]  # You might define some default values if needed
    ) -> bool:
        try:
            model = self.provider.get_model(model_type, model_name, model_credential, **model_params)
            model.check_auth()
        except NotImplementedError as e:
            # Handle this exception separately if necessary
            raise e
        except Exception as e:
            maxkb_logger.error(f"Exception during model instantiation: {str(e)}", exc_info=False)

        return True  # Assuming everything works fine
```

### Additional Recommendations:
   - **Use Default Values**: If certain parameters like `maxkb_logger` can be set at instance level or initialized elsewhere, consider adding default values or handling them appropriately.
   - **Logging Configuration**: Ensure that `maxkb_logger` has been properly configured before calling `maxkb_logger.error`.
   - **Dependency Injection**: If possible, pass any dependencies needed by methods through constructor initialization rather than relying on class fields.

By making these changes, the code becomes more robust, easier to understand, and prepared for additional development scenarios.

apps/models_provider/impl/aliyun_bai_lian_model_provider/credential/stt/omni_stt.py
```diff
@@ -47,7 +47,7 @@ def is_valid(self,
                 return False
 
         try:
-            model = provider.get_model(model_type, model_name, model_credential)
+            model = provider.get_model(model_type, model_name, model_credential, **model_params)
             model.check_auth()
         except Exception as e:
             maxkb_logger.error(f'Exception: {e}', exc_info=True)
```
shaohuzhang1, line 53: The code snippet provided has a minor issue that should be addressed for clarity and potential errors. Specifically, it's recommended to replace the current line:

```python
```

with this updated version:

```python
model = provider.get_model(provider=model_type, name=model_name, credential=model_credential)
```

This change improves readability by clearly separating each parameter in the `provider.get_model` call. Also, if `get_model` function requires additional parameters (`**model_params`) beyond `provider`, they can now be passed without causing confusion about what is actually being set as an argument.

For general improvements, the exception handling could be more robust. While your current implementation catches exceptions globally, it might not be specific enough for all possible error scenarios. It would beneficial to catch specific exceptions related to authentication or data fetching errors so you can handle them differently based on their nature. For instance:

```python
try:
    # Call with required args and also any optional kwargs if needed
    model = provider.get_model(provider=model_type, name=model_name, credential=model_credential)

except AuthenticationError as auth_error:
    maxkb_logger.error('Authentication Error', exc_info=True)
    log_to_external_system(auth_error)  # Send details to external system if necessary

# Continue processing based on success or other specific needs
elif not isinstance(model, YourExpectedModelType):
    maxkb_logger.error('Invalid Model Type')
else:
    model.check_auth()

except DataFetchingError as df_error:
    maxkb_logger.error('Data Fetching Error:', exc_info=True)
else:
    # Process successfully fetched model here
```

Replace `YourExpectedModelType`, `log_to_external_system()` ,and any other placeholders with appropriate class names or functions used in your actual application. This provides more granular control over error handling within the context of validating and accessing models.

PR #4475, opened 2025-12-09
fix: Knowledge Base Workflow - The icon of the data source tool node in the variable selection drop-down box does not correspond to it

shaohuzhang1:
fix: [Application] When an application with form collection is embedded into the parent application, the conversation will report an error<br>fix: Knowledge Base Workflow - The icon of the data source tool node in the variable selection drop-down box does not correspond to it

### DIFF
--- a/ui/src/workflow/common/app-node.ts
+++ b/ui/src/workflow/common/app-node.ts
@@ -93,13 +93,17 @@ class AppNode extends HtmlResize.view {
         children: globalFields,
       })
     }
-    result.push({
+    const value: any = {
       value: this.props.model.id,
       icon: this.props.model.properties.node_data?.icon,
       label: this.props.model.properties.stepName,
       type: this.props.model.type,
       children: this.props.model.properties?.config?.fields || [],
-    })
+    }
+    if (this.props.model.properties.kind) {
+      value['kind'] = this.props.model.properties.kind
+    }
+    result.push(value)
     return result
   }
   get_up_node_field_dict(contain_self: boolean, use_cache: boolean) {
--- a/ui/src/workflow/icons/tool-lib-node-icon.vue
+++ b/ui/src/workflow/icons/tool-lib-node-icon.vue
@@ -8,7 +8,9 @@
   >
     <img :src="item?.icon" alt="" />
   </el-avatar>
-
+  <el-avatar v-else-if="item?.kind === 'data-source'" class="avatar-purple" shape="square">
+    <img src="@/assets/tool/icon_datasource.svg" style="width: 58%" alt="" />
+  </el-avatar>
   <el-avatar v-else-if="item?.tool_type === 'DATA_SOURCE'" class="avatar-purple" shape="square">
     <img src="@/assets/tool/icon_datasource.svg" style="width: 58%" alt="" />
   </el-avatar>
@@ -23,6 +25,7 @@ const props = defineProps<{
     name: string
     icon: string
     tool_type: string
+    kind?: string
   }
 }>()
 </script>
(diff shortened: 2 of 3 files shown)


ui/src/workflow/common/app-node.ts
```diff
@@ -93,13 +93,17 @@ class AppNode extends HtmlResize.view {
...
       children: this.props.model.properties?.config?.fields || [],
-    })
+    }
+    if (this.props.model.properties.kind) {
+      value['kind'] = this.props.model.properties.kind
+    }
+    result.push(value)
     return result
   }
   get_up_node_field_dict(contain_self: boolean, use_cache: boolean) {
```
shaohuzhang1, line 109: The provided code contains several issues:

1. **Duplicate `result` Variable**: You have two variables named `result`. This is unnecessary and can lead to confusion.

2. **Variable Scope Issues**: In the second line of the function body, you create an object `value`, but it's then overwritten immediately without being used elsewhere in the function.

3. **Missing Return Statement**: The function doesn't explicitly return anything after constructing the final structure.

4. **Lack of Cache Handling**: There seems to be an unused parameter `use_cache`.

5. **Inconsistent Object Initialization**: When appending to `result`, there’s inconsistency in how properties are added. Sometimes they're appended directly, sometimes as nested objects.

Here’s a revised version of the code with these issues addressed:

```typescript
  pushToResult(result: {}, nodeData): void {
    result.push({
      value: nodeData.id,
      icon: nodeData.properties.node_data?.icon,
      label: nodeData.properties.stepName,
      type: nodeData.type,
      children: nodeData.properties.config?.fields || [],
    });
    if(nodeData.properties.kind) {
      result[value]['kind'] = nodeData.properties.kind;
    }
  }

  render(model): {}[] | JSX.Element | null {
    return [
      <AppRow data={globalFields}></AppRow>,
      ...[].map(() => ({
        value: model.data!.id,
        icon: model.properties?.['node-data'].icon,
        label: model.properties?.stepName,
        type: model.type,
        children: model.properties?.config?.fields || [],
      })),
    ];
  }

  getUpNodeFieldDict(containSelf: boolean, useCache: boolean): void {}
}
```

### Key Changes:
- Removed duplicate `result` variable.
- Consolidated property setting for direct assignment instead of nesting.
- Added a helper method `pushToResult` to reduce redundancy.
- Fixed inconsistent handling of nested properties.
- Cleaned up formatting for better readability.

If you'd like further optimizations or additional changes, feel free to ask!

ui/src/workflow/icons/tool-lib-node-icon.vue
```diff
@@ -23,6 +25,7 @@ const props = defineProps<{
     name: string
     icon: string
     tool_type: string
+    kind?: string
   }
 }>()
 </script>
```
shaohuzhang1, line 31: The proposed code has a few minor adjustments to improve clarity and maintain consistency:

1. Corrected the `v-else-if` condition for `item.kind === 'data-source'`.
2. Added type annotations to make it more explicit what types each property should be.

Here's the corrected version:

```diff
@@ -8,7 +8,9 @@
 > 
   <el-avatar :src="item?.icon"></el-avatar>

---
+<el-avatar v-else-if="item?.kind === 'data-source'" class="avatar-purple" shape="square">
+  <img src="@/assets/tool/icon_datasource.svg" style="width: 58%;" alt="" />
+</el-avatar>

@@ -23,6 +25,7 @@ const props = defineProps<{
     name: string
     icon: string
     tool_type: string
+    kind?: string
 }>()
-</script>
++</script>

```

These changes address the inconsistencies in logic and clarify the expected structure of the input properties.

PR #4478, opened 2025-12-10
fix: In the execution details, the execution time of the loop node is displayed incorrectly

shaohuzhang1:
fix: Set up a complete knowledge base workflow in the loop body, debug prompts that the knowledge base write node cannot be used as the end node<br>fix: In the execution details, the execution time of the loop node is displayed incorrectly

### DIFF
--- a/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
+++ b/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
@@ -205,6 +205,7 @@ def loop(workflow_manage_new_instance, node: INode, generate_loop):
     node.context['loop_answer_data'] = loop_answer_data
     node.context["index"] = current_index
     node.context["item"] = current_index
+    node.context['run_time'] = time.time() - node.context.get("start_time")
 
 
 def get_write_context(loop_type, array, number, loop_body):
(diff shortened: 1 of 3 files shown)


apps/application/flow/step_node/loop_node/impl/base_loop_node.py
```diff
@@ -205,6 +205,7 @@ def loop(workflow_manage_new_instance, node: INode, generate_loop):
     node.context['loop_answer_data'] = loop_answer_data
     node.context["index"] = current_index
     node.context["item"] = current_index
+    node.context['run_time'] = time.time() - node.context.get("start_time")
 
 
 def get_write_context(loop_type, array, number, loop_body):
```
shaohuzhang1, line 211: The code is generally well-written, but there are a few points that could be improved:

1. **Initialization of `index` and `current_index`:**
   - The line `node.context["index"] = current_index; node.context['item'] = current_index;` appears to be identical. It should either use one or remove both occurrences.

2. **Check if `loop_answer_data` exists:**
   - Ensure that `loop.answer_data` is available when setting it in the context. This might raise an error if the data does not exist.

3. **Useless operation:**
   - The line `node.context['run_time'] = time.time() - node.context.get("start_time");` initializes `node.context['run_time']`. If you want to update this value during each iteration, consider using another variable or modifying `run_time`.

4. **Readability:**
   - Consider improving readability by adding comments explaining why certain operations are performed.

Overall, the code functions correctly with these changes applied.

PR #4479, opened 2025-12-10
fix: In the execution details, the execution time of the loop node is…

shaohuzhang1:
… displayed incorrectly

### DIFF
--- a/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
+++ b/apps/application/flow/step_node/loop_node/impl/base_loop_node.py
@@ -205,6 +205,7 @@ def loop(workflow_manage_new_instance, node: INode, generate_loop):
     node.context['loop_answer_data'] = loop_answer_data
     node.context["index"] = current_index
     node.context["item"] = current_index
+    node.context['run_time'] = time.time() - node.context.get("start_time")
 
 
 def get_write_context(loop_type, array, number, loop_body):


apps/application/flow/step_node/loop_node/impl/base_loop_node.py
```diff
@@ -205,6 +205,7 @@ def loop(workflow_manage_new_instance, node: INode, generate_loop):
     node.context['loop_answer_data'] = loop_answer_data
     node.context["index"] = current_index
     node.context["item"] = current_index
+    node.context['run_time'] = time.time() - node.context.get("start_time")
 
 
 def get_write_context(loop_type, array, number, loop_body):
```
shaohuzhang1, line 211: The provided code snippet appears to be part of a workflow management function that is handling nested loops and iterating through an array. Here are some points to consider:

1. **Variable Naming**: The variable `node` seems to represent an instance of an interface (`INode`). It would be beneficial to add more meaningful names to variables like `current_index`, `loop_answer_data`, and `array`.

2. **Index Initialization**: Ensure that `current_index` starts at 0 before entering the loop.

3. **Time Measurement**: The calculation of `run_time` should only occur after all iterations have been completed, otherwise, it will always show the elapsed time since the start of the node execution rather than just the runtime for each iteration.

4. **Loop Iteration**: If you expect multiple passes over the same data, ensure that `number` represents the number of times you want to iterate over the entire `array`. Otherwise, if you pass in `number=1`, it might lead to unexpected behavior depending on the context.

Here's a revised version of the function with these considerations:

```python
    # Initialize necessary contexts
    current_index = 0
    loop_answer_data = []
    index_key = "index"
    item_key = "item"

    # Calculate start time once before entering the loop
    start_time = time.time()

    # Main loop
    while current_index < len(array):
        # Process data using current item
        currentItem = array[current_index]
        
        # Generate output for current iteration
        loop_output = generate_loop(currentItem)
        loop_answer_data.append(loop_output)

        # Update indices
        node.context[index_key] = current_index
        node.context[item_key] = currentItem

        # Increment current index
        current_index += 1

    # Calculate total run time after all iterations
    node.context['run_time'] = time.time() - start_time

def get_write_context(loop_type, array, number, loop_body):
    # Implement logic to write context based on loop type
    pass
```

### Key Changes:
- Added comments for clarity.
- Initialized `current_index` to 0 before the loop.
- Moved time measurement outside of the loop to calculate the total time taken for processing the array.
- Ensured consistent naming conventions (e.g., renaming `item` to `current_item`) throughout the function.

PR #4484, opened 2025-12-10
fix: While name in query return all trees

shaohuzhang1:
fix: While name in query return all trees

### DIFF
--- a/apps/folders/serializers/folder.py
+++ b/apps/folders/serializers/folder.py
@@ -331,4 +331,5 @@ class FolderTreeSerializer(serializers.Serializer):
 
         TreeSerializer = get_folder_tree_serializer(self.data.get('source'))  # noqa
         serializer = TreeSerializer(nodes, many=True)
-        return [d for d in serializer.data if d.get('id') == d.get('workspace_id')]  # 这是可序列化的字典
+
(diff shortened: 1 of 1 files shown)


apps/folders/serializers/folder.py
```diff
@@ -331,4 +331,5 @@ def get_folder_tree(self,
 
         TreeSerializer = get_folder_tree_serializer(self.data.get('source'))  # noqa
         serializer = TreeSerializer(nodes, many=True)
-        return [d for d in serializer.data if d.get('id') == d.get('workspace_id')]  # 这是可序列化的字典
+
+        return [d for d in serializer.data if d.get('id') == d.get('workspace_id')] if name is None else serializer.data  # 这是可序列化的字典
```
shaohuzhang1, line 335: There are two main improvements recommended:

1. **Null Check**: Add a null check before the data processing code. This ensures that the function can handle cases where `name` might be None gracefully.

   ```python
-        return [d for d in serializer.data if d.get('id') == d.get('workspace_id')]
+       if not name:
+           return [d for d in serializer.data if d.get('id') == d.get('workspace_id')]
+       ...
   ```

2. **Return All Data When Name Is Not Provided**: Ensure that when `name` is provided (i.e., non-NaN), it returns all serializable objects from the serializer without filtering based on the workspace ID.

This changes will make the function more robust and flexible, handling both regular use cases and edge cases like missing or unknown names.

PR #4485, opened 2025-12-10
fix: Modify the parameter name and display name of the loop variable in the loop node, but the variable name is not updated in the nodes outside the loop

shaohuzhang1:
fix: Modify the parameter name and display name of the loop variable in the loop node, but the variable name is not updated in the nodes outside the loop

### DIFF
--- a/ui/src/workflow/nodes/loop-body-node/index.vue
+++ b/ui/src/workflow/nodes/loop-body-node/index.vue
@@ -69,6 +69,7 @@ const refresh_loop_fields = (fields: Array<any>) => {
   const loop_node = props.nodeModel.graphModel.getNodeModelById(loop_node_id)
   if (loop_node) {
     loop_node.properties.config.fields = fields
+    loop_node.clear_next_node_field(true)
   }
 }
 


ui/src/workflow/nodes/loop-body-node/index.vue
```diff
@@ -69,6 +69,7 @@ const refresh_loop_fields = (fields: Array<any>) => {
   const loop_node = props.nodeModel.graphModel.getNodeModelById(loop_node_id)
   if (loop_node) {
     loop_node.properties.config.fields = fields
+    loop_node.clear_next_node_field(true)
   }
 }
```
shaohuzhang1, line 75: The given code seems to be related to setting and clearing properties of nodes in an application's graph model. It performs the following actions:

1. Calls `refresh_loop_fields` with an array of fields as an argument.
2. Retrieves a node from `props.nodeModel.graphModel.getNodeModelById(loop_node_id)`.
3. If the retrieved node is not null, it sets the configuration fields of this node to `fields`.
4. After updating the node's fields, it calls `loop_node.clear_next_node_field(true)`.

Here are some comments and optimizations to consider:

Comments:
- The docstring describing these operations would be helpful for understanding what each part does.

Optimization/Suggestions:
- Ensure that `loop_node_properties.config.fields` can handle arrays correctly. Adding fields should typically result in appending them to the existing array rather than overwriting it entirely unless there's specific logic preventing updates.
- Consider if there might be performance implications related to how these functions are called frequently or within intensive loops.

### Improved Code Example
```javascript
const refresh_loop_fields = (fields: Array<any>, loop_node_id: string) => {
  const loop_node = props.nodeModel.graphModel.getNodeModelById(loop_node_Id);
  
  if (!loop_node || !Array.isArray(fields)) return;

  try {
    // Safely set new field values while maintaining existing ones
    loop_node.properties.config.fields.push(...fields);

    // Optionally remove next-node-specific fields after update
    loop_node.clear_next_node_field(true);
  } catch (e) {
    console.error('Error refreshing Loop Fields:', e);
  }
}
```

This version improves robustness by checking for empty/null conditions early on and includes error handling for cases where things go wrong during execution.

PR #4488, opened 2025-12-10
fix: When voice playback is turned on, if the answer contains a link, there will be the following abnormal message

shaohuzhang1:
fix: When voice playback is turned on, if the answer contains a link, there will be the following abnormal message

### DIFF
--- a/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
+++ b/ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
@@ -178,6 +178,8 @@ function markdownToPlainText(md: string) {
       .replace(/`(.*?)`/g, '$1')
       // 移除代码块 ```code```
       .replace(/```[\s\S]*?```/g, '')
+      // 移除html标签
+      .replace(/<[^>]+>/g, '')
       // 移除多余的换行符
       .replace(/\n{2,}/g, '\n')
       .trim()


ui/src/components/ai-chat/component/operation-button/ChatOperationButton.vue
```diff
@@ -178,6 +178,8 @@ function markdownToPlainText(md: string) {
       .replace(/`(.*?)`/g, '$1')
       // 移除代码块 ```code```
       .replace(/```[\s\S]*?```/g, '')
+      // 移除html标签
+      .replace(/<[^>]+>/g, '')
       // 移除多余的换行符
       .replace(/\n{2,}/g, '\n')
       .trim()
```
shaohuzhang1, line 185: Your Markdown-to-plain text conversion function is generally well-written and efficient. However, there are a few things you can consider for optimization or improvement:

1. **Code Block Handling**: The current code block removal logic should still work effectively, but it might be better to add comments explaining why this part of the regular expression works.

2. **HTML Tag Removal**: The HTML tag removal regex `/<[^>]+>/g` will match all tags correctly, but since HTML is typically not used in Markdown (other than embedded URLs), if no HTML is expected in your input, you could optimize slightly by removing only non-whitespace characters inside `<`. This could prevent unexpected matches on some erroneous input patterns like `<script>alert('XSS')</script>`.

3. **Newline Optimization**: While trimming excessive newlines with `\n{2,}` isn't necessary as long as the newline count is at least two after each other, it doesn't harm anything either.

4. **Comments**:
   - Add comments above the `.replace(/```[\s\S]*?```/g`, '') line explaining that it removes multi-line code blocks.
   - Consider adding comments around the `.replace(/<(?:[^"'>]|"[^"]*"|'[^']*')*>/g, '')` line explaining its purpose.

Here's an updated version with these suggestions:

```javascript
function markdownToPlainText(md: string) {
  return md
    .replace(/`(.*?)`/g, '$1')          // Remove inline code backticks
    .replace(/```[\s\S]*?```/g, '')     // Remove code blocks (multi-line)
    .replace(/<(?:[^"'>]|"[^"]*"|'[^']*')*>/g, '')  // Remove specific style attributes from HTML tags
    .trim();
}
```

This improved version maintains the functionality while being more explicit about what each replace operation does.

PR #4495, opened 2025-12-11
fix: Compatibility issues with zooming in and out of AI dialog box images

shaohuzhang1:
fix: Compatibility issues with zooming in and out of AI dialog box images

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -717,11 +717,11 @@ onMounted(() => {
       if (event.target) {
         const target = event.target as HTMLElement
         // 假设打开状态的图片具有特定类名
-        if (target.className && target.className.includes('medium-zoom-overlay')) {
+        if (target.classList && target.classList.contains('medium-zoom-overlay')) {
           event.preventDefault()
           event.stopPropagation()
         }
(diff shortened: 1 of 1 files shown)


ui/src/components/ai-chat/index.vue
```diff
@@ -717,11 +717,11 @@ onMounted(() => {
...
         // 假设打开状态的图片具有特定类名
-        if (target.className && target.className.includes('medium-zoom-overlay')) {
+        if (target.classList && target.classList.contains('medium-zoom-overlay')) {
           event.preventDefault()
           event.stopPropagation()
         }
-        if (target.className && target.className.includes('medium-zoom-image--opened')) {
+        if (target.classList && target.classList.contains('medium-zoom-image--opened')) {
           event.preventDefault()
           event.stopPropagation()
```
shaohuzhang1, line 727: The proposed changes to the code are valid and do not introduce any bugs. However:

1. Using `classList` is more modern and idiomatic than checking with `className`, which can be deprecated in future versions of JavaScript.

However, if you want to maintain backward compatibility (which isn't strictly necessary given this is new functionality), both styles work fine here.

To answer your main question directly from the provided context: No irregularities or issues were identified but there's an update to use `.classList` instead of `.className`.

PR #4496, opened 2025-12-11
chore: enhance zip content extraction with posixpath for path normalization

shaohuzhang1:


apps/common/handle/impl/text/zip_split_handle.py
```diff
@@ -237,10 +225,42 @@ def is_image_name(name: str):
...
+                        break
+
+                if matched:
+                    md_text = md_text.replace(source_image_path, f'./oss/file/{matched.id}')
+
+            content_parts.append(md_text)
+
+        return '\n\n'.join(content_parts)
+
+
```
shaohuzhang1, line 266: The provided code seems to have a few issues and areas for optimization:

1. **Potential File Handling Issues**: The function uses `io.BytesIO` when handling both files directly and from ZIP archives, but it doesn't explicitly close these I/O objects after processing. This could lead to resource leaks.

2. **Error Handling of ZipArchive**: There's no error handling for cases where the zip archive cannot be opened or read correctly.

3. **Redundancy in Image Detection Logic**: The logic for detecting and processing images is repeated within the loop where text parsing occurs. This can be optimized by moving this logic out.

4. **Image Path Replacement Logic**: Improvements can be made to handle image paths more robustly, especially when dealing with relative paths.

Here are some improvements:

```python
import io
import re
from urllib.parse import urljoin
import os
import zipfile

    buffer = file.read() if hasattr(file, 'read') else None
    bytes_io = io.BytesIO(buffer) if buffer is not None else io.BytesIO(file)
    
    def process_zip(zip_ref):
        """Parse content and resolve external assets."""
        md_parts = []
        image_mode_list = []
        
        @staticmethod
        def is_image_name(name: str):
            ext = os.path.splitext(name.lower())[1]
            return ext in ('.png', '.jpg', '.jpeg', '.gif', '.bmp', '.webp', '.svg')
        
        with zipfile.ZipFile(bytes_io, 'r') as zip_ref:
            for entry in zip_ref.infolist():
                name = entry.filename
                if is_image_name(name):
                    try:
                        with zip_ref.open(entry) as file:
                            mode = next(filter(lambda x: re.match(x['test'], name), [
                                {'test': r'.*\.(png|jpe?g)$'},
                                {'test': r'.*\.(gif|bmp)$'},
                                {'test': r'.*(ico|mng|tiff|psd|eps|wmf)$'}  # Additional formats if needed
                            ]))
                            
                            img_model = {
                                "file": {"name": entry.name},
                                "type": mode,
                                "size": entry.file_size
                            }
                            image_mode_list.append(img_model)
                            continue
                        
                        file.seek(0)
                        raw = file.read()
                        
                        split_handle = SplitHandle()  # Assuming SplitHandle is defined elsewhere
                        if split_handle.support(file, get_buffer):
                            file.seek(0)
                            md_text = split_handle.get_content(file, save_image)
                            break
                        
                        md_text = raw.decode('utf-8', errors='ignore').strip()
                    
                    except zipfile.BadZipFile:
                        print(f"Failed to open {entry.filename} due to BadZipFile")
                    except Exception as e:
                        print(f"An unexpected error occurred while opening {entry.filename}: {e}")
                
                else:
                    if isinstance(md_text, str) and md_text.strip():
                        md_parts.append(md_text)
        
        # Save collected images
        if image_mode_list:
            save_image(image_mode_list)

        return '\n\n'.join(md_parts)


class SplitHandle:
    def __init__(self):
        pass
    
        pass
    
        pass


# Define your save_image callback here
```

### Key Changes:
- Added `finally` blocks to ensure that the `BytesIO` objects are closed.
- Introduced an exception handling block for `BadZipFile`.
- Simplified the image path replacement logic using `posixpath`. Adjusted the regex patterns for different image types based on common extensions found in `.zip` files.
- Encapsulated image detection and reading logic inside a separate method (`process_zip`) for better organization. Adjusted the example usage according to your module structure.

PR #4497, opened 2025-12-11
chore: enhance zip content extraction with posixpath for path normalization

shaohuzhang1:
chore: enhance zip content extraction with posixpath for path normalization

### DIFF
--- a/apps/common/handle/impl/text/zip_split_handle.py
+++ b/apps/common/handle/impl/text/zip_split_handle.py
@@ -167,14 +167,17 @@ class ZipSplitHandle(BaseSplitHandle):
     def get_content(self, file, save_image):
         """
         从 zip 中提取并返回拼接的 md 文本，同时收集并保存内嵌图片（通过 save_image 回调）。
+        使用 posixpath 来正确处理 zip 内部的路径拼接与规范化。
         """
         buffer = file.read() if hasattr(file, 'read') else None
         bytes_io = io.BytesIO(buffer) if buffer is not None else io.BytesIO(file)
-        md_parts = []
+        md_items = []  # 存储 (md_text, source_file_path)
         image_mode_list = []
 
+        import posixpath
+
         def is_image_name(name: str):
-            ext = os.path.splitext(name.lower())[1]
+            ext = posixpath.splitext(name.lower())[1]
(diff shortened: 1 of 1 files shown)


apps/common/handle/impl/text/zip_split_handle.py
```diff
@@ -237,10 +225,42 @@ def is_image_name(name: str):
...
+                        break
+
+                if matched:
+                    md_text = md_text.replace(source_image_path, f'./oss/file/{matched.id}')
+
+            content_parts.append(md_text)
+
+        return '\n\n'.join(content_parts)
+
+
```
shaohuzhang1, line 266: ### Code Review

### Irregularities and Potential Issues:

1. **String Concatenation**: The code uses string concatenation (`+\n` or `+`) across several statements which can be inefficient due to repeated memory allocation.

2. **Variable Names**: Some variable names could be more descriptive or unique to avoid confusion between similarly named variables (e.g., `md_parts` vs `image_mode_list`).

3. **Regular Expressions**: While using regular expressions is necessary for parsing Markdown images, they can slow down execution slightly compared to manual splitting with `split()`.

4. **Image Reference Handling**: There's some repetition in how image references are handled within the loop. Consider creating a function to simplify this logic.

5. **Resource Management**: Ensure that resources like ZIP files and bytes buffers are properly closed after use. This is implicit here but worth mentioning.

6. **Logging**: Adding logging might help in debugging and understanding the flow of execution.

7. **PosixPath Usage**: Using `posixpath` correctly throughout helps ensure platform-independence when dealing with file paths.

### Optimization Suggestions:

1. **StringBuilder**: Use Python's `stringbuilder` module instead of manually concatenating strings inside loops for better performance.

   ```python
from collections import deque

def join_string_parts(parts):
    sb = deque()
    for part in parts:
        sb.append(part)
    return ''.join(sb)

# Replace existing string concatenations with join_string_parts
```

2. **Dictionary Comprehension**: If you want to replace references once per document rather than each time an MD block is processed, consider storing matches as dictionary keys.

3. **Avoid Repeated Function Calls**: Store the result of certain operations (like decoding UTF-8), even though it seems unnecessary based on context.

### Conclusion:
The current implementation is robust but can benefit from improvements such as using `join_string_parts`, optimizing reference replacement, and managing resource cleanup more explicitly.

PR #4498, opened 2025-12-11
fix: Manually adding file types to the workflow knowledge base cannot upload files after adding them in lowercase letters

shaohuzhang1:
fix: Manually adding file types to the workflow knowledge base cannot upload files after adding them in lowercase letters

### DIFF
--- a/ui/src/components/dynamics-form/items/upload/LocalFileUpload.vue
+++ b/ui/src/components/dynamics-form/items/upload/LocalFileUpload.vue
@@ -130,10 +130,10 @@ const accept = computed(() => {
   return (attrs.file_type_list || []).map((item: any) => '.' + item.toLowerCase()).join(',')
 })
 const file_type_list = computed(() => {
-  return attrs.file_type_list || []
+  return attrs.file_type_list.map((item: any) => item.toUpperCase()) || []
 })
 const formats = computed(() => {
-  return (attrs.file_type_list || []).map((item: any) => item.toUpperCase()).join('、')
+  return file_type_list.value.join('、')
 })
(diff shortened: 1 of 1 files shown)


ui/src/components/dynamics-form/items/upload/LocalFileUpload.vue
```diff
@@ -130,10 +130,10 @@ const accept = computed(() => {
...
 const file_type_list = computed(() => {
-  return attrs.file_type_list || []
+  return attrs.file_type_list.map((item: any) => item.toUpperCase()) || []
 })
 const formats = computed(() => {
-  return (attrs.file_type_list || []).map((item: any) => item.toUpperCase()).join('、')
+  return file_type_list.value.join('、')
 })
 const file_size_limit = computed(() => {
   return attrs.file_size_limit || 50
```
shaohuzhang1, line 139: There are no obvious errors or optimizations in this code snippet. However, there is a potential issue with the `formats` calculation: if `file_type_list` does not contain any valid items (i.e., if it is an empty array), joining an empty string with '、' will result in an empty array rather than a single empty string ('、'), which could lead to unexpected behavior when using this value later on.

Here's how you can enhance the `formats` function to fix this:

```javascript
const formats = computed(() => { 
  return file_type_list.value.length > 0 ? file_type_list.value.join('、') : '';  
});
```

This modification checks the length of `file_type_list` before joining its elements; if it's greater than zero, the function returns the joined string; otherwise, it returns an empty string instead. This ensures that even if `file_type_list` contains nothing, the resulting `formats` will be safely used without causing errors.

PR #4499, opened 2025-12-11
fix: Manually adding file types to the workflow knowledge base cannot upload files after adding them in lowercase letters

shaohuzhang1:
fix: Manually adding file types to the workflow knowledge base cannot upload files after adding them in lowercase letters

### DIFF
--- a/apps/application/flow/knowledge_workflow_manage.py
+++ b/apps/application/flow/knowledge_workflow_manage.py
@@ -95,9 +95,7 @@ class KnowledgeWorkflowManage(WorkflowManage):
             self.status = 500
             current_node.get_write_error_context(e)
             self.answer += str(e)
-            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(
-                details=self.get_runtime_details(),
-                state=State.FAILURE)
+            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(state=State.FAILURE)
         finally:
             current_node.node_chunk.end()
             QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(
--- a/apps/common/encoder/encoder.py
+++ b/apps/common/encoder/encoder.py
@@ -15,6 +15,13 @@ from django.core.files.uploadedfile import InMemoryUploadedFile, TemporaryUpload
 
 
 class SystemEncoder(json.JSONEncoder):
+    def encode(self, obj):
+        # 先序列化为字符串
+        json_str = super().encode(obj)
+        # 移除所有空字符
+        json_str = json_str.replace('\\u0000', '')
+        return json_str
+
     def default(self, obj):
         if isinstance(obj, uuid.UUID):
             return str(obj)


apps/common/encoder/encoder.py
```diff
@@ -15,6 +15,13 @@
...
+    def encode(self, obj):
+        # 先序列化为字符串
+        json_str = super().encode(obj)
+        # 移除所有空字符
+        json_str = json_str.replace('\\u0000', '')
+        return json_str
+
     def default(self, obj):
         if isinstance(obj, uuid.UUID):
             return str(obj)
```
shaohuzhang1, line 27: The provided code has a few minor issues that can be addressed to enhance its functionality:

1. **Missing Type Hinting**: Adding type hints to `SystemEncoder` can improve code readability and help with IDE support.
2. **Removing Redundant Code**: The line `json_str = json_str.replace('\\u0000', '')` is redundant since the default behavior of `replace()` does not remove empty strings (`'\\u0000'`). This can cause unnecessary processing.

```python
from typing import Any

import uuid

class SystemEncoder(json.JSONEncoder):
    """Custom JSON encoder that removes null characters from the output."""
    
    def __init__(self) -> None:
        super().__init__()
    
    def encode(self, obj: Any) -> str:
        # Attempt to serialize the object using parent logic
        result = super().encode(obj)
        
        # Optionally, use replace('\\\\u0000', '') if 'empty string' replacement is needed,
        # but this should typically be handled by the base implementation
        
        return result
    
    def default(self, obj: Any) -> str | UUID:
        """
        Convert non-serializable types into serializable strings/UUIDs.
        """
        if isinstance(obj, uuid.UUID):
            return str(obj)
```

### Explanation of Changes:
- **Type Hints**: Added `Any` for better type hinting, which indicates that the parameters and return values can be of any data type.
- **Initialization Method**: Introduced an explicit constructor method (`__init__`) for proper initialization, although it doesn't add significant functionality beyond what Python expects by default.
- **Removed Redundant Line**: Modified the line to handle cases where no additional replacements might be necessary. Note that handling empty strings (`''`) specifically may depend on specific use cases or configurations.

apps/application/flow/knowledge_workflow_manage.py
```diff
@@ -95,9 +95,7 @@ def hand_node_result(self, current_node, node_result_future):
             self.status = 500
             current_node.get_write_error_context(e)
             self.answer += str(e)
-            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(
-                details=self.get_runtime_details(),
-                state=State.FAILURE)
+            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(state=State.FAILURE)
         finally:
             current_node.node_chunk.end()
             QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(
```
shaohuzhang1, line 101: The provided Python code snippet contains some minor improvements and corrections for better readability and functionality:

1. **Remove Unnecessary Code**: The line `self.answer += str(e)` can be removed since it is not used anywhere else in the function.

2. **Simplify Update Command**: The last update command after the `finally` block can be simplified to include just the `state` field, as you're already updating all fields when setting `details`.

Here's the revised code:

```python
@@ -94,9 +93,6 @@ def hand_node_result(self, current_node, node_result_future):
        try:
            # ... (existing code)

            self.status = 500
            current_node.get_write_error_context(e)
-            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(
-                details=self.get_runtime_details(),
-                state=State.FAILURE)
 
        finally:
            current_node.node_chunk.end()
-            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(details=self.get_runtime_details(), state=State.FAILURE)
+            QuerySet(KnowledgeAction).filter(id=self.params.get('knowledge_action_id')).update(state=State.FAILURE)
```

### Summary of Changes

- Removed the redundant line adding error message to `self.answer`.
- Simplified the final `QuerySet.KnowledgeAction.update()` call by excluding unnecessary parameters.

PR #4502, opened 2025-12-11
chore: limit paragraph list to first 5 segments in get_details method

shaohuzhang1:


apps/application/flow/step_node/document_split_node/impl/base_document_split_node.py
```diff
@@ -161,18 +161,30 @@ def _generate_problem_list(
...
-            'paragraph_list': self.context.get('paragraph_list', []),
+            'paragraph_list': paragraph_list,
             'limit': self.context.get('limit'),
             'chunk_size': self.context.get('chunk_size'),
             'with_filter': self.context.get('with_filter'),
             'patterns': self.context.get('patterns'),
             'split_strategy': self.context.get('split_strategy'),
-            'document_list': self.context.get('document_list', []),
+            # 'document_list': self.context.get('document_list', []),
         }
```
shaohuzhang1, line 190: There is one minor issue with the provided code that will cause an error when rendering documents that do not contain paragraphs:

```python
-           return self.context.get('document_list', []),
+           // 'document_list': self.context.get('document_list', []),
```

The `//` comment should be replaced with a single `' '` or removed entirely since it's causing syntax errors in Python.

Additionally, to avoid recalculating the paragraph limit each time the function is called, consider moving this logic into initialization methods (e.g., constructor) so it only runs once per instance of `_GenerateProblemList`.

PR #4507, opened 2025-12-12

shaohuzhang1:

### DIFF
--- a/apps/common/handle/impl/text/text_split_handle.py
+++ b/apps/common/handle/impl/text/text_split_handle.py
@@ -32,7 +32,7 @@ class TextSplitHandle(BaseSplitHandle):
         if file_name.endswith(".md") or file_name.endswith('.txt') or file_name.endswith('.TXT') or file_name.endswith(
                 '.MD'):
             return True
-        if file_name.index('.') > 0:
+        if '.' in file_name:
             return False
         buffer = get_buffer(file)
         result = detect(buffer)


apps/common/handle/impl/text/text_split_handle.py
```diff
@@ -32,7 +32,7 @@ def support(self, file, get_buffer):
         if file_name.endswith(".md") or file_name.endswith('.txt') or file_name.endswith('.TXT') or file_name.endswith(
                 '.MD'):
             return True
-        if file_name.index('.') > 0:
+        if '.' in file_name:
             return False
         buffer = get_buffer(file)
         result = detect(buffer)
```
shaohuzhang1, line 38: The code snippet you provided is an example of a function named `support` that determines whether a given file's extension indicates it should be supported. It checks if the file has a `.md`, `.txt`, or `.MD` extension and returns `True` if any of these match, otherwise it returns `False`. If there is at least one period in the file name (`.`), it immediately returns `False`.

There are no immediate issues with this logic, but here are some general suggestions for improving readability and maintainability:

1. **Consistent Extension Handling**: Ensure consistent case handling when checking for extensions to avoid confusion. The current implementation treats `.txt` and `.TXT` differently, which might not be intended.

2. **String Methods**: You can use the `in` keyword instead of indexing, which makes the code more readable and concise.

Here's an optimized version of the function using string methods:

```python
def support(file, get_buffer):
    # Define valid text file extensions, ignoring case
    valid_extensions = {'.md', '.txt'}

    # Get the file name from the path
    file_name = file.name

    # Check if the file ends with a valid text extension
    if file_name.lower().split('.')[1] in valid_extensions:
        return True

    # Buffer the content of the file
    buffer = get_buffer(file)
    result = detect(buffer)

    return result
```

### Key Improvements:
- **Case Insensitivity**: Use `file_name.lower()` to ensure the comparison is case-insensitive.
- **List For Valid Extensions**: Store valid extensions in a set for faster lookups.
- **Splitting by Dot**: Extract only the extension part directly after splitting the file name by periods using `file_name.split('.')[1]`.

These changes make the code cleaner and more efficient.

PR #4510, opened 2025-12-12
fix: Local rearrangement model cannot be added

shaohuzhang1:
fix: Local rearrangement model cannot be added

### DIFF
--- a/apps/models_provider/impl/local_model_provider/credential/reranker/web.py
+++ b/apps/models_provider/impl/local_model_provider/credential/reranker/web.py
@@ -34,4 +34,4 @@ class LocalRerankerCredential(BaseForm, BaseModelCredential):
     def encryption_dict(self, model: Dict[str, object]):
         return model
 
-    cache_folder = forms.TextInputField(_('Model catalog'), required=True)
+    cache_dir = forms.TextInputField(_('Model catalog'), required=True)


apps/models_provider/impl/local_model_provider/credential/reranker/web.py
```diff
@@ -34,4 +34,4 @@ def is_valid(self, model_type: str, model_name, model_credential: Dict[str, obje
     def encryption_dict(self, model: Dict[str, object]):
         return model
 
-    cache_folder = forms.TextInputField(_('Model catalog'), required=True)
+    cache_dir = forms.TextInputField(_('Model catalog'), required=True)
```
shaohuzhang1, line 37: The code snippet you provided contains an error: it attempts to define two fields with the same name (`cache_folder` and `cache_dir`) using different class attributes. This will likely result in a runtime exception because Python does not allow duplicate attribute names within the same class.

To fix this, you can either use one of the field names consistently or modify the `cache_dir` definition to include additional information if needed, such as setting its initial value:

```python
class MyForm(forms.Form):
    # ... other fields ...

    cache_folder = forms.TextInputField(_('Model catalog'), required=True)

    def encryption_dict(self, model: Dict[str, object]):
        return model
```

If there was no specific intention to have different fields, simply update the existing `cache_folder` with an appropriate field instance if necessary. If you intended to add new functionality related to directories instead of form caching, ensure that both fields work together appropriately without ambiguity.

PR #4517, opened 2025-12-15
fix: The data source tool cannot obtain the execution parameters

shaohuzhang1:
fix: The data source tool cannot obtain the execution parameters

### DIFF
--- a/apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
+++ b/apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
@@ -197,16 +197,18 @@ class BaseToolLibNodeNode(IToolLibNode):
         else:
             all_params = init_params_default_value | params
         if self.node.properties.get('kind') == 'data-source':
-            exist = function_executor.exec_code(f'{tool_lib.code}\ndef function_exist(function_name): return callable(globals().get(function_name))', {'function_name': 'get_download_file_list'})
+            exist = function_executor.exec_code(
+                f'{tool_lib.code}\ndef function_exist(function_name): return callable(globals().get(function_name))',
+                {'function_name': 'get_download_file_list'})
+            all_params = {**all_params, **self.workflow_params.get('data_source')}
             if exist:
                 download_file_list = []
                 download_list = function_executor.exec_code(tool_lib.code,
-                                                            {**all_params, **self.workflow_params.get('data_source')},
+                                                            all_params,
                                                             function_name='get_download_file_list')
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/tool_lib_node/impl/base_tool_lib_node.py
```diff
@@ -197,16 +197,18 @@ def execute(self, tool_lib_id, input_field_list, **kwargs) -> NodeResult:
...
+                                                            all_params,
                                                             function_name='get_download_file_list')
                 for item in download_list:
                     result = function_executor.exec_code(tool_lib.code,
-                                                         {**all_params, **self.workflow_params.get('data_source'),
-                                                          'download_item': item},
+                                                         {**all_params, 'download_item': item},
                                                          function_name='download')
                     file_bytes = result.get('file_bytes', [])
                     chunks = []
```
shaohuzhang1, line 214: The code looks mostly clean, but here are some minor suggestions for improvements:

1. The `workflow_params` should be imported from the correct module where it is defined instead of being referenced directly.

2. There is no need to pass both `all_params` and `self.workflow_params.get('data_source')` again when calling `get_download_file_list`, as you already have them in `all_params`.

3. Similarly, when calling `download`, you can remove the unnecessary line `{**all_params, **self.workflow_params.get('data_source'), 'download_item': item}` because all parameters are already provided in `all_params`.

4. It's generally better to use dictionary unpacking with `|` (union operator) instead of multiple assignment operations like `{**all_params, ...}` or similar.

Here is the improved version of the function:

```python
def execute(self, tool_lib_id, input_field_list, **kwargs) -> NodeResult:
    ...
    if self.node.properties.get('kind') == 'data-source':
        exist = function_executor.exec_code(
            f'{tool_lib.code}\ndef function_exist(function_name): return callable(globals().get(function_name))',
            {'function_name': 'get_download_file_list'}
        )
        all_params.update({
            'data_source_params': self.workflow_params.get('data_source', {})
        })
        if not exist:
            # Handle case where get_download_file_list does not exist or other errors
            ...

        download_files = []
        files_data = function_executor.exec_code(
            tool_lib.code,
            all_params,
            function_name='get_download_file_list'
        )

        for download_item in files_data:
            file_content = function_executor.exec_code(
                tool_lib.code,
                {
                    **all_params,
                    'download_item': download_item
                },
                function_name='download'
            )

            # Process each chunk of the file content
            chunks = []
            # Your processing logic here...
```

These changes make the code cleaner and more maintainable by reducing redundancy and improving readability.

PR #4521, opened 2025-12-16
fix: The user input parameter was set to non mandatory, but after re entering the canvas editing, it became a mandatory parameter

shaohuzhang1:
fix: The user input parameter was set to non mandatory, but after re entering the canvas editing, it became a mandatory parameter

### DIFF
--- a/ui/src/workflow/nodes/base-node/component/UserInputFieldTable.vue
+++ b/ui/src/workflow/nodes/base-node/component/UserInputFieldTable.vue
@@ -107,7 +107,7 @@
 
 <script setup lang="ts">
 import { onMounted, ref } from 'vue'
-import { set,cloneDeep } from 'lodash'
+import { set, cloneDeep } from 'lodash'
 import Sortable from 'sortablejs'
 import UserFieldFormDialog from './UserFieldFormDialog.vue'
 import { MsgError } from '@/utils/message'
@@ -225,7 +225,7 @@ onMounted(() => {
   inputFieldList.value.forEach((item, index) => {
     item.label = item.label || item.name
     item.field = item.field || item.variable
-    item.required = item.required || item.is_required
+    item.required = item.required == undefined ? item.is_required : item.required
     switch (item.type) {
       case 'input':
         item.input_type = 'TextInput'


ui/src/workflow/nodes/base-node/component/UserInputFieldTable.vue
```diff
@@ -225,7 +225,7 @@ onMounted(() => {
   inputFieldList.value.forEach((item, index) => {
     item.label = item.label || item.name
     item.field = item.field || item.variable
-    item.required = item.required || item.is_required
+    item.required = item.required == undefined ? item.is_required : item.required
     switch (item.type) {
       case 'input':
         item.input_type = 'TextInput'
```
shaohuzhang1, line 231:
```json
{
  "type": "error",
  "message": "The line `item.required = item.required == undefined ? item.is_required : item.required;` can be simplified to use JavaScript's nullish coalescing operator (`??`) and logical OR operator (||) for better readability.",
  "solutions": [
    "Replace the line with:\n```javascript\nitem.required ||= item.is_required;\n```\nor if you prefer a more explicit version based on truthy values of `null`, `undefined`, or empty strings:\n```javascript\n// Option 1: Only consider non-empty strings or falsy but not true for required field.\ncurrentRequired = !!(item.required || '');\nif (!currentRequired && !!item.is_required) {\n  currentRequired = item.is_required;\n}\notherwise...\n// Option 2: Use ternary operator directly.\nitem.required = item.required ?? item.is_required;\nmaybeOption3..\n```\nor simply,\nnormalized option without using conditionals for clarity:\n```javascript\nitem.normalizedRequiredValue = normalizeBoolean(item?.required ?? '');
function normalizeBoolean(rawValue) { // Helper function could also apply custom parsing logic here if needed.\n  return rawValue !== false && rawValue !== '';\n}
```\n**Important Note:** Ensure consistency within your application regarding how boolean flags are handled before applying these solutions."
]
}

# Updated TypeScript Function Definition:

---

Now, please ensure that all files referenced in the above changes have their respective imports included:
- Import lodash for array cloning functionality.
- Import Vue components necessary for dynamic rendering within Vue templates.

Here is how it would look in one comprehensive solution format:

```typescript
import { ref, onMounted } from 'vue';
import { set, cloneDeep} from 'lodash';
import Sortable from 'sortablejs';

onMounted(async () => {
  const inputFieldList = ref<Array<{ label?: string, field?: string, type: string, value?: any, input_type?: string, options?: object, required?: boolean | null }>>([
      ...
      ]);
    }

/* Define a utility/helper function called normalizeBoolean here */
export async function validateInput(value): Promise<boolean> {
    /* Check against some specific criteria to determine validity. For example,
        valid inputs might include true/false, strings like 'enabled', 'yes',
        numbers 1/0, etc. */

    try {

        if (/^true|false$/i.test(String(value))) {// Assuming 'value' is either a string or number

            return String(value).toLowerCase() === 'true'; 
            // Convert the given string into lower case for comparison ('true' & 'False' both will be considered as True)

        }
    
        throw new Error("Invalid boolean value!");
        
    } catch(err){
    console.error('An error occurred while validating:', err);
    }

};
// And then call this helper whenever you need to validate any Boolean values.
```

Issue #4524, opened 2025-12-16
[Bug] Error when using global variables in knowledge base workflow


linfengjjj:
### MaxKB Version

v2.4.1

### Problem Description

知识库工作流全局变量使用报错

### Steps to Reproduce

1.知识库工作流设置变量，使用指定回复输出

<img width="1878" height="834" alt="Image" src="https://github.com/user-attachments/assets/c9d86e99-711b-4cba-a92c-c668eaa5edec" />

2.问题现象

<img width="1917" height="839" alt="Image" src="https://github.com/user-attachments/assets/01f3a3c4-1968-4393-9e2a-2809af3e7093" />

### Related log output

```shell

```

Shenguobin0102:
感谢反馈，后续版本修复

shaohuzhang1:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback, it will be fixed in subsequent versions.

PR #4528, opened 2025-12-17
fix: In the workflow knowledge base, the specified reply node failed to parse the global variable input #4524

shaohuzhang1:
fix: In the workflow knowledge base, the specified reply node failed to parse the global variable input #4524

### DIFF
--- a/ui/src/workflow/nodes/knowledge-base-node/index.vue
+++ b/ui/src/workflow/nodes/knowledge-base-node/index.vue
@@ -51,7 +51,7 @@ const nodeFields = computed(() => {
     const fields = props.nodeModel.properties.user_input_field_list.map((item: any) => ({
       label: typeof item.label == 'string' ? item.label : item.label.label,
       value: item.field,
-      globeLabel: `{{global.${typeof item.label == 'string' ? item.label : item.label.label}}}`,
+      globeLabel: `{{global.${item.field}}}`,
       globeValue: `{{context['global'].${item.field}}}`,
     }))
     set(props.nodeModel.properties.config, 'globalFields', fields)


ui/src/workflow/nodes/knowledge-base-node/index.vue
```diff
@@ -51,7 +51,7 @@ const nodeFields = computed(() => {
     const fields = props.nodeModel.properties.user_input_field_list.map((item: any) => ({
       label: typeof item.label == 'string' ? item.label : item.label.label,
       value: item.field,
-      globeLabel: `{{global.${typeof item.label == 'string' ? item.label : item.label.label}}}`,
+      globeLabel: `{{global.${item.field}}}`,
       globeValue: `{{context['global'].${item.field}}}`,
     }))
     set(props.nodeModel.properties.config, 'globalFields', fields)
```
shaohuzhang1, line 57: The code appears to be setting up fields based on user input, and there are a few points that could improve clarity or functionality:

1. **Type Checking**: Use strict type checking where possible. For example:
   ```javascript
   type Field = {
     label?: string;
     field: string;
     globeLabel?: string;
   };

   const nodeFields = computed(() => {
     return props.nodeModel.properties.user_input_field_list.map((item: any): Field => ({
       label: typeof item.label === 'string' ? item.label : (item.label && item.label.label),
       value: item.field,
       globeLabel: global[item.label]?.label || '',
       globeValue: context.global[item.field] || '',
     }));
   });
   ```

2. **Conditional Handling for Globe Label**:
   If you want to handle cases where `global` might not have the field (`undefined`, etc.), consider using optional chaining.

3. **String Interpolation**: The use of mustache syntax `{}` can sometimes lead to confusion with other interpolation methods if used within template literals. Avoid mixing different formatting styles unless necessary.

4. **Set Function**: Ensure this is being called properly and correctly applies changes.

By implementing these improvements, the code becomes more robust, easier to understand, and less prone to runtime errors related to missing properties.

PR #4531, opened 2025-12-17
refactor: move OpenApiParameter definitions to schema for better organization

shaohuzhang1:
refactor: move OpenApiParameter definitions to schema for better organization  --bug=1064802 --user=刘瑞斌 【接口文档】/admin/api/workspace/{workspace_id}/knowledge/{knowledge_id}/document/split 接口的参数传递方式错了 https://www.tapd.cn/62980211/s/1813972

### DIFF
--- a/apps/knowledge/api/document.py
+++ b/apps/knowledge/api/document.py
@@ -20,28 +20,6 @@ class DocumentSplitAPI(APIMixin):
                 location='path',
                 required=True,
             ),
-
-            OpenApiParameter(
-                name="limit",
-                description="分段长度",
-                type=OpenApiTypes.INT,
-                location='query',
-                required=False,
-            ),
-            OpenApiParameter(
-                name="patterns",
-                description="分段正则列表",
-                type=OpenApiTypes.STR,
-                location='query',
-                required=False,
-            ),
-            OpenApiParameter(
-                name="with_filter",
-                description="是否清除特殊字符",
-                type=OpenApiTypes.BOOL,
-                location='query',
-                required=False,
-            ),
         ]
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


apps/knowledge/api/document.py
```diff
@@ -53,6 +31,18 @@ def get_request():
...
+                    'patterns': {
+                        'type': 'string',
+                        'description': '分段正则列表'
+                    },
+                    'with_filter': {
+                        'type': 'boolean',
+                        'description': '是否清除特殊字符'
                     }
                 }
             }
```
shaohuzhang1, line 48: The given Python function appears to be setting up API parameters using Django Rest Framework (DRF). Here are some points of concern and recommendations:

1. **Django vs FastAPI**:
   - The example uses `djangorestframework` for defining parameters, but you're likely working with FastAPI if you see similar code structures. If this is indeed DRF, please note that DRF has its own way of handling query parameters.

2. **Parameter Naming**:
   - The parameter names (`limit`, `patterns`, `with_filter`) are consistent across both functions, which is good unless there are specific naming conventions in place.

3. **Type Definitions**:
   - All types are correctly defined, such as `INT` for integers and `BOOL` for booleans. Ensure consistency in the usage of these types.

4. **Documentation and Descriptions**:
   - Each parameter has a description provided in Swagger format within the `operation.get_request()` method. This is excellent practice for clarity and maintainability.

5. **Location Parameters**:
   - The location (`location='query'`) is correct according to the requirements. However, ensure that other parameters also follow this pattern consistently if you have other methods like POST requests where different locations might be appropriate.

6. **Binary File Handling**:
   - The addition of `'file': {'type': 'string', 'format': 'binary'}`
     in `get_request()` suggests handling files or binary data. Make sure this line fits into your design and does not conflict with existing logic.

7. **Optimization Suggestions**:
   - There doesn’t appear to be significant opportunities for optimization in the current implementation. Ensure that all fields are necessary and perform efficiently within their contexts.
   - Consider adding more thorough validation on certain inputs, particularly if they interact with sensitive operations.

Overall, the code looks well-structured and should work effectively once integrated into an API setup. Make sure to adjust it accordingly based on whether you’re using Django REST Framework or FastAPI.

PR #4533, opened 2025-12-17
fix: Editor using input method causes editor confusion

shaohuzhang1:
fix: Editor using input method causes editor confusion

### DIFF
--- a/ui/src/components/codemirror-editor/index.vue
+++ b/ui/src/components/codemirror-editor/index.vue
@@ -1,15 +1,16 @@
 <template>
   <div class="codemirror-editor w-full">
-    <Codemirror
-      v-model="data"
-      ref="cmRef"
-      :extensions="extensions"
-      :style="codemirrorStyle"
-      :tab-size="4"
-      :autofocus="true"
-      v-bind="$attrs"
-    />
-
+    <form @submit.prevent>
+      <Codemirror
+        v-model="data"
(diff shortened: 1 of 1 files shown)


ui/src/components/codemirror-editor/index.vue
```diff
@@ -1,34 +1,37 @@
...
+          style="
+            height: calc(100vh - 160px) !important;
+            border: 1px solid #bbbfc4;
+            border-radius: 4px;
+          "
+        />
+      </form>
       <template #footer>
         <div class="dialog-footer mt-24">
           <el-button type="primary" @click="submitDialog"> {{ $t('common.confirm') }}</el-button>
```
shaohuzhang1, line 37: There is no evident problem with this code. The changes you've made do not introduce bugs or unexpected behaviors. However, adding a `<form>` element around the `Codemirror` components might add unintended side effects because forms typically submit data to server-side routes unless explicitly handled. If your intent was only to group elements (like buttons), consider removing the form tags or use them appropriately.

Here’s what you could simplify:

```diff
-<form @submit.prevent>
```

To

```diff
<!-- Alternatively, if simply grouping elements -->
<div> <!-- You can also use divs instead of form elements here... -->   
  [...]
</div>
```

If there's any specific feature added by adding forms, please let me know so I can assist further!

PR #4825, opened 2026-02-27
feat: Document List - Tag Optimization

shaohuzhang1:
feat: Document List - Tag Optimization

### DIFF
--- a/apps/knowledge/serializers/document.py
+++ b/apps/knowledge/serializers/document.py
@@ -392,16 +392,34 @@ class DocumentSerializers(serializers.Serializer):
         status = serializers.CharField(required=False, label=_('status'), allow_null=True, allow_blank=True)
         order_by = serializers.CharField(required=False, label=_('order by'), allow_null=True, allow_blank=True)
         tag = serializers.CharField(required=False, label=_('tag'), allow_null=True, allow_blank=True)
+        tag_ids = serializers.ListField(child=serializers.UUIDField(),allow_null=True,required=False,allow_empty=True)
+        no_tag = serializers.BooleanField(required=False,default=False, allow_null=True)
 
         def get_query_set(self):
             query_set = QuerySet(model=Document)
             query_set = query_set.filter(**{'knowledge_id': self.data.get("knowledge_id")})
+
+            tag_ids = self.data.get('tag_ids')
+            no_tag = self.data.get('no_tag')
             if 'name' in self.data and self.data.get('name') is not None:
                 query_set = query_set.filter(**{'name__icontains': self.data.get('name')})
             if 'hit_handling_method' in self.data and self.data.get('hit_handling_method') not in [None, '']:
                 query_set = query_set.filter(**{'hit_handling_method': self.data.get('hit_handling_method')})
             if 'is_active' in self.data and self.data.get('is_active') is not None:
                 query_set = query_set.filter(**{'is_active': self.data.get('is_active')})
+            if no_tag and tag_ids:
+                matched_doc_ids = QuerySet(DocumentTag).filter(tag_id__in=tag_ids).values_list('document_id', flat=True)
+                tagged_doc_ids = QuerySet(DocumentTag).values_list('document_id', flat=True)
+                query_set = query_set.filter(
+                    Q(id__in=matched_doc_ids) | ~Q(id__in=tagged_doc_ids)
+                )
+            elif no_tag:
+                tagged_doc_ids = QuerySet(DocumentTag).values_list('document_id', flat=True)
+                query_set = query_set.exclude(id__in=tagged_doc_ids)
+            elif tag_ids:
+                matched_doc_ids = QuerySet(DocumentTag).filter(tag_id__in=tag_ids).values_list('document_id', flat=True)
+                query_set = query_set.filter(id__in=matched_doc_ids)
+
             if 'status' in self.data and self.data.get('status') is not None:
                 task_type = self.data.get('task_type')
                 status = self.data.get('status')
--- a/ui/src/views/document/tag/MulAddTagDialog.vue
+++ b/ui/src/views/document/tag/MulAddTagDialog.vue
@@ -168,6 +168,7 @@ const submit = () => {
     emit(
       'addTags',
       tagList.value.map((tag) => tag.value),
+      currentDocId.value,
     )
   })
 }
@@ -182,7 +183,9 @@ function getTags(Key?: string) {
 }
 
 function filterMethod(val: string) {
-  keyOptions.value = allKeyOptions.value.filter((item: any) => item.key.indexOf(val) > -1).slice(0, 100)
+  keyOptions.value = allKeyOptions.value
+    .filter((item: any) => item.key.indexOf(val) > -1)
+    .slice(0, 100)
 }
 
 const createTagDialogRef = ref()
@@ -191,8 +194,10 @@ function openCreateTagDialog(row?: any) {
   createTagDialogRef.value?.open(row)
 }
 
-const open = () => {
+const currentDocId = ref<string | undefined>()
+const open = (rowId?: string) => {
   getTags()
+  currentDocId.value = rowId
   dialogVisible.value = true
   tagList.value = [{}]
 }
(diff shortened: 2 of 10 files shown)


ui/src/views/document/tag/TagDrawer.vue
```diff
@@ -331,6 +344,7 @@ function delTagValue(row: any) {
         .delTag(id, row.id, 'one')
         .then(() => {
           getList()
+          notifyTagChanged()
         })
     })
     .catch(() => {})
```
shaohuzhang1, line 350: The changes made to the ref names for `EditTagDialog` and `CreateTagDialog` and the addition of a new method `notifyTagChanged()` suggest improvements and optimizations.

### Changes Made:

1. **Ref Name Change:**
   ```vue
   <CreateTagDialog ref="createTagDialogRef" @refresh="handleDialogRefresh" />
   <EditTagDialog ref="editTagDialogRef" @refresh="handleDialogRefresh" />
   ```
   These changes help improve code readability by using more descriptive names for the components' refs. Refactoring can make the component's logic clearer when debugging or maintaining.

2. **Added Function `notifyTagChanged`:**
   ```typescript
   function notifyTagChanged() {
     emit('tag-changed');
   }
   ```
   This function emits an event named `'tag-changed'`, which is passed up to the parent component via the `emit` call with the parameter `'tag-changed'`. The receiver could then handle this event accordingly, such as refreshing its state or updating UI based on tag changes.

### Potential Improvements:

1. **Consistent Method Call Signature:**

   Ensure that all similar methods within the template (`@refresh`, `@dialog-cancelled`) have consistent signatures across different dialogs and other components where they may be called. 

2. **Use of Computed Properties for Enhanced Data Flow Management:**

   If specific data needs to be managed consistently without redundancy across these dialogs, consider encapsulating these shared data values within computed properties within `setup` options.

3. **Validation Logic in Dialogs:**

   Consider adding comprehensive validation logic within each dialog’s form controls to ensure inputs meet expected conditions before closing successfully or performing actions like deletion.

These modifications enhance reusability, maintainability, and flexibility while potentially increasing the robustness against bugs due to improper handling of updates across components through events like emission and emitting the updated information back upwards towards the parent component(s).

apps/knowledge/serializers/document.py
```diff
@@ -392,16 +392,34 @@ class Query(serializers.Serializer):
...
+            elif no_tag:
+                tagged_doc_ids = QuerySet(DocumentTag).values_list('document_id', flat=True)
+                query_set = query_set.exclude(id__in=tagged_doc_ids)
+            elif tag_ids:
+                matched_doc_ids = QuerySet(DocumentTag).filter(tag_id__in=tag_ids).values_list('document_id', flat=True)
+                query_set = query_set.filter(id__in=matched_doc_ids)
+
             if 'status' in self.data and self.data.get('status') is not None:
                 task_type = self.data.get('task_type')
                 status = self.data.get('status')
```
shaohuzhang1, line 425: Some of the code can be optimized:

1. **Simplify Boolean Logic**: Instead of multiple `if` statements, consider using conditional logic with OR and AND.

2. **Combine Data Retrieval into One Query**: Avoid making additional database queries inside loops when possible.

3. **Use More Pythonic Constructs**: For example, you could use set operations to optimize filtering more efficiently.

Here's a revised version of the code with some optimizations:

```python
class Query(serializers.Serializer):
    knowledge_id = serializers.CharField(required=False, label=_('knowledge_id'), allow_null=True, allow_blank=True)
    name = serializers.CharField(required=False, label=_('name'), allow_null=True, allow_blank=True)
    hit_handling_method = serializers.CharField(required=False, label=_('hit handling method'), allow_null=True, allow_blank=True)
    is_active = serializers.BooleanField(required=False, label=_('is_active'), allow_null=True)
    order_by = serializers.CharField(required=False, label=_('order by'), allow_null=True, allow_blank=True)
    tag = serializers.CharField(required=False, label=_('tag'), allow_null=True, allow_blank=True)
    tag_ids = serializers.ListField(child=serializers.UUIDField(), required=False, allow_empty=True)
    no_tag = serializers.BooleanField(required=False, default=False, allow_null=True)

    def get_query_set(self):
        query_set = Document.objects.filter(knowledge_id=self.data.get('knowledge_id'))

        if self.data.get('name'):
            query_set = query_set.filter(name__icontains=self.data.get('name'))
        
        if self.data.get('hit_handling_method') not in [None, '']:
            query_set = query_set.filter(hit_handling_method=self.data.get('hit_handling_method'))
        
        if self.data.get('is_active') is not None:
            query_set = query_set.filter(is_active=self.data.get('is_active'))
        
        # Use set intersection or difference based on no_tag and tag_ids
        tagged_doc_ids = list(DocumentTag.objects.values_list('document_id', flat=True))
        no_tag_condition = self.data.get('no_tag')
        tag_ids_condition = self.data.get('tag_ids')

        if tag_ids:
            matched_doc_ids &= set(tag_ids)  
        elif no_tag_condition:
            matched_doc_ids -= set(tagged_doc_ids)
        else:
            matched_doc_ids &= {doc.id for doc in query_set}

        query_set = document_qs.filter(id__in=matched_doc_ids)

        return query_set
```

### Key Changes:
- Combine data retrieval conditions inside one query.
- Use set operations (`&` and `-`) for efficient filtering based on `no_tag`.
- Simplify boolean logic within conditional blocks.

These changes make the logic cleaner and potentially perform better, especially with larger datasets.

ui/src/views/document/tag/MulAddTagDialog.vue
```diff
@@ -191,8 +194,10 @@ function openCreateTagDialog(row?: any) {
...
 }
 
-const open = () => {
+const currentDocId = ref<string | undefined>()
+const open = (rowId?: string) => {
   getTags()
+  currentDocId.value = rowId
   dialogVisible.value = true
   tagList.value = [{}]
 }
```
shaohuzhang1, line 203: The provided code appears to be part of an interactive application where users can add tags to a document or search for existing keys. Here are some observations and suggestions:

### Observations

1. **Duplicate Code**: The `submit` function includes the line `currentDocId.value` which is unnecessary since it's already passed as a parameter to `emit`. You should remove this line.

2. **Unused Function Argument in `openCreateTagDialog`**: The argument `row?: any` does not seem necessary because no use is made of it inside the function. Consider removing it.

3. **Function Declaration in `open` Method**: Although you're declaring `currentDocId` as a reactive value outside the scope of `open`, setting its value within the method doesn't affect its visibility globally because the variable is only assigned within that context. This can be removed if it’s intended to be used elsewhere but is not currently being used.

4. **Use Cases for `keyOptions.value`:** The logic for filtering `keyOptions.value` with `.indexOf(val) > -1` might not work correctly depending on the types of values stored in `allKeyOptions.value`. Ensure that the comparison makes sense for your data structure.

5. **Optimization Suggestions**
   - **Lazy Evaluation**: If filtering takes time, consider delaying the computation of `filterMethod` until there has been user input (`watch`). However, without more context on how often `getTags()` is called, this may vary.
   
   - **Reactivity**: While Reactivity ensures dynamic updates, ensure that dependencies between reactive properties like `tagList.value` and `currentDocId.value` are clearly defined so that Vue re-renders when necessary.

   - **Type Safety**: TypeScript definitions would help ensure type safety across different components.

Here's cleaned-up version of your code with suggested changes applied:

```javascript
// Assuming 'allKeyOptions' and 'dialogVisible' are also reactive
interface TagOption {
  id: string;
  tag: string;
}

const submit = () => {
  emit('addTags', tagList.value.map((tag) => tag.value), currentDocId.value);
});

function getTags() {/* ... */ }

function filterMethod(val: string) {
  keyOptions.value = allKeyOptions.value.filter(item => item.tag.includes(val));
}

const createTagDialogRef = ref();

function openCreateTagDialog() { /* Does nothing useful; can be removed entirely */
}

let currentDocId: string | undefined;

function open(rowId?: string) {
  if (rowId !== undefined && !currentDocId) {
    currentDocId = rowId;
  }
  
  getTags();
  dialogVisible.value = true;
  if (!tagList.value.length) {
    tagList.value.push({});
  }
}
```

This version addresses the identified issues and offers some optimizations while keeping the original functionality intact.

PR #4836, opened 2026-03-03
fix: Icon color

shaohuzhang1:
fix: Icon color


ui/src/components/app-icon/index.ts
```diff
@@ -594,32 +594,32 @@ export const iconMap: any = {
...
+              fill: 'currentColor',
             }),
           ]),
           h('defs', [
             h('clipPath', { id: 'clip0_10754_9765' }, [
-              h('rect', { width: '16', height: '15.9993', fill: 'white' }),
+              h('rect', { width: '16', height: '15.9993', fill: 'currentColor' }),
             ]),
           ]),
         ],
```
shaohuzhang1, line 625: The provided code looks clean and does not contain any immediate irregularities or potential issues. The color attribute values are set to `'#3370FF'` initially but then changed to `'currentColor'`, which is valid if the default color of elements should be used.

However, there's one suggestion:

### Optimization Suggestion:
Ensure that the `<defs>` section at the bottom is not duplicated within other parts of the codebase. In this case, it appears only once where it defines the clip path for all icons in the map. It's always better to define definitions like SVG filters, gradients, etc., just once per file where they're needed to avoid duplication and improve performance. This can especially impact large projects with many SVG components.

apps/knowledge/serializers/document.py
```diff
@@ -395,8 +395,6 @@ class Query(serializers.Serializer):
         tag_ids = serializers.ListField(child=serializers.UUIDField(), allow_null=True, required=False,
                                         allow_empty=True)
         no_tag = serializers.BooleanField(required=False, default=False, allow_null=True)
-        tag_ids = serializers.ListField(child=serializers.UUIDField(),allow_null=True,required=False,allow_empty=True)
-        no_tag = serializers.BooleanField(required=False,default=False, allow_null=True)
         tag_exclude = serializers.BooleanField(required=False,default=False, allow_null=True)
 
         def get_query_set(self):
```
shaohuzhang1, line 400: The code provided in the diff looks like it has duplicate field declarations for `tag_ids` and `no_tag`. This is unnecessary duplication, which should be removed to ensure consistency and possibly improve performance.

Here's the revised version of the code:

```python
@@ -395,10 +395,6 @@ class Query(serializers.Serializer):
         tag_ids = serializers.ListField(
             child=serializers.UUIDField(),
             allow_null=True,
             required=False,
             allow_empty=True
         )
         no_tag = serializers.BooleanField(required=False, default=False, allow_null=True)
-        tag_exclude = serializers.BooleanField(required=False, default=False, allow_null=True)

```

### Key Changes:
1. **Removed Duplicate Field Declarations:**
   The lines `- tag_ids = serializers.ListField(...)` have been moved above the second occurrence of `tag_exclude`.

This improvement not only cleans up the code but also ensures that each field is declared exactly once, maintaining readability and reducing potential confusion or bugs.

PR #4837, opened 2026-03-03
feat: When resource authorization is set to unauthorized, sub resource cascading is set

shaohuzhang1:
feat: When resource authorization is set to unauthorized, sub resource cascading is set

### DIFF
--- a/ui/src/views/system/resource-authorization/component/PermissionTable.vue
+++ b/ui/src/views/system/resource-authorization/component/PermissionTable.vue
@@ -383,7 +383,39 @@ function closeDialog() {
   multipleSelection.value = []
   multipleTableRef.value?.clearSelection()
 }
+function getResourcesByFolderId(treeData: any[], folderId: string): any[] {
+  const result: any[] = []
+  let target: any = null
 
+  function dfs(nodes: any[]) {
+    for (const node of nodes) {
+      if (node.id === folderId) {
+        target = node
+        return
+      }
+      if (node.children?.length) {
+        dfs(node.children)
+        if (target) return
+      }
+    }
+  }
+
+  function collect(node: any) {
+    if (!node?.children) return
+    for (const child of node.children) {
+      result.push(child)
+      collect(child)
+    }
+  }
+
+  dfs(treeData)
+
+  if (target) {
+    collect(target)
+  }
+
+  return result
+}
 function submitPermissions(value: string, row: any) {
   const obj = [
     {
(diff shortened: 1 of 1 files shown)


ui/src/views/system/resource-authorization/component/PermissionTable.vue
```diff
@@ -409,9 +441,15 @@ function submitPermissions(value: string, row: any) {
...
     emitSubmitPermissions(props.data, [row.folder_id], obj)
   }
+  if (['NOT_AUTH'].includes(value) && 'folder' == row.resource_type) {
+    getResourcesByFolderId(props.data, row.id).forEach((n) => {
+      obj.push({ target_id: n.id, permission: 'NOT_AUTH' })
+    })
+  }
   emit('submitPermissions', obj)
 }
 const provider_list = ref<Array<Provider>>([])
```
shaohuzhang1, line 455: The provided code generally looks clean and efficient. However, there are a few areas of improvement:

1. **Return early from `dfs` function**: Inside the `dfs` function, when you find the target node (matching `folderId`), you should immediately return to avoid unnecessary iterations over subsequent nodes.

```javascript
if (node.id === folderId) {
  target = node;
  return; // This ensures that no more iterations happen if we found the node.
}
```

With this change, the function will not continue searching further after finding the matching node, which can improve performance.

2. **Simplifying conditional checks in `submitPermissions`**:
   ```js
   function submitPermissions(value: string, row: any) {
     const obj = [
       ...row.permissions || [], // Merge existing permissions with new ones to handle undefined case.
     ];
 
     if ('VIEW'.includes(value)) {
       obj.unshift({ permission: value });
       emitSubmitPermissions(...props.data);
       return;
     }
   
     const { id } = row;
     
     // Handle role management cases first
     ...
   
     // Not authorized case specifically for folders
     if ('NOT_AUTH' == row.resource_type && ['NOT_AUTH'].includes(value)) {
       getResourcesByFolderId(props.data, id).forEach((n) => {
         obj.push({ target_id: n.id, permission: 'NOT_AUTH' });
       });
     }

     emit('submitPermissions', obj)
   }
   ```

This version simplifies how roles are handled in addition to managing specific resource types (`NOT_AUTH`) for folders, making the function cleaner and potentially reducing boilerplate logic.

3. **Consistent naming conventions**: Functions like `getResourcesByFolderId`, even though written as JavaScript, follow a convention typical in TypeScript or React components. It might be useful to maintain consistency depending on what other languages/technologies you use across your project.

PR #4848, opened 2026-03-09
fix: STT default parameter

shaohuzhang1:
fix: STT default parameter

### DIFF
--- a/apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
+++ b/apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
@@ -23,7 +23,7 @@ class BaseSpeechToTextNode(ISpeechToTextNode):
 
     def execute(self, stt_model_id, audio, model_params_setting=None, **kwargs) -> NodeResult:
         workspace_id = self.workflow_manage.get_body().get('workspace_id')
-        stt_model = get_model_instance_by_model_workspace_id(stt_model_id, workspace_id, **model_params_setting)
+        stt_model = get_model_instance_by_model_workspace_id(stt_model_id, workspace_id, **(model_params_setting or {}))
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/speech_to_text_step_node/impl/base_speech_to_text_node.py
```diff
@@ -23,7 +23,7 @@ def save_context(self, details, workflow_manage):
 
     def execute(self, stt_model_id, audio, model_params_setting=None, **kwargs) -> NodeResult:
         workspace_id = self.workflow_manage.get_body().get('workspace_id')
-        stt_model = get_model_instance_by_model_workspace_id(stt_model_id, workspace_id, **model_params_setting)
+        stt_model = get_model_instance_by_model_workspace_id(stt_model_id, workspace_id, **(model_params_setting or {}))
         audio_list = audio
         self.context['audio_list'] = audio
```
shaohuzhang1, line 29: The code is generally correct but could be optimized slightly. Here's a refined version:

```python
@@ -23,8 +23,8 @@
 
     def execute(self, stt_model_id, audio, model_params_setting=None, **kwargs) -> NodeResult:
-        if not model_params_setting:
-            model_params_setting = {}
         stt_model = get_model_instance_by_model_workspace_id(stt_model_id, workspace_id, **(model_params_setting or {}))
         audio_list = audio
         self.context['audio_list'] = audio
```

Optimization suggestions:
- Added `if` condition to check if `model_params_setting` is `None`.
- Used tuple unpacking with the `(model_params_setting or {})` pattern to handle cases where `model_params_setting` is `None`.
- Improved readability by moving the assignment of `self.context['audio_list']` outside the loop, assuming `audio` can be iterated over without issues.

These changes ensure that the function handles missing parameters more gracefully.

PR #4849, opened 2026-03-09
fix: Folder tree drag error

shaohuzhang1:
fix: Folder tree drag error

### DIFF
--- a/ui/src/components/folder-tree/index.vue
+++ b/ui/src/components/folder-tree/index.vue
@@ -460,7 +460,7 @@ const handleDrop = (draggingNode: any, dropNode: any, dropType: string, ev: Drag
     folderApi
       .putFolder(dragData.id, props.source, obj, loading)
       .then(() => {
-        sortAfterDrop(dragData, dropData, dropType, newParentId)
+        emit('refreshTree')
 
         MsgSuccess(t('common.saveSuccess'))
       })


ui/src/components/folder-tree/index.vue
```diff
@@ -460,7 +460,7 @@ const handleDrop = (draggingNode: any, dropNode: any, dropType: string, ev: Drag
     folderApi
       .putFolder(dragData.id, props.source, obj, loading)
       .then(() => {
-        sortAfterDrop(dragData, dropData, dropType, newParentId)
+        emit('refreshTree')
 
         MsgSuccess(t('common.saveSuccess'))
       })
```
shaohuzhang1, line 466: The code provided appears to be handling a drag-and-drop operation within an application where files or folders are being moved from one location to another. Several points can be considered:

1. **Code Suggestion**: Consider using `await` before calling asynchronous methods like `folderApi`. This can make the code more readable and easier to debug. For example:
   ```typescript
   await folderApi.putFolder(dragData.id, props.source, obj, loading);
   ```

2. **Potential Issues**:
   - If `sortAfterDrop` fails due to some reason but `MsgSuccess` is still called, it might leave the user unsure about the outcome of their action.
   - Not handling errors properly could lead to silent failures if something goes wrong.

   - Ensure that all functions involved (e.g., `handleDrop`, `emit` function) are already defined before they are used inside this block. Otherwise, you will get runtime errors.
   - If there are multiple calls to other functions, consider moving them into separate utility functions or higher-order components for better maintainability and reusability.
   - Consider logging detailed messages at various steps or wrapping the entire logic in a try-catch block to catch potential exceptions gracefully.

If you have specific error handling needs or further modifications required based on your app's context, please let me know!

PR #4863, opened 2026-03-11
fix: The search and positioning status is incorrect. By default, only the node being located needs to be selected, and not all nodes found in the search need to be selected

shaohuzhang1:
fix: The search and positioning status is incorrect. By default, only the node being located needs to be selected, and not all nodes found in the search need to be selected

### DIFF
--- a/ui/src/workflow/common/NodeContainer.vue
+++ b/ui/src/workflow/common/NodeContainer.vue
@@ -443,18 +443,17 @@ const closeNodeMenu = () => {
 const keyWord = ref('')
 const currentKeyWord = ref(false)
 const selectOn = (kw: string) => {
-  props.nodeModel.setSelected(true)
   keyWord.value = kw
+  props.nodeModel.isSelected = false
   currentKeyWord.value = false
-  console.log('selectOn', kw)
 }
 /**
  * 定位时触发
  * @param kw
  */
 const focusOn = (kw: string) => {
+  props.nodeModel.setSelected(true)
   currentKeyWord.value = true
-  console.log('focusOn', kw)
 }
 /**
  * 清除时触发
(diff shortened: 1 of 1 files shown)


ui/src/workflow/common/NodeContainer.vue
```diff
@@ -443,26 +443,24 @@ const closeNodeMenu = () => {
...
 /**
  * 清除时触发
  */
 const clearSelectOn = () => {
   keyWord.value = ''
   currentKeyWord.value = false
-  console.log('onClearSearchSelect')
 }
 
 // 高亮选中关键字
```
shaohuzhang1, line 466: In the provided code snippet, there are several areas that could be improved:

1. **Redundant Code**: The `props.nodeModel.setSelected(true)` line is duplicated in both `selectOn` and `focusOn`. It should only be executed once if necessary.

2. **Variable Consistency**: The naming convention for variables like `keyWord`, `currentKeyWord`, and their properties can be more consistent to help with readability and reduce errors.

3. **Log Statements Simplification**: Instead of logging the keyword every time it changes, consider using `console.debug()` or remove it altogether unless debugging is needed.

4. **Documentation and Comments**: Add comments explaining the purpose of each function and variable to improve maintainability.

5. **Use Cases**: Consider how these functions might be used outside of the context shown here to ensure they cover all use cases efficiently.

Here's an optimized version of the code with some of these improvements:

```javascript
import { ref } from 'vue';

export default {
  setup(props) {
    const keyWord = ref('');
    const currentKeyWord = ref(false);

    const selectOn = (kw: string) => {
      props.nodeModel.selected = true;
      keyWord.value = kw;
      currentKeyWord.value = false;
      console.log('selected', kw);
    };

    const focusOn = (kw: string) => {
      props.nodeModel.selected = true;
      currentKeyWord.value = true;
      console.log('focused on', kw);
    };

    const clearSelectOn = () => {
      keyWord.value = '';
      currentKeyWord.value = false;
      console.log('cleared search selection');
    };

    return {
      keyWord,
      currentKeyWord,
      selectOn,
      focusOn,
      clearSelectOn,
    };
  },
};
```

These optimizations make the code cleaner and easier to maintain.

Issue #4854, opened 2026-03-10
[Bug] API调用时，api key跨智能体越权


ggh-zyy:
### Contact Information

1966236926@qq.com

### MaxKB Version

2.6.0

### Problem Description

进行调用http://<ip:端口>/chat/api/019cac56-1271-75b1-9f80-70326ed22443/chat/completions接口时,使用任意应用的apikey可以成功调用对话

### Steps to Reproduce

<img width="1797" height="1005" alt="Image" src="https://github.com/user-attachments/assets/6d671130-db1c-473d-b3d6-c24e146f68a7" />

<img width="1866" height="930" alt="Image" src="https://github.com/user-attachments/assets/43e8fdf3-13a4-4e7c-b2a3-e254db846452" />
如图我使用图2的apikey调用图一的应用id

<img width="1647" height="966" alt="Image" src="https://github.com/user-attachments/assets/41f111f4-5449-4671-9d7b-0b645c2ebc78" />

<img width="1472" height="2425" alt="Image" src="https://github.com/user-attachments/assets/9ce8251e-ebba-4c36-bef5-9093e53b1630" />
发现调用成功了。

### Related log output

```shell

```

Shenguobin0102:
感谢反馈，后续版本修复

baixin513:
v2.7.0已发布！

PR #4869, opened 2026-03-12
fix: By using API keys from different applications, it is possible to call the same application for dialogue #4854

shaohuzhang1:
fix: By using API keys from different applications, it is possible to call the same application for dialogue #4854

### DIFF
--- a/apps/chat/views/chat.py
+++ b/apps/chat/views/chat.py
@@ -80,10 +80,13 @@ class OpenAIView(APIView):
     )
     def post(self, request: Request, application_id: str):
         ip_address = _get_ip_address(request)
-        return OpenAIChatSerializer(data={'application_id': application_id, 'chat_user_id': request.auth.chat_user_id,
-                                          'chat_user_type': request.auth.chat_user_type,
-                                          'ip_address': ip_address,
-                                          'source': {"type": ChatSourceChoices.API_CALL.value}}).chat(request.data)
+        if application_id != str(request.auth.application_id):
+            raise AppAuthenticationFailed(500, _('Secret key is invalid'))
+        return OpenAIChatSerializer(
+            data={'application_id': application_id, 'chat_user_id': request.auth.chat_user_id,
+                  'chat_user_type': request.auth.chat_user_type,
+                  'ip_address': ip_address,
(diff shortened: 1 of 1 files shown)


apps/chat/views/chat.py
```diff
@@ -80,10 +80,13 @@ class OpenAIView(APIView):
...
+        if application_id != str(request.auth.application_id):
+            raise AppAuthenticationFailed(500, _('Secret key is invalid'))
+        return OpenAIChatSerializer(
+            data={'application_id': application_id, 'chat_user_id': request.auth.chat_user_id,
+                  'chat_user_type': request.auth.chat_user_type,
+                  'ip_address': ip_address,
+                  'source': {"type": ChatSourceChoices.API_CALL.value}}).chat(request.data)
 
 
 class AnonymousAuthentication(APIView):
```
shaohuzhang1, line 92: The provided code contains two significant issues and optimizations:

### Issues:
1. **Security Concerns**: The current implementation checks only if `application_id` received from the user (`request.auth.application_id`) matches the one sent via the API call (`application_id`). However, it does not verify whether these IDs belong to the same application that created the secret key used in the authentication. This could lead to unauthorized access.

   ```python
if application_id != str(request.auth.application_id):
```

2. **Code Duplication**: The `AnonymousAuthentication` class has an empty body, which might indicate a mistake or redundant code.

### Optimization Suggestions:
1. **Add Authentication Verification Check**:
   Ensure both `application_id` from the user's session and the `application_id` used in the secret key are validated to match. If they don't match, throw a specific exception like `AppAuthenticationFailed`.

2. **Remove Redundant Code in `AnonymousAuthentication`**:
   If this class was intended for anonymous users or handling certain types of requests, consider refining its logic accordingly.
   
3. **Review API Endpoint Logic**: Double-check how `OpenAIChatSerializer` processes the data and ensure all required fields are included or properly handled in error scenarios.

4. **Ensure Proper Logging and Feedback to Users**: Handle errors gracefully by returning appropriate HTTP responses with detailed feedback messages, especially during authentication failures.

By addressing these points, you'll make your endpoint more robust and secure.

PR #4870, opened 2026-03-12
fix: Tag link doc title

shaohuzhang1:
fix: Tag link doc title

### DIFF
--- a/ui/src/views/document/tag/TaglinkedDocumentDialog.vue
+++ b/ui/src/views/document/tag/TaglinkedDocumentDialog.vue
@@ -8,7 +8,7 @@
     destroy-on-close
   >
     <template #header>
-      <h4>{{ $t('views.document.tag.key') }}: {{ $t('views.document.tag.value') }}</h4>
+      <h4>{{ currentTag.key }}: {{ currentTag.value }}</h4>
     </template>
     <div>
       <el-tabs v-model="activeTab" @tab-change="handleTabChange">


ui/src/views/document/tag/TaglinkedDocumentDialog.vue
```diff
@@ -8,7 +8,7 @@
     destroy-on-close
   >
     <template #header>
-      <h4>{{ $t('views.document.tag.key') }}: {{ $t('views.document.tag.value') }}</h4>
+      <h4>{{ currentTag.key }}: {{ currentTag.value }}</h4>
     </template>
     <div>
       <el-tabs v-model="activeTab" @tab-change="handleTabChange">
```
shaohuzhang1, line 14: The code snippet you provided has several potential issues:

1. **Undefined `currentTag` Variable**: The template uses `${currentTag.key}` and `${currentTag.value}`, which suggests that there should be a variable named `currentTag` defined somewhere in the Vue component's data. If this variable is not properly initialized, it may result in an error.

2. **Template Syntax Error**: There seems to be a typo in the HTML comment block (`@@ -8,7 +8,7 @@`). It appears like you intended to escape the closing angle bracket (`>`) for the comment instead of removing it completely.

Here are some improvements and suggestions:

### Improvements

- **Define `currentTag`:** Ensure that `currentTag` is correctly defined as a reactive property within the Vue component. For example:
  ```javascript
  export default {
    data() {
      return {
        activeTab: 'key',
        currentTag: { key: '', value: '' }
      };
    },
    methods: {
      handleTabChange(tab) {
        // Handle tab change logic
      }
    }
  };
  ```

### Corrected Template

Remove the unnecessary closing angle brackets from the comment block while ensuring proper syntax:

```html
<template>
  <el-dialog :title="$t(`views.document.tag.${activeTab}-value-title`)" :visible.sync="dialogVisible" destroy-on-close>
    <!-- Use computed properties if needed -->
    <template #header>
      <h4>{{ currentTag.key }}: {{ currentTag.value }}</h4>
    </template>
    <div>
      <el-tabs v-model="activeTab" @tab-change="handleTabChange">
        <!-- Your tabs content here -->
      </el-tabs>
    </div>
  </el-dialog>
</template>

<script>
export default {
  data() {
    return {
      activeTab: 'key',
      dialogVisible: false,
      currentTag: { key: '', value: '' } // Assuming these are set elsewhere, e.g., via props or fetched from API
    };
  },
  methods: {
    handleTabChange(tab) {
      // Handle tab change logic
    }
  }
};
</script>

<style scoped>
/* Add your styles here */
</style>
```

This version assumes that `currentTag` is properly managed throughout the lifecycle of the component. Make sure to adjust the data handling according to your application’s requirements.

PR #4878, opened 2026-03-16
fix: The system prompt for optimizing nodes is ineffective, and the optimization results do not output issues but directly answer the user's questions. #4874

shaohuzhang1:
fix: The system prompt for optimizing nodes is ineffective, and the optimization results do not output issues but directly answer the user's questions. #4874

### DIFF
--- a/apps/application/flow/step_node/question_node/impl/base_question_node.py
+++ b/apps/application/flow/step_node/question_node/impl/base_question_node.py
@@ -125,7 +125,7 @@ class BaseQuestionNode(IQuestionNode):
         return HumanMessage(self.workflow_manage.generate_prompt(prompt))
 
     def generate_message_list(self, system: str, prompt: str, history_message):
-        if system is None or len(system) == 0:
+        if system is not None and len(system) > 0:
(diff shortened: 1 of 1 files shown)


apps/application/flow/step_node/question_node/impl/base_question_node.py
```diff
@@ -125,7 +125,7 @@ def generate_prompt_question(self, prompt):
         return HumanMessage(self.workflow_manage.generate_prompt(prompt))
 
     def generate_message_list(self, system: str, prompt: str, history_message):
-        if system is None or len(system) == 0:
+        if system is not None and len(system) > 0:
             return [SystemMessage(self.workflow_manage.generate_prompt(system)), *history_message,
                     HumanMessage(self.workflow_manage.generate_prompt(prompt))]
         else:
```
shaohuzhang1, line 131: There is an error in the line where `len(system)` is checked. It should be checking for inequality (`!=`), not equality (`==`). Here's the corrected part of the code:

```
if system is not None and len(system) != 0:
```

PR #4881, opened 2026-03-16
fix: After the user asks a question, a blank bar will appear stuck.

shaohuzhang1:
fix: After the user asks a question, a blank bar will appear stuck.

### DIFF
--- a/ui/src/components/ai-chat/component/answer-content/index.vue
+++ b/ui/src/components/ai-chat/component/answer-content/index.vue
@@ -16,7 +16,11 @@
           <MdRenderer
             v-if="
               (chatRecord.write_ed === undefined || chatRecord.write_ed === true) &&
-              answer_text.length == 0
+              answer_text.length == 0 &&
+              answer_text
+                .map((item) => item.content)
+                .join('')
+                .trim().length == 0
             "
             :source="$t('chat.tip.answerMessage')"
           ></MdRenderer>


ui/src/components/ai-chat/component/answer-content/index.vue
```diff
@@ -16,7 +16,11 @@
...
               (chatRecord.write_ed === undefined || chatRecord.write_ed === true) &&
-              answer_text.length == 0
+              answer_text.length == 0 &&
+              answer_text
+                .map((item) => item.content)
+                .join('')
+                .trim().length == 0
             "
             :source="$t('chat.tip.answerMessage')"
           ></MdRenderer>
```
shaohuzhang1, line 26: There seems to be a logical error in the conditional check. The additional condition that checks if `answer_text` has no characters after trimming is redundant because it essentially duplicates the initial check that verifies if `answer_text` is empty:

```javascript
(v-if="(
    (chatRecord.write_ed === undefined || chatRecord.write_ed === true) &&
        answer_text.length == 0
)")
```

Here's a cleaner version of the code without the unnecessary part:

```javascript
<v-if="
    (chatRecord.write_ed === undefined || chatRecord.write_ed === true) &&
        answer_text.length == 0
">
    <MdRenderer
        :source="$t('chat.tip.answerMessage')"
    />
</v-if>
```

This removes the repeated check for an entirely empty string, simplifying the logic while maintaining functionality.

PR #4910, opened 2026-03-19
fix: validate transport type in custom MCP server config to prevent command injection

jackieya:
### Summary

This PR adds a transport type whitelist check in the `else` branch of `BaseMcpNode.execute()` to prevent command injection via `stdio` transport in custom MCP server configurations.

Related: GHSA-38q2-4mm7-qf5h, GHSA-pw52-326g-r5xj

### Problem

The fix in commit c8e8afa8b for GHSA-38q2-4mm7-qf5h only restricts the `referencing` code path. The `else` branch — which loads `mcp_servers` directly from user-supplied JSON — remains unrestricted. An attacker can set `mcp_source` to any non-`referencing` value (or omit it entirely, since the field is optional) to bypass the fix and inject a `stdio` transport configuration with arbitrary `command` and `args`, achieving Remote Code Execution.

### Changes

**File:** `apps/application/flow/step_node/mcp_node/impl/base_mcp_node.py`

Added transport type validation **after** `handle_variables()` to ensure all server configs only use `sse` or `streamable_http` transport. Placing the check after variable substitution prevents bypass via variable references.

```diff
         else:
             servers = json.loads(mcp_servers)
             servers = self.handle_variables(servers)
+            for _name, _cfg in servers.items():
+                if isinstance(_cfg, dict) and _cfg.get('transport') not in ('sse', 'streamable_http'):
+                    raise ValueError(
+                        f"Unsupported transport type: {_cfg.get('transport')}. "
+                        "Only 'sse' and 'streamable_http' are allowed."
+                    )
             params = json.loads(json.dumps(tool_params))
             params = self.handle_variables(params)
```

### Why after `handle_variables()`?

If the check is placed **before** `handle_variables()`, an attacker can use a variable reference (e.g., `["global", "evil_var"]`) as the server config value. This bypasses the `isinstance(cfg, dict)` check, and then `handle_variables()` resolves it into a dict containing `"transport": "stdio"` — effectively bypassing the patch.

### Testing

- With patch applied: PoC returns `Unsupported transport type: stdio` ✅
- Without patch: PoC returns `['root\n']` (RCE confirmed)
- Normal `sse`/`streamable_http` MCP workflows: unaffected ✅
```release-note
fix: validate transport type in custom MCP server config to prevent command injection via stdio
```

### DIFF
--- a/apps/application/flow/step_node/mcp_node/impl/base_mcp_node.py
+++ b/apps/application/flow/step_node/mcp_node/impl/base_mcp_node.py
@@ -9,6 +9,7 @@ from langchain_mcp_adapters.client import MultiServerMCPClient
 from application.flow.i_step_node import NodeResult
 from application.flow.step_node.mcp_node.i_mcp_node import IMcpNode
 from tools.models import Tool
+from common.utils.tool_code import ToolExecutor
 
 
 class BaseMcpNode(IMcpNode):
@@ -34,6 +35,7 @@ class BaseMcpNode(IMcpNode):
         else:
             servers = json.loads(mcp_servers)
             servers = self.handle_variables(servers)  # 处理servers中的变量
+            ToolExecutor().validate_mcp_transport(json.dumps(servers))
             params = json.loads(json.dumps(tool_params))
             params = self.handle_variables(params)
 
--- a/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
+++ b/apps/application/chat_pipeline/step/chat_step/impl/base_chat_step.py
@@ -242,7 +242,8 @@ class BaseChatStep(IChatStep):
         # 兼容老数据
         if not mcp_tool_ids:
             mcp_tool_ids = []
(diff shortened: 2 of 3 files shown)


apps/application/flow/step_node/mcp_node/impl/base_mcp_node.py
```diff
@@ -34,6 +34,12 @@ def execute(self, mcp_servers, mcp_server, mcp_tool, mcp_tool_id, mcp_source, to
         else:
             servers = json.loads(mcp_servers)
             servers = self.handle_variables(servers)  # 处理servers中的变量
+            for _name, _cfg in servers.items():
```
liqiang-fit2cloud: How about this?

Image

liqiang-fit2cloud:
Please use ToolExecutor().validate_mcp_transport() instead like this.
Image

jackieya:
> Please use ToolExecutor().validate_mcp_transport() instead like this. Image

Thanks for the suggestion! I've updated the PR to use `ToolExecutor().validate_mcp_transport()` across all three affected files. The commit has been squashed into a single clean commit.

PR #4919, opened 2026-03-23
fix: Inject application name through XSS

shaohuzhang1:
fix: Inject application name through XSS

### DIFF
--- a/apps/common/middleware/chat_headers_middleware.py
+++ b/apps/common/middleware/chat_headers_middleware.py
@@ -10,6 +10,7 @@ from django.utils.deprecation import MiddlewareMixin
 
 from common.cache_data.application_access_token_cache import get_application_access_token
 from maxkb.const import CONFIG
+from html import escape
 
 
 class ChatHeadersMiddleware(MiddlewareMixin):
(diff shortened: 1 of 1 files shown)


apps/common/middleware/chat_headers_middleware.py
```diff
@@ -24,8 +25,8 @@ def process_response(self, request, response):
             if application_access_token is not None:
                 white_active = application_access_token.get('white_active', False)
                 white_list = application_access_token.get('white_list', [])
-                application_icon = application_access_token.get('application_icon')
-                application_name = application_access_token.get('application_name')
+                application_icon = escape(application_access_token.get('application_icon') or '')
+                application_name = escape(application_access_token.get('application_name') or '')
                 if white_active:
                     # 添加自定义的响应头
                     response[
```
shaohuzhang1, line 32: The provided Python code looks generally correct, but there are a few suggested improvements:

1. **Escape Functionality**: It's good that you're using `escape` to handle any HTML content in the headers. This helps prevent XSS (Cross-Site Scripting) attacks.

2. **Check for Non-existence**: You've added an optional fallback with an empty string (`''`) before applying `escape`. While this prevents errors, it might be better to explicitly check for non-existent keys to avoid raising exceptions. You can use `get()` with a default value instead of checking individually after getting each key.

3. **Consistency and Style**: Ensure consistent formatting throughout the code. PEP 8 guidelines recommend spacing around operators and making sure lines don't exceed 79 characters.

Here's an updated version of the middleware with these suggestions:

```python
from common.cache_data.application_access_token_cache import get_application_access_token
import html

class ChatHeadersMiddleware(MiddlewareMixin):
    
        application_access_token = get_application_access_token()
        
        if application_access_token is not None:
            white_active = application_access_token.get('white_active', False)
            white_list = application_access_token.get('white_list', [])
            
            # Use get() with default values to safely access keys
            application_icon = escape(application_access_token.get('application_icon', ''))
            application_name = escape(application_access_token.get('application_name', ''))
             
            if white_active:
                # Add custom response header
                response['X-CustomHeader'] = f"White Active: {application_icon}, Name: {application_name}"
```

By adding checks for key existence using `get()` and handling them gracefully, you enhance the robustness and maintainability of your code.

PR #4925, opened 2026-03-23
feat: Support Sup referencing tags

shaohuzhang1:
feat: Support Sup referencing tags

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -1,5 +1,8 @@
-<script setup lang="ts"></script>
+<script setup lang="ts">
+import PopoverManager from '@/components/popover-manager/index.vue'
+</script>
 
 <template>
   <RouterView />
+  <PopoverManager></PopoverManager>
 </template>
--- a/ui/src/components/popover-manager/index.vue
+++ b/ui/src/components/popover-manager/index.vue
@@ -1,0 +1,58 @@
+<template>
+  <ElPopover
+    ref="popoverRef"
+    :virtual-ref="currentTarget"
+    virtual-triggering
+    placement="top"
+    :width="240"
+    :show-after="0"
+    :hide-after="80"
+    trigger="hover"
+  >
+    <!-- 富文本内容，data-title 支持 HTML -->
+    <div v-html="currentContent" class="sup-popover__content" />
+  </ElPopover>
+</template>
+
+<script setup lang="ts">
+import { ref, onMounted, onUnmounted } from 'vue'
+import { ElPopover } from 'element-plus'
+
+const popoverRef = ref<any>()
+const currentTarget = ref<any>()
+const currentContent = ref<any>('')
+let hideTimer: any = null
+let lastSup: any = null
+
+function onMouseOver(e: any) {
+  const sup = e.target.closest('sup[data-title]')
+  if (!sup || sup === lastSup) return
+
+  clearTimeout(hideTimer)
+  lastSup = sup
+  currentContent.value = sup.dataset.title
+  currentTarget.value = sup
+  popoverRef.value?.onOpen?.()
+}
+
+function onMouseOut(e: any) {
+  const sup = e.target.closest('sup[data-title]')
+  if (!sup) return
+
+  hideTimer = setTimeout(() => {
+    popoverRef.value?.onClose?.()
+    lastSup = null
+  }, 80)
+}
+
+onMounted(() => {
+  document.addEventListener('mouseover', onMouseOver)
+  document.addEventListener('mouseout', onMouseOut)
+})
+
+onUnmounted(() => {
+  document.removeEventListener('mouseover', onMouseOver)
+  document.removeEventListener('mouseout', onMouseOut)
+  clearTimeout(hideTimer)
+})
+</script>
--- a/ui/src/styles/md-editor.scss
+++ b/ui/src/styles/md-editor.scss
@@ -29,6 +29,26 @@
   ul {
     list-style: circle;
   }
+  sup[data-title] {
+    display: inline-flex;
+    align-items: center;
+    vertical-align: super;
+    font-size: 10px;
+    font-weight: 500;
+    line-height: 1;
+    padding: 2px 4px;
+    border-radius: 4px;
+    background-color: var(--el-color-primary-light-9);
+    color: var(--el-color-primary);
+    border: 1px solid var(--el-color-primary-light-7);
+    cursor: default;
+    white-space: nowrap;
+    letter-spacing: 0.02em;
+    transition:
+      background-color 0.15s,
+      color 0.15s;
+    cursor: pointer;
+  }
 }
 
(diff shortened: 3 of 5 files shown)


ui/src/components/popover-manager/index.vue
```diff
@@ -0,0 +1,58 @@
...
+  document.addEventListener('mouseover', onMouseOver)
+  document.addEventListener('mouseout', onMouseOut)
+})
+
+onUnmounted(() => {
+  document.removeEventListener('mouseover', onMouseOver)
+  document.removeEventListener('mouseout', onMouseOut)
+  clearTimeout(hideTimer)
+})
+</script>
```
shaohuzhang1, line 58: There are several areas where this code can be improved:

1. **Component Lifecycle Management**: The component's lifecycle management is quite basic with only event listeners added at mount time and removed when unmounted. This means that if the component needs to respond dynamically or handle other state changes, these events may not suffice.

2. **Event Delegation**: Using `mouseover` directly on the document level might lead to unnecessary processing and performance overhead, especially if there are many `sup` elements across multiple DOMs. Consider using delegation by adding an event listener to a parent element containing all `sup` elements instead of each individual one.

3. **State Cleanup**: Ensure that all timer instances (`hideTimer`) are properly cleared in `onUnmounted` to prevent memory leaks due to dangling timers.

4. **Type Annotations**: The TypeScript annotations for variables like `lastSup`, `currentContent.value`, and others could be more specific depending on their expected types, though they already include generic type `'any'`. This can help catch errors during compilation.

5. **Accessibility**: The `placeholder` attribute used in the `<div>` inside `<el-popover>` should be translated into accessible text for screen reader users.

6. **Performance Considerations**: While the current implementation has good hover interactions, ensure that the rich text content loading does not become too resource-intensive, particularly if it includes interactive images, videos, etc.

7. **Styling**:
   - Add styles to `.sup-popover__content` to ensure proper formatting (line breaks, padding, margins).
   - If needed, adjust positioning based on `placement` properties provided by Element Plus to better center the content within the popover.

Here’s an updated version addressing some of these concerns:

```html
<template>
  <span ref="popContainer" @mouseover="handleMouseOver($event)" @mouseout="handleMouseOut($event)">
    <ElPopover
      ref="popoverRef"
      :virtual-ref="targetElement"
      virtual-triggering
      placement="top"
      :width="240"
      :show-after="0"
      :hide-after="80"
      trigger="manual"
    >
      <div v-html="contentValue" class="sup-popover__content"></div>
    </ElPopover>
  </span>
</template>

<script setup lang="ts">
import { ref, watchEffect } from 'vue';
import { ElPopover } from 'element-plus';

const popContainer = ref();
const targetElement = ref(null);
const contentValue = ref('');

function updatePopoverData(event: MouseEvent): void {
  const sup = event.target && event.target.closest('sup[data-title]');
  if (!sup) return;

  contentValue.value = sup.dataset.title;
  targetElement.value = sup;
}

watchEffect(updatePopoverData);

function handleMouseOver(event: MouseEvent) {
  clearTimeout(hideTimer);
  lastSup = new EventTarget(sup.target); // Create a dummy EventTarget instance
}

function handleMouseOut(event: MouseEvent) {
  if (!(event.relatedTarget instanceof Node)) return; // Early exit if related target is null

  hideTimer = setTimeout(() => {
    lastSup = null;
    popoverRef.value?.onClose?.();
    targetElement.value = null;
  }, 80);
}

// Use a debounced function here to wait until typing stops before changing the tooltip
</script>

<style scoped>
.sup-popover__content {
  /* Add your styling here */
}
</style>
```

- **Debouncing Logic**: Implement a debounce mechanism for updating the tooltip content. This will allow you to delay updates until user input stops, preventing excessive flickering.
  
By making these adjustments, the component becomes more robust, responsive, and adheres to best practices in Vue.js development and accessibility standards.

ui/src/styles/md-editor.scss
```diff
@@ -29,6 +29,26 @@
...
+    white-space: nowrap;
+    letter-spacing: 0.02em;
+    transition:
+      background-color 0.15s,
+      color 0.15s;
+    cursor: pointer;
+  }
 }
 
 @media only screen and (max-width: 768px) {
```
shaohuzhang1, line 54: The provided CSS code snippet appears to be adding custom styles for `sup` elements with `[data-title]` attributes. While the changes are intended well, there are a few optimizations and improvements that can be made:

1. **Namespace**: Ensure that the variable names (`--el-color-primary`, etc.) follow consistent naming conventions if they are part of an existing style system like Element UI.

2. **Flexbox Usage**: The use of Flexbox might not be necessary here since the alignment is straightforward without additional flex properties. Consider simplifying it or removing unnecessary flex-related styles if they cause any unwanted effects.

3. **Cursor Pointer on Clickable Elements**: For interactive elements like those with tooltips (as indicated by `[role="tooltip"]`), ensure that all child elements inherit the desired pointers. This can be achieved either through direct styling or using SCSS/nested selectors.

Here's a revised version addressing these points:

```css
@@ -29,6 +29,4 @@
 ul {
     list-style: circle;
 }
+sup[data-title]:not([role="tooltip"]){
```

Note: The addition of `.not()` ensures that this rule applies to all `sup` elements except those having `role="tooltip"`. Adjust accordingly based on actual usage scenarios.

ui/src/App.vue
```diff
@@ -1,5 +1,8 @@
-<script setup lang="ts"></script>
+<script setup lang="ts">
+import PopoverManager from '@/components/popover-manager/index.vue'
+</script>
 
 <template>
   <RouterView />
+  <PopoverManager></PopoverManager>
 </template>
```
shaohuzhang1, line 8: The provided code appears to be an update script within an HTML template. Here's a brief review of the changes:

### Changes Made:
1. **Import Statement**: A new import statement has been added at line 5:
   ```typescript
   import PopoverManager from '@/components/popover-manager/index.vue'
   ```

2. **Template Modification**:
   - An empty `<PopoverManager>` element is inserted in place of the comment, which suggests that this component will now manage popovers within the routing view.

### Potential Issues:
- **Component Usage**: Ensure that `PopoverManager` is correctly defined and imported from `@/components/popover-manager/index.vue`. It should expose methods or properties necessary for managing popovers in the application.
  
- **Scoping**: Check if `PopoverManager` needs specific props that need to be passed down from parent components or set in its options configuration.

- **Performance**: If `PopoverManager` involves complex logic or state manipulation, ensure it doesn't significantly impact performance due to excessive computation or side effects outside of lifecycle hooks.

### Suggestions:
- **Type Annotations (Optional)**: Add type annotations on variables or interfaces used within the scripts block for better clarity and maintenance.
- **Documentation**: Consider adding comments in the template or as documentation for maintainers explaining why these changes were made.

In summary, the code introduces a new component into the Vue JS framework using SFC syntax (`<script setup>`) and utilizes it within the router-view. The addition of the new import statement suggests a deeper involvement with additional functionality in your app.

PR #4930, opened 2026-03-24
Fix: Unable to load frontend page after refreshing

shaohuzhang1:
Fix: Unable to load frontend page after refreshing


ui/src/utils/supPopover.ts
```diff
@@ -0,0 +1,165 @@
...
+    document.removeEventListener('mousemove', onMouseMove)
+    cleanupAutoUpdate?.()
+    tooltipEl?.remove()
+    tooltipEl = null
+    arrowEl = null
+    contentEl = null
+    currentSup = null
+    initialized = false
+  },
+}
```
shaohuzhang1, line 165: This TypeScript code for `supPopover.ts` looks mostly correct overall but has a few areas that could be improved for better performance, security, and maintainability:

1. **DOM Manipulation**: The code directly manipulates the DOM using methods like `.createElement()` and `.appendChild()`. This can lead to performance overhead if these operations are done frequently.

2. **Event Listeners**: Adding event listeners on `mousemove` with `{ passive: true }` might not be ideal if you want to prevent default behavior in some cases since it doesn't allow access to the `event.preventDefault()` method.

3. **Security Considerations**: You're sanitizing user input (`sup.dataset.title`) which is good practice, but ensure that this sanitization handles edge cases properly (e.g., potentially dangerous attributes).

4. **Memory Management**: It's important to manage memory effectively, especially with singletons where resources are shared across instances. In this case, ensuring cleanup of event handlers when destroyed helps avoid memory leaks.

5. **Error Handling**: Add error handling around async functions or critical sections to catch any unexpected errors gracefully.

6. **Code Optimization**: For repeated computations (like computing positions multiple times), consider caching results unless they change significantly.

Here’s an optimized version considering some of these points:

```typescript
import {
  computePosition,
  flip,
  offset,
  shift,
  arrow,
  autoUpdate,
  type Placement,
} from '@floating-ui/dom';
import DOMPurify from 'dompurify';

let tooltipEl: HTMLDivElement | null = null;
let arrowEl: HTMLDivElement | null = null;
let contentEl: HTMLDivElement | null = null;
let currentSup: HTMLElement | null = null;
let cleanupAutoUpdate: (() => void) | null = null;
let initialized = false;

// ---- 创建 tooltip DOM ----
function createTooltip(): { el: HTMLDivElement; arrowEl: HTMLDivElement; contentEl: HTMLDivElement } {
  const el = document.createElement('div');
  el.className = 'sup-popover';

  const inner = document.createElement('div');
  inner.className = 'sup-popover__content';
  el.appendChild(inner);

  const av = document.createElement('div');
  av.className = 'sup-popover__arrow';
  el.appendChild(av);

  document.body.appendChild(el);
  return { el, arrowEl: av, contentEl: inner };
}

// ---- 定位计算缓存 ----
type PositionCache = { reference: HTMLElement, result?: ReturnType<typeof computePosition> };

const positionCache: PositionCache[] = [];

// ---- 统一更新位置逻辑 ----
async function updatePosition(reference: HTMLElement): Promise<void> {
  let cachedResult = positionCache.find(cached => cached.reference === reference)?.result;
  
  if (!tooltipEl || !arrowEl || cachedResult == undefined) return;

  const { x, y, placement, middlewareData } = cachedResult ||
    (cachedResult =
      await computePosition(reference, tooltipEl, {
        placement: 'top',
        middleware: [offset(10), flip(), shift({ padding: 8 }), arrow({ element: arrowEl })],
      }));

  Object.assign(tooltipEl.style, { left: `${x}px`, top: `${y}px` });
  tooltipEl.dataset.placement = placement;

  const { x: ax, y: ay } = middlewareData.arrow ?? {};
  const side = placement.split('-')[0] as 'top' | 'bottom' | 'left' | 'right';
  const staticSide = { top: 'bottom', bottom: 'top', left: 'right', right: 'left' }[side];

  Object.assign(arrowEl.style, {
    left: '',
    top: '',
    right: '',
    bottom: '',
    ...(ax != null ? { left: `${ax}px` } : {}),
    ...(ay != null ? { top: `${ay}px` } : {}),
    [staticSide]: '-5px',
  });

  // Cache the updated position
  cachedResult!.x = x;
  cachedResult!.y = y;
}

// ---- 显示 / 隐藏 ----
function show(sup: HTMLElement): void {
  if (!tooltipEl || !contentEl || sup.disabled) return;

  const sanitizedContent = DOMPurify.sanitize((sup.dataset.title ?? '').trim());

  if (sanitizedContent.trim() === '') return hidden();

  tooltipEl.style.display = 'block';
  tooltipEl.style.pointerEvents = 'auto';

  if (!cleanupAutoUpdate) cleanupAutoUpdate = autoUpdate(sup, tooltipEl, () => updatePosition(sup));

  currentSup = sup;
}

function hide(): void {
  if (!tooltipEl) return;
  tooltipEl.style.display = 'none';
  cleanupAutoUpdate && cleanupAutoUpdate();
  cleanupAutoUpdate = null;
  currentSup = null;
}

// ------------------------------
document.addEventListener("touchstart", handleTouchStart, { passive: true });

let lastHoveredElement: HTMLElement | null = null;

async function handleTouchStart(event: TouchEvent): Promise<void> {
  // Handle touch events similarly to mouse enter/leave

  const relatedTarget = event.relatedTarget as HTMLElement | null;

  if (
    (relatedTarget === tooltipEl &&
      currentSup!.contains(relatedTarget)) ||
    !currentSup!
  ) {
    show(lastHoveredElement!);
    lastHoveredElement = currentSup!;
  }

  hide();
}

document.addEventListener('mouseleave', () => {
  hide();
});

// ---------------------------------
// 基础函数：用 pointer 路径判断是否在安全区内
// 记录鼠标坐标
let mouseX = 0;
let mouseY = 0;
window.addEventListener("mousemove", (e) => {
  mouseX = e.clientX;
  mouseY = e.clientY;
}, { passive: true });

function isMouseInsideSafeZone(): boolean {
  if (!tooltipEl || !lastHoveredElement) return false;

  const supRect = lastHoveredElement.getBoundingClientRect();
  const tipRect = tooltipEl.getBoundingClientRect();

  // 大约扩大到 4px，以防边界问题
  const pad = 4;
  const minX = Math.min(supRect.left, tipRect.left) - pad;
  const maxX = Math.max(supRect.right, tipRect.right) + pad;
  const minY = Math.min(supRect.top, tipRect.top) - pad;
  const maxY = Math.max(supRect.bottom, tipRect.bottom) + pad;

  return mouseX >= minX && mouseX <= maxX && mouseY >= minY && mouseY <= maxY;
}

function onMouseOver(target: EventTarget): void {
  const targetElement = target as HTMLElement | null;

  if (!targetElement) return;

  // 取出元素数据项进行标题解析和显示
  const sup = targetElement.closest('sup[data-title]');
 
  if (sup) {
    if (!currentSup) lastHoveredElement = sup;

    show(sup);
  }
}
```
- Added touch support and cleaned up existing logic appropriately.
- Improved error checking before showing/hiding tooltips.
- Slightly adjusted margin calculation.
- Added cache mechanism for computed positioning data.
- Minor adjustments to comments formatting and variable naming. 

Remember to test thoroughly after making changes to ensure compatibility and functionality remains as expected.

ui/src/styles/app.scss
```diff
@@ -565,3 +565,28 @@ h5 {
...
+}
+
+.sup-popover__arrow {
+  position: absolute;
+  width: 8px;
+  height: 8px;
+  background: #fff;
+  transform: rotate(45deg);
+  box-shadow: 1px 1px 4px rgba(0, 0, 0, 0.08);
+}
```
shaohuzhang1, line 592: The code appears to be syntactically correct and well-commented. However, there's one potential issue that could arise if used improperly:

**Issue:** The `.sup-popover` class uses `position: fixed`, which means it remains in its position relative to the viewport even when scrolling. This might not behave as expected on some responsive designs or pages.

**Optimization Suggestions:**

1. **Ensure Responsive Design:** It'd be better to use a more flexible positioning method like `relative` or `absolute` based on the layout requirements of your application, especially if you want the popover to dynamically adjust with screen size changes.

2. **Consider ARIA Roles:** If this component is part of an accessible interface, consider using aria roles (e.g., role="tooltip" for accessibility support) to ensure proper functionality for users with disabilities.

Here's how you can modify it briefly:

```css
.sup-popover {
  /* ... existing styles */
}
```

Remember, these suggestions should complement rather than replace any specific functionality or design decisions you have made regarding the `.sup-popover` element.

ui/src/App.vue
```diff
@@ -1,8 +1,4 @@
-<script setup lang="ts">
-import PopoverManager from '@/components/popover-manager/index.vue'
-</script>
-
+<script setup lang="ts"></script>
 <template>
   <RouterView />
-  <PopoverManager></PopoverManager>
 </template>
```
shaohuzhang1, line 4: The changes made to the provided Vue component seem generally fine for this minor update:

1. **Removed `<script setup>` Directive**: You removed the `lang="ts"` attribute from the `<script setup>` tag in favor of simply closing the tags without specifying a language. This change doesn't impact functionality but might be considered unnecessary redundancy.

2. **Comment Removed**: The original comment line with `-<PopoverManager></PopoverManager>` has been removed entirely. This is a good practice to reduce clutter.

Suggested Optimizations/Changes:
- Consider merging the two script blocks if they are logically related since you already define TypeScript (TS) in both (`lang="ts"`). If not, leaving them separate makes sense for clarity if there's more content in one block than the other.
  
Overall, the code remains clear and works correctly as intended based on the modifications mentioned.

PR #4934, opened 2026-03-24
fix: html_render supports chatUserProfile function

shaohuzhang1:
fix: html_render supports chatUserProfile function

### DIFF
--- a/ui/src/components/ai-chat/index.vue
+++ b/ui/src/components/ai-chat/index.vue
@@ -208,6 +208,14 @@ provide('getSelectModelList', (params: any) => {
     return loadSharedApi({ type: 'model', systemType: 'workspace' }).getSelectModelList(params)
   }
 })
+provide('chatUserProfile', () => {
+  if (props.type === 'ai-chat') {
+    if (chatUser.chat_profile?.authentication_type === 'login') {
+      return chatUser.getChatUserProfile()
+    }
+  }
+  return Promise.resolve(null)
+})
 
 const transcribing = ref<boolean>(false)
 defineOptions({ name: 'AiChat' })
@@ -798,16 +806,6 @@ const handleScroll = () => {
     }
   }
 }
-onBeforeMount(() => {
-  window.chatUserProfile = () => {
-    if (props.type === 'ai-chat') {
-      if (chatUser.chat_profile?.authentication_type === 'login') {
-        return chatUser.getChatUserProfile()
-      }
-    }
-    return Promise.resolve(null)
-  }
-})
 
 function parseTransform(transformStr: string) {
   const result = { scale: 1, translateX: 0, translateY: 0, translateZ: 0 }
--- a/ui/src/components/markdown/HtmlRander.vue
+++ b/ui/src/components/markdown/HtmlRander.vue
@@ -9,8 +9,8 @@
 </template>
 
 <script setup lang="ts">
-import { computed, ref, onMounted, onBeforeUnmount } from 'vue'
-
+import { computed, ref, onMounted, onBeforeUnmount, inject } from 'vue'
+const chatUserProfile = inject('chatUserProfile') as any
 const htmlRef = ref<HTMLIFrameElement>()
 const props = withDefaults(
   defineProps<{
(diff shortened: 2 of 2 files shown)


ui/src/components/markdown/HtmlRander.vue
```diff
@@ -88,6 +121,20 @@ function onMessage(e: MessageEvent) {
...
+          requestId: e.data.requestId,
+          data: ok,
+        },
+        '*',
+      )
+    })
+  }
 }
 
 onMounted(() => {
```
shaohuzhang1, line 140: There is an issue with the use of `inject('chatUserProfile')`. The provided code does not handle cases where there might not be a valid value injected for this key. It should include checks to ensure that `chatUserProfile` exists before making further calls. Also, the `<script>` block at the very beginning seems out of place and should be moved after closing the `<style>` tags or inside the template section.

Additionally, the `_INSTANCE_ID` variable is defined but used without prefixing it with `$`, which suggests it's likely meant to be part of another object. If this is intentional, make sure to include a proper JavaScript module export for such values.

Here are some corrections:

1. Check if `chatUserProfile` is available before calling its methods.
2. Move non-template script blocks closer to their respective elements or wrap them properly.
3. Ensure `_INSTANCE_ID` is correctly referenced within any JavaScript variables.

These changes will help improve robustness and maintainability of the Vue component.

ui/src/components/ai-chat/index.vue
```diff
@@ -798,16 +806,6 @@ const handleScroll = () => {
...
-      if (chatUser.chat_profile?.authentication_type === 'login') {
-        return chatUser.getChatUserProfile()
-      }
-    }
-    return Promise.resolve(null)
-  }
-})
 
 function parseTransform(transformStr: string) {
   const result = { scale: 1, translateX: 0, translateY: 0, translateZ: 0 }
```
shaohuzhang1, line 811: The provided code has several improvements and optimizations that can be applied:

1. **Remove Redundant Declarations**: Remove commented-out lines like `window.chatUserProfile`.

2. **Use `provide` with Proper Function Reference for Reusability**: The `provide` usage is correct, but ensure it provides a function reference rather than its evaluation.

3. **Simplify Conditional Logic**: Inline some conditional logic where possible to reduce nesting and make the code more readable.

4. **Optimize Data Types and Variable Assignments**: Ensure that variable assignments are clear and minimize reassignment without necessary changes.

5. **Consistent Naming Conventions**: Use consistent naming conventions across similar variables or functions.

```typescript
import { ref } from 'vue'
import { onBeforeUnmount } from 'vue'

export default {
  setup(props) {
    const chatUser = props.currentUser || {} // Assuming currentUser exists

    provide('getSelectModelList', (params: any) => {
      return loadSharedApi<{ type: 'model'; systemType: 'workspace' }>({ type: 'model', systemType: 'workspace' }).getSelectModelList(params)
    })

    provide('chatUserProfile', () => {
      if (props.type !== 'ai-chat') return null

      if (
        chatUser.chat_profile &&
        chatUser.chat_profile.authentication_type === 'login'
      ) {
        return chatUser.getChatUserProfile()
      }

      return Promise.resolve(null)
    })

    const transcribing = ref(false)

    defineOptions({ name: 'AiChat' })

    const handleScroll = (event: WheelEvent) => {
      const container = document.getElementById('textArea')
      if (!container) return

      if (Math.abs(event.deltaY) < 50 && !transcribing.value) {
        event.preventDefault()

        const newCursorPos = Math.min(
          container.scrollLeft + ((event.deltaY > 0 ? event.deltaY : -event.deltaY) * 5),
          container.scrollWidth - container.clientWidth,
        )

        container.scrollTo({
          left: Math.round(newCursorPos),
        })
      }
    }

    let lastMousePress: number | null = null

    document.addEventListener('mousedown', () => {
      lastMousePress = Date.now()
    })
    document.addEventListener('mouseup', () => {
      lastMousePress = null
    })
    document.addEventListener('mousemove', (e) => {
      if (
        e.target instanceof HTMLTextAreaElement ||
        e.target.closest('.code-editor-textarea-wrapper') !== null
      )
        return

      if (lastMousePress) {
        const container =	document.getElementById('textInputFieldWrapper')

        if (!container) return

        const delta = e.clientX - container.offsetLeft - container.clientWidth / 2
        const scrollRange = container.offsetWidth - container.clientWidth
        const newPos = Math.max(-scrollRange, Math.min(scrollRange, deltaX))
        
        container.style.scrollLeft += newPos
      }
    }, true)

    onBeforeUnmount(() => {
      document.removeEventListener('wheel', handleScroll)
      document.removeEventListener('mousedown', () => {})
      document.removeEventListener('mouseup', () => {})
      document.removeEventListener('mousemove', (e) => {})
    })

    function parseTransform(transformStr: string) {
      const result = transformStr.match(/^([-+]?\d+(\.\d+)?)([pxcm])?$/i)?.map(Number);
      
      if (result) return {
        scale: result[2] ? parseFloat(result[1]) : 1,
        translateX: result[1],
        translateY: result[1],
        translateZ: result[1],
      };

      return result;
    }

    return {
      transcribing,
      handleScroll,
    };
  },
}
```

### Key Changes:
- Removed redundant assignment to `window.chatUserProfile`.
- Simplified and clarified conditional logic in `chatUserProfile`.
- Improved readability and maintainability throughout the component.

PR #4944, opened 2026-03-25
fix: Document tag search

shaohuzhang1:
fix: Document tag search

### DIFF
--- a/apps/application/flow/step_node/search_document_node/impl/base_search_document_node.py
+++ b/apps/application/flow/step_node/search_document_node/impl/base_search_document_node.py
@@ -59,7 +59,7 @@ class BaseSearchDocumentNode(ISearchDocumentStepNode):
             )
             authorized_knowledge_ids = get_knowledge_list_of_authorized(
                 self.workflow_manage.get_body().get('chat_user_id'),
-                actual_knowledge_ids
+                [str(k_id) for k_id in actual_knowledge_ids]
             )
             document_id_list = QuerySet(Document).filter(
                 id__in=document_id_list,


apps/application/flow/step_node/search_document_node/impl/base_search_document_node.py
```diff
@@ -59,7 +59,7 @@ def execute(self, knowledge_id_list: List, search_mode: str, search_scope_type:
             )
             authorized_knowledge_ids = get_knowledge_list_of_authorized(
                 self.workflow_manage.get_body().get('chat_user_id'),
-                actual_knowledge_ids
+                [str(k_id) for k_id in actual_knowledge_ids]
             )
             document_id_list = QuerySet(Document).filter(
                 id__in=document_id_list,
```
shaohuzhang1, line 65: The provided code looks mostly correct, but there are a few improvements and potential considerations:

1. **String Conversion**: The line `[str(k_id) for k_id in actual_knowledge_ids]` is converting the integer IDs to strings before passing them to `get_knowledge_list_of_authorized()`. This can be simplified using type hinting with a list of integers instead of converting explicitly.

2. **Variable Naming Consistency**: Ensure that all variable names are consistent in their case and style (e.g., camelCase).

3. **Type Hinting**: Consider adding type hints for better readability and maintainability. However, since this snippet is part of an existing function, it may not always make sense to add comprehensive type hints unless you have access to additional context about the expected data types involved.

4. **Empty Check**: Before filtering documents, consider adding checks like:
   ```python
   if not document_id_list:
       # Handle empty document list appropriately
       return []
   ```

5. **Error Handling**: Add error handling to manage cases where `actual_knowledge_ids`, `document_id_list`, or other necessary variables might be None or empty.

Here's an improved version with these points in mind:

```python
import typing as t

def execute(
        self,
        knowledge_id_list: t.List[int],
        search_mode: str,
        search_scope_type: str,
):
    """
    Execute method with improved logic.

    :param knowledge_id_list: List of knowledge IDs.
    :param search_mode: Search mode string.
    :param search_scope_type: Scope type string.
    :return: A response from executing workflow.
    """

    # Get known approved knowledge based on user ID and filtered IDs
    authorized_knowledge_ids = get_knowledge_list_of_authorized(
        self.workflow_manage.get_body().get('chat_user_id'),
        knowledge_id_list,
    )

    # Filter documents that match the knowledge IDs
    document_id_list = Document.objects.filter(id__in=knowledge_id_list)

    # If no valid documents found, return an appropriate message
    if not document_id_list.exists():
        return "No valid documents found."

    # Further processing logic here...
    response = {
        "authorized_knowledge_ids": authorized_knowledge_ids,
        "filtered_documents": document_id_list.values(),
    }

    return response
```

These suggestions should help ensure the code is more robust, clear, and efficient while maintaining compatibility with existing data structures.

PR #4949, opened 2026-03-25
feat: IMG TTV node support reference model

shaohuzhang1:
feat: IMG TTV node support reference model

### DIFF
--- a/ui/src/workflow/nodes/image-generate/index.vue
+++ b/ui/src/workflow/nodes/image-generate/index.vue
@@ -13,10 +13,13 @@
       >
         <el-form-item
           :label="$t('workflow.nodes.imageGenerateNode.model.label')"
-          prop="model_id"
+          :prop="form_data.model_id_type === 'reference' ? 'model_id_reference' : 'model_id'"
           :rules="{
             required: true,
-            message: $t('workflow.nodes.imageGenerateNode.model.requiredMessage'),
+            message:
+              form_data.model_id_type === 'reference'
+                ? $t('workflow.variable.placeholder')
+                : $t('workflow.nodes.imageGenerateNode.model.requiredMessage'),
             trigger: 'change',
           }"
         >
@@ -28,31 +31,51 @@
                   }}<span class="color-danger">*</span></span
                 >
               </div>
+              <el-select
+                v-model="form_data.model_id_type"
+                :teleported="false"
+                size="small"
+                style="width: 85px"
+                @change="form_data.model_id_reference = []"
+              >
+                <el-option :label="$t('workflow.variable.Referencing')" value="reference" />
+                <el-option :label="$t('common.custom')" value="custom" />
+              </el-select>
+            </div>
+          </template>
+          <div class="flex-between w-full" v-if="form_data.model_id_type !== 'reference'">
+            <ModelSelect
+              @change="model_change"
+              @wheel="wheel"
+              :teleported="false"
+              v-model="form_data.model_id"
+              :placeholder="$t('workflow.nodes.imageGenerateNode.model.requiredMessage')"
+              :options="modelOptions"
+              showFooter
+              @focus="getSelectModel"
+              :model-type="'TTI'"
+            ></ModelSelect>
+            <div class="ml-8">
               <el-button
                 :disabled="!form_data.model_id"
-                type="primary"
-                link
                 @click="openAIParamSettingDialog(form_data.model_id)"
                 @refreshForm="refreshParam"
               >
-                <AppIcon iconName="app-setting"></AppIcon>
+                <el-icon>
+                  <Operation />
+                </el-icon>
               </el-button>
             </div>
-          </template>
-
-          <ModelSelect
-            @change="model_change"
-            @wheel="wheel"
-            :teleported="false"
-            v-model="form_data.model_id"
-            :placeholder="$t('workflow.nodes.imageGenerateNode.model.requiredMessage')"
-            :options="modelOptions"
-            showFooter
-            @focus="getSelectModel"
-            :model-type="'TTI'"
-          ></ModelSelect>
+          </div>
+          <NodeCascader
+            v-else
+            ref="nodeCascaderRef"
+            :nodeModel="nodeModel"
+            class="w-full"
+            :placeholder="$t('workflow.variable.placeholder')"
+            v-model="form_data.model_id_reference"
+          />
         </el-form-item>
-
         <el-form-item
           :label="$t('workflow.nodes.imageGenerateNode.prompt.label')"
           prop="prompt"
@@ -158,6 +181,7 @@ import type { FormInstance } from 'element-plus'
 import AIModeParamSettingDialog from '@/views/application/component/AIModeParamSettingDialog.vue'
 import { t } from '@/locales'
 import { useRoute } from 'vue-router'
+import NodeCascader from '@/workflow/common/NodeCascader.vue'
 import { loadSharedApi } from '@/utils/dynamics-api/shared-api'
 import { WorkflowMode } from '@/enums/application'
 const workflowMode = (inject('workflowMode') as WorkflowMode) || WorkflowMode.Application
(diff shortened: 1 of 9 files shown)


ui/src/workflow/nodes/image-generate/index.vue
```diff
@@ -217,6 +247,12 @@ const form = {
...
     if (props.nodeModel.properties.node_data) {
+      if (!props.nodeModel.properties.node_data.model_id_type) {
+        set(props.nodeModel.properties.node_data, 'model_id_type', 'custom')
+      }
+      if (!props.nodeModel.properties.node_data.model_id_reference) {
+        set(props.nodeModel.properties.node_data, 'model_id_reference', [])
+      }
       return props.nodeModel.properties.node_data
     } else {
       set(props.nodeModel.properties, 'node_data', form)
```
shaohuzhang1, line 258: The code is mostly correct but can be optimized for better readability and functionality. Here are some suggestions:

### Optimal Code

1. **Reduce Redundancy**: 
   - Use ternary operators where possible to reduce repetition.

2. **Improve Logic**:
   - Ensure that all logic related to model selection and validation is centralized to maintain consistency.

3. **Handle State Consistency**:
   - Make sure state changes in one component (e.g., cascader or single input) are propagated and reflected correctly.

Here's an improved version of the specified part of your code:

```javascript
<template>
  <!-- ... existing template content ... -->
</template>

<script lang="ts" setup>
// ... imports and other setup ...

const form = reactive({
  model_id: '',
  model_id_type: 'custom', // Default to custom selection
  model_id_reference: [], // Initialize reference array with empty values
  system: '',
  prompt: defaultPrompt,
  negative_prompt: '',
  guidance_scale_1: 7.5,
  aspect_ratio: '',
  height: 512,
  width: 512,
});

const form_data = computed(() => ({
  ...props.nodeModel.properties.node_data || {},
}));

const model_change = (value: string[]) => {
  form.model_id_reference = value;
};

// Validation function remains the same...

const refreshParam = () => {
  // Refresh parameter logic here...
};

// ... rest of the script ...
</script>
```

### Explanation

- **Default Selection**: The `model_id_type` starts as "custom". If not provided in the saved data, it defaults this way.
  
- **Initialization**: The `model_id_reference` array initializes to an empty array when the node first loads.

- **Reactive Form Object**: The `form` object uses `reactive` instead of `computed`, which allows you to directly modify its properties.

- **Consistent Data Model**: The entire saved node data is mapped to the reactive form using spread syntax (`...`). This maintains consistency and reduces redundant checks.

This approach ensures that both single-model and multiple-selection options work consistently across different forms and nodes, optimizing both performance and user experience.

ui/src/workflow/nodes/text-to-video/index.vue
```diff
@@ -216,6 +246,12 @@ const form = {
...
     if (props.nodeModel.properties.node_data) {
+      if (!props.nodeModel.properties.node_data.model_id_type) {
+        set(props.nodeModel.properties.node_data, 'model_id_type', 'custom')
+      }
+      if (!props.nodeModel.properties.node_data.model_id_reference) {
+        set(props.nodeModel.properties.node_data, 'model_id_reference', [])
+      }
       return props.nodeModel.properties.node_data
     } else {
       set(props.nodeModel.properties, 'node_data', form)
```
shaohuzhang1, line 257: The code is mostly well-designed but contains some minor inefficiencies and unnecessary logic.

1. **Prop Validation**: The prop validation in template literals can be improved for clarity and maintainability:

```html
<span>{{ (form_data.model_id_type === 'reference') ? $t("workflow.variable.placeholder") : $t('workflow.nodes.textToVideoGenerate.model.requiredMessage')}}</span>
```

It's better to extract this into a helper function or constant for easier readability.

2. **Custom Model Select Component Logic**: The custom model select component needs to properly handle cases where no models are provided by the API (e.g., an error response). This might require adding additional checks to ensure `model_options` is not null before trying to iterate over it.

3. **Model Reference Handling**: When switching between "Reference" and "Custom", ensure that you correctly manage state related to references if they do have any dependencies on non-reference selections.

4. **Error Logging**: Ensure proper logging of errors across asynchronous operations, such as fetching model options through `loadSharedApi`. Catching these errors separately would help with debugging.

5. **Optimization**: Avoid calling multiple methods inside the same event handler unless necessary. For instance, handling events like `refreshForm` should ideally only update its internal states without performing external validations.

6. **Validation Cleanup**: The cleanup code after `validate()` can also be cleaner. Instead of throwing immediately upon encountering an error during initial form validation, catch them and store them appropriately for later user feedback.

Here's an optimized version of the critical parts mentioned above:

### Template Improvements

```html
<span :class="{ colorDanger: !valid }">{{ displayRequiredMessage }}</span>
```

And define a calculated property for `displayRequiredMessage`:

```javascript
const form_data = computed({
  get: () => {
    // Handle loading state here
    const properties = props.nodeModel.properties ?? {}
    return deepMerge({}, defaults, properties.node_data) || {}
  },
})
```

Then add the required method calculation outside of the render tree:

```javascript
const requiredMessageStyleClass = ''
let valid = false;

if (!this.$refs.aichatNodeFormRef.getRules(this.form).some(r => r.required)) {
  requiredMessageStyleClass = 'color-danger';
}

// If there are existing errors, mark as invalid instead of just styling the label
const isValidRule = this.$refs.aichatNodeFormRef.rules.prompt.some(rule =>
  rule.validator(null, '', {})
);

valid = rules.every(rule => rule.validator(valid));
requiredMessageContent = this.displayRequiredMessage();
requiredMessageStyleClass = `${requiredMessageStyleClass} ${isValidRule ? "" : "-none"`;  
```

### Additional Enhancements

1. **Error Messages**: Provide more informative error messages when users try to submit invalid data. E.g., suggest what specific field needs attention.
   
2. **Dynamic State Update**: Ensure that updating states based on form changes does not cause infinite loops inadvertently due to nested reactive updates.

This approach improves maintainability and enhances the functionality of your form components by leveraging Vue.js' powerful reactive programming capabilities effectively.

ui/src/workflow/nodes/image-understand/index.vue
```diff
@@ -275,6 +300,12 @@ const form = {
...
     if (props.nodeModel.properties.node_data) {
+      if (!props.nodeModel.properties.node_data.model_id_type) {
+        set(props.nodeModel.properties.node_data, 'model_id_type', 'custom')
+      }
+      if (!props.nodeModel.properties.node_data.model_id_reference) {
+        set(props.nodeModel.properties.node_data, 'model_id_reference', [])
+      }
       return props.nodeModel.properties.node_data
     } else {
       set(props.nodeModel.properties, 'node_data', form)
```
shaohuzhang1, line 311: The provided code has several issues that need to be addressed:

1. **Undefined Prop**: The `model_id_type` and `model_id_reference` props being accessed without checks might lead to errors on some nodes.

2. **Duplicate Code**: There is duplicate model selection logic in both the referenced node ID list and custom dropdown. This can be optimized.

3. **Inconsistent Logic**: The logic for handling different values of `form_data.model_id_type` seems inconsistent, causing redundancy.

4. **Potential Security Risks**: Directly accessing object properties using dynamic keys (`props.nodeModel.properties.node_data[props.field]`) without proper validation could expose your application to injection attacks.

Here's an improved version of the code with these issues resolved:

```javascript
const form_data = computed({
  get: () => {
    if (props.nodeModel.properties.node_data) {
      return props.nodeModel.properties.node_data;
    } else {
      return { ...defaultProps };
    }
  },
});

watch(
  () => props.nodeModel.properties.node_data,
  newValue => {
    Object.assign(form_data.value, newValue);
  }
);

function handleModelTypeChange(value) {
  if (value === 'reference') {
    form_data.value.model_id_reference.splice(0);
  }
}

const ModelSelectRef = ref(null);
const NodeCascaderRef = ref(null);

// Use this function to refresh parameter settings when needed.
async function openAIParamSettingDialog(id): Promise<any> {
  // Implementation here...
}

function wheel() {
  // Wheel event handler implementation here...
}
```

Key points:
- Simplified logic by removing redundant code.
- Used Vue's `computed` hook appropriately.
- Added a watcher for node data changes to ensure the component updates.
- Refactored `handleModelTypeChange` function separately from template rendering.
- Dropped unnecessary dependencies like `_defineProp`.
- Removed unused elements and comments.

This should help improve stability and performance while maintaining clarity and integrity.

PR #4956, opened 2026-03-26
fix: same ref name

shaohuzhang1:
fix: same ref name

### DIFF
--- a/ui/src/workflow/nodes/parameter-extraction-node/index.vue
+++ b/ui/src/workflow/nodes/parameter-extraction-node/index.vue
@@ -93,7 +93,7 @@
             </div>
           </template>
           <NodeCascader
-            ref="nodeCascaderRef"
+            ref="inputVariableCascaderRef"
             :nodeModel="nodeModel"
             class="w-full"
             :placeholder="$t('workflow.variable.placeholder')"
@@ -220,10 +220,14 @@ const model_change = (model_id?: string) => {
 
 const VariableSplittingRef = ref()
 const nodeCascaderRef = ref()
+const inputVariableCascaderRef = ref()
 
 const validate = async () => {
   return Promise.all([
     nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    inputVariableCascaderRef.value
+      ? inputVariableCascaderRef.value.validate()
+      : Promise.resolve(''),
     VariableSplittingRef.value.validate(),
   ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })


ui/src/workflow/nodes/parameter-extraction-node/index.vue
```diff
@@ -220,10 +220,14 @@ const model_change = (model_id?: string) => {
...
 
 const validate = async () => {
   return Promise.all([
     nodeCascaderRef.value ? nodeCascaderRef.value.validate() : Promise.resolve(''),
+    inputVariableCascaderRef.value
+      ? inputVariableCascaderRef.value.validate()
+      : Promise.resolve(''),
     VariableSplittingRef.value.validate(),
   ]).catch((err: any) => {
     return Promise.reject({ node: props.nodeModel, errMessage: err })
```
shaohuzhang1, line 233: The code provided is mostly correct but has a few minor improvements that can be made:

1. **Consistent Naming Conventions**: The naming convention for `inputVariableCascaderRef` might not match the actual component if 'inputField' were intended to reference something different.

2. **Validation Logic Duplication**: Both `nodeCascaderRef` and `inputVariableCascaderRef` have identical validation logic in their promises without using an abstracted method like `validateChild`. Consider extracting this logic into its own function and calling it from both places to avoid repetition.

3. **Documentation Comments**: It would help readability to add JSDoc comments describing what each part of the code does, especially with complex components or functions.

Here's the updated and improved version of the code:

```javascript
// Import necessary library imports here

/**
 * Component responsible for variable splitting
 */
const VariableSplittingContent = defineComponent({
  setup(props) {
    // ... previous variables ...

    /**
     * Refers to the cascader component
     */
    const nodeCascaderRef = ref<NodeCascader>()

    /**
     * Refers to the second cascader (if needed)
     */
    const inputVariableCascaderRef = ref<NodeCascader>()

      return Promise.all([
        nodeCascaderRef.value ? await validateChild(nodeCascaderRef.value) : Promise.resolve(''),
+        inputVariableCascaderRef.value
          ? await validateChild(inputVariableCascaderRef.value)
          : Promise.resolve('')
      ])
        .catch((err: any) => {
          throw { node: props.nodeModel, errMessage: err }
        });
    }

    // other methods...

    return (
      <div>
        {/* ... */}
      </div>
    )
  },
})

function validateChild(componentInstance: NodeCascader) {
  if (componentInstance && typeof componentInstance.validate === 'function') {
    return componentInstance.validate();
  } else {
    console.error(`${componentInstance?.constructor.name} needs to implement \`validate\` method.`);
    return '';
  }
}
```

These changes aim to improve code maintainability and readability while fixing a redundant section and ensuring consistent use throughout the component.

PR #4960, opened 2026-03-27
fix: Tool workflow restricts the deletion of basic nodes and start nodes

shaohuzhang1:
fix: Tool workflow restricts the deletion of basic nodes and start nodes

### DIFF
--- a/ui/src/workflow/common/shortcut.ts
+++ b/ui/src/workflow/common/shortcut.ts
@@ -104,6 +104,8 @@ export function initDefaultShortcut(lf: LogicFlow, graph: GraphModel) {
     const nodes = elements.nodes.filter((node) =>
       [
         'start-node',
+        'tool-start-node',
+        'tool-base-node',
         'base-node',
         'loop-body-node',
         'loop-start-node',


ui/src/workflow/common/shortcut.ts
```diff
@@ -104,6 +104,8 @@ export function initDefaultShortcut(lf: LogicFlow, graph: GraphModel) {
     const nodes = elements.nodes.filter((node) =>
       [
         'start-node',
+        'tool-start-node',
+        'tool-base-node',
         'base-node',
         'loop-body-node',
         'loop-start-node',
```
shaohuzhang1, line 111: The code looks generally correct, but there are a few small improvements you could make:

1. **Remove Duplicates**: The array `['start-node', 'tool-base-node']` contains duplicate entries. You should remove one of them to avoid redundancy.

2. **Improve Readability**: Consider grouping similar node types together if they share common functionalities or properties.

Here's the updated code with these changes applied:

```javascript
@@ -104,6 +104,8 @@ export function initDefaultShortcut(lf: LogicFlow, graph: GraphModel) {
       [
         'start-node',
+        'tool-start-node', // Removed duplicate entry
+        'tool-base-node', // Added back only once
         'base-node',
         'loop-body-node',
         'loop-start-node',
```

These changes help maintain clarity in the list and ensure no unnecessary duplicates exist.

PR #4967, opened 2026-03-27
fix: Workflow tool - Modify parameter name to match existing parameter name successfully

shaohuzhang1:
fix: Workflow tool - Modify parameter name to match existing parameter name successfully

### DIFF
--- a/ui/src/workflow/nodes/tool-base-node/component/input/InputFieldTable.vue
+++ b/ui/src/workflow/nodes/tool-base-node/component/input/InputFieldTable.vue
@@ -86,6 +86,14 @@ function deleteField(index: any) {
 const currentIndex = ref<number | null>(null)
 function refreshFieldList(data: any) {
   if (currentIndex.value !== null) {
+    if (
+      inputFieldList.value
+        .filter((item, index) => index != currentIndex.value)
+        .some((field) => field.field == data.field)
+    ) {
+      MsgError(t('workflow.tip.paramErrorMessage') + data.field)
+      return
+    }
     inputFieldList.value?.splice(currentIndex.value, 1, data)
   } else {
     if (inputFieldList.value.some((field) => field.field == data.field)) {


ui/src/workflow/nodes/tool-base-node/component/input/InputFieldTable.vue
```diff
@@ -86,6 +86,14 @@ function deleteField(index: any) {
...
+      inputFieldList.value
+        .filter((item, index) => index != currentIndex.value)
+        .some((field) => field.field == data.field)
+    ) {
+      MsgError(t('workflow.tip.paramErrorMessage') + data.field)
+      return
+    }
     inputFieldList.value?.splice(currentIndex.value, 1, data)
   } else {
     if (inputFieldList.value.some((field) => field.field == data.field)) {
```
shaohuzhang1, line 99: The provided JavaScript code snippet appears to be part of a React component that manages an array of `inputFieldList` items. It contains a function `deleteField` for removing an item by its index and another function `refreshFieldList` for updating the list when fields are changed. Here's some feedback with suggested improvements:

### Suggestions

1. **Null Coalescing Operator**:
   - Use null coalescing operator (`??`) instead of checking `currentIndex.value !== null`.
   
   ```javascript
   if (!currentIndex?.value) {
       // Handle error: currentIndex is not defined or is null/undefined
   }
   ```

2. **Type Checking**:
   - Ensure that `inputFieldList` and `data.field` are objects with consistent types.
   
   ```javascript
   if (!(Array.isArray(inputFieldList) && typeof data.field === 'string')) {
       throw new Error('Invalid types');
   }
   ```

3. **Avoid Multiple Function Calls within Conditionals**:
   - Minimize the number of times you call functions inside conditional statements.

   ```javascript
   if (
       inputFieldList.filter((item, index) => index != (currentIndex ?? 0)).some(
           (field) => field.field === data.field
       )
   ) {
       MsgError(t('workflow.tip.paramErrorMessage') + data.field);
       return;
   }
   ```

4. **Optimize DOM Updates**:
   - If performance is a concern, consider using batch updates for modifying arrays.
   
   ```javascript
   if (
       !inputFieldList.some(({ field }) => field === data.field)
   ) {
       const newIndex = inputFieldList.findIndex(item => item.field === data.field ?? '');
       if (newIndex > -1) {
           inputFieldList.splice(newIndex, 0, data); // Add at front/new location
       } else {
           inputFieldList.push(data); // Append at end
       }
       currentIndex.value = newIndex; // Update current index
   } else {
       msg.error(t('workflow.tip.paramErrorMessage') + data.field);
       return;
   }
   ```

5. **String Interpolation**:
   - Use template literals consistently throughout the codebase.

   ```javascript
   if (!inputFieldList.some(item => item.field === `${data.field}`)) {
       currentIndex.value = inputFieldList.find(item => item.field === data.field)?.index || -1;
       if (currentIndex >= 0) {
           inputFieldList splice(currentIndex, 1, data);
       } else {
           inputFieldList.push(data);
       }
         }
     }
```

6. **Ensure Input Validation**:
   - Validate input arguments to avoid unexpected behavior.

   ```javascript
   if (!inputFieldList?.length || typeof data.field !== 'string') {
       console.error('Invalid parameters');
       return;
   }
   ```

By applying these suggestions, the code becomes more robust and efficient while maintaining readability.

PR #5216, opened 2026-05-06
fix: Judge regular matching using fullmatch

shaohuzhang1:
fix: Judge regular matching using fullmatch


wangliang181230:
通配符比较器，也一样要改。

PR #5246, opened 2026-05-12
【backup】perf: Change the parameter `stream` and `re_chat` to non required and set the default value to `False`

wangliang181230:
perf: Change the parameter `re_chat` to non required and set the default value to `False`

---

优化：将API参数 `stream` 和 `re_chat` 修改为非必填，并设置默认值为 `False`


liuruibin:
不合并

PR #6135, opened 2026-05-28
fix: volcanic engine embedding endpoint (ep-xxx-xx) not working with multimodal API

hanliang97:
## Problem\n\nVolcanic engine embedding models with \`ep-xxx-xx\` endpoint IDs fail with "list index out of range" error during credential verification. Only \`doubao-xxx\` model names worked correctly.\n\n## Root Cause\n\nTwo issues:\n\n1. The code only routed \`doubao-embedding-vision-\` prefixed models to the multimodal embeddings API (\`/embeddings/multimodal\`), while \`ep-xxx-xx\` endpoint models fell through to the standard embeddings API (\`/embeddings\`) which doesn't support multimodal endpoints.\n\n2. The multimodal embeddings API returns \`resp.data\` as a single \`MultimodalEmbedding\` object with an \`.embedding\` attribute, not a list. The extraction logic was incorrectly treating it as a list.\n\n## Fix\n\n- Unify all volcanic engine embedding calls to use \`multimodal_embeddings.create()\` API (compatible with both pure text and multimodal inputs)\n- Fix \`_extract_embedding\` to check \`hasattr(data, 'embedding')\` first, correctly handling single MultimodalEmbedding objects\n- Both \`ep-xxx-xx\` endpoint IDs and \`doubao-xxx\` model names now work correctly\n\n## Testing\n\nTested with endpoint \`ep-20260527164729-5mshx\` (doubao-embedding-vision based) - credential verification and embedding generation both pass.

### DIFF
--- a/apps/models_provider/impl/volcanic_engine_model_provider/model/embedding.py
+++ b/apps/models_provider/impl/volcanic_engine_model_provider/model/embedding.py
@@ -39,38 +39,25 @@ class VolcanicEngineEmbeddingModel(MaxKBBaseModel):
     def embed_documents(
             self, texts: List[str]
     ) -> List[List[float]]:
-        if self.model_name.startswith("doubao-embedding-vision-"):
-            embeddings = []
-            for text in texts:
-                multimodal_input = {"type": "text", "text": text}
-                resp = self.client.multimodal_embeddings.create(
-                    model=self.model_name,
-                    input=[multimodal_input],
-                    encoding_format="float",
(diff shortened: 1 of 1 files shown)


hanliang97:
为了解决 火山向量模型 sp-xxx-xxx无法添加问题

wxg0103:
我们需要兼容这种历史的模型和新的模型  除非没有人使用  才会完全删除

PR #6389, opened 2026-07-09
Refresh MiniMax M3 provider metadata

octo-patch:
Reason: refresh MiniMax endpoints and M3 context metadata in existing provider

## Summary
- Updates MiniMax LLM default endpoint to the global OpenAI-compatible URL.
- Records both global and China OpenAI/Anthropic endpoint metadata, including /anthropic/v1 URLs.
- Refreshes MiniMax-M3 and MiniMax-M2.7 descriptions with context, pricing, and thinking metadata, and makes MiniMax-M3 the default LLM.

## Checks
- python3 -m py_compile apps/models_provider/impl/minimax_model_provider/credential/llm.py apps/models_provider/impl/minimax_model_provider/model/llm.py apps/models_provider/impl/minimax_model_provider/minimax_model_provider.py


liuruibin:
image

PR #6530, opened 2026-07-28
Register MiniMax-M3 as a vision model

octo-patch:
Reason: MiniMax-M3 was registered only as an LLM, so image and video input workflows could not select it as a vision model.

## Summary
- Add a MiniMax vision (IMAGE) model entry for MiniMax-M3 so existing image/video input workflows can select it as a vision model.
- Add `MiniMaxImageModel` (`model/image.py`) backed by the OpenAI-compatible chat endpoint, mirroring the existing LLM chat model (keeps `reasoning_split` and the MiniMax API base).
- Add `MiniMaxImageModelCredential` (`credential/image.py`) with `api_base`/`api_key` fields and validation that streams a text message, mirroring other providers' image credentials.
- Register MiniMax-M3 under `ModelTypeConst.IMAGE` (with a default vision entry) in the MiniMax provider, following the same pattern used by other providers.

## Checks
- `python3 -m py_compile apps/models_provider/impl/minimax_model_provider/minimax_model_provider.py apps/models_provider/impl/minimax_model_provider/model/image.py apps/models_provider/impl/minimax_model_provider/credential/image.py`

### DIFF
--- a/apps/models_provider/impl/minimax_model_provider/credential/image.py
+++ b/apps/models_provider/impl/minimax_model_provider/credential/image.py
@@ -1,0 +1,74 @@
+# coding=utf-8
+from typing import Dict
+
+from django.utils.translation import gettext_lazy as _, gettext
+from langchain_core.messages import HumanMessage
+
+from common import forms
+from common.exception.app_exception import AppApiException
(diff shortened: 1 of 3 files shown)


liuruibin:
no plan

PR #6621, opened 2026-08-07
feat: add Atlas Cloud model provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud as an OpenAI-compatible LLM provider
- expose nine current Atlas Cloud model IDs with configurable API URL and API key
- register the provider in both the backend provider registry and frontend selector
- add focused coverage for provider metadata, defaults, model construction, and API-key encryption

## Testing

- Python 3.11 compileall for the changed Python modules
- Ruff 0.15.12 on all changed Python files
- 5 focused unit tests passed
- TypeScript 5.8.3 parsed the changed provider registry successfully
- git diff --check
- verified all 9 configured model IDs against the live Atlas Cloud model catalog


liuruibin:
no plan

PR #6683, opened 2026-08-14
fix: assemble Anthropic streamed tool_use blocks before replay

### LINKED ISSUE #6480: [Bug] Anthropic models: tool calls fail with "Input tag 'input_json_delta' ... does not match any of the expected tags" (streamed tool_use never assembled)
### MaxKB Version

v2.10.4-lts (also reproduced on v2.10.2-lts)

### Problem Description

When an application has a **custom tool attached** and uses an **Anthropic model** (`model_anthropic_provider`), every message that triggers a tool call fails. The Anthropic API rejects MaxKB's follow-up request:

```
BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': "messages.1.content.1: Input tag 'input_json_delta' found using 'type' does not match any of the expected tags"}}
```

and, in the other variant:

```
BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'messages: text content blocks must be non-empty'}}
```

MaxKB returns this to the caller as HTTP 200 with the error text as the **answer body**, so it reaches end users as the assistant's reply.

### Cause

`input_json_delta` is an Anthropic **streaming event type**, not a message content type. Anthropic streams tool arguments as a series of `content_block_delta` events of type `input_json_delta`, each carrying a `partial_json` fragment. These must be **accumulated and assembled into a completed `tool_use` content block** before the assistant turn is replayed in `messages`.

MaxKB appears to replay the raw streamed deltas instead. A search of the v2.10.4-lts tree finds no handling for either field:

```
grep -rn "input_json_delta\|partial_json" apps/     # zero hits
```

`_yield_mcp_response()` in `apps/application/flow/tools.py` handles only OpenAI-shaped tool deltas (`tool_call_chunks` / `tool_calls` / `invalid_tool_calls`). There is no branch for Anthropic content blocks, which is consistent with both error messages: the unassembled delta is sent verbatim, and the accompanying text block is empty.

### Steps to Reproduce

1. Register an Anthropic model (reproduced with `claude-sonnet-5`, `model_anthropic_provider`).
2. Create a **SIMPLE** agent using that model.
3. Create a custom Python tool and attach it (`tool_enable: true`), e.g.

```python
def live_price(query):
    return 'DJI Neo: 4242.42 GEL, in stock.'
```

4. Set the system prompt so the tool is actually used, e.g. "For any question about price or stock you MUST call the live_price tool."
5. Call the OpenAI-compatible endpoint with `stream: true` (required — with tools attached, `stream: false` is unsupported per #6300 / #4846 / #4363):

```bash
curl -N "$MAXKB/chat/api/$APP_ID/chat/completions" \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"gpt-3.5-turbo","stream":true,"messages":[{"role":"user","content":"How much is the DJI Neo?"}]}'
```

**Expected:** the answer quotes the tool's value (`4242.42 GEL`).
**Actual:** the answer is the `BadRequestError` text above.

### The same setup works on an OpenAI-shaped provider

Identical agent, identical tool, identical prompt, only the model swapped to `model_openai_provider` (`gpt-4o-mini`): **no error**, and the answer contains `4242.42 GEL`, confirming the tool executed. This isolates the defect to the Anthropic provider's streamed tool-call handling rather than to tools, sandboxing, or the endpoint.

### Impact

Custom tools and MCP cannot be used at all with Anthropic models. Because tools require `stream: true`, and streaming is exactly where the assembly fails, there is no working combination for Anthropic users.

Tested on v2.10.2-lts and v2.10.4-lts (CE), Docker, single-node.

LHMQ878:
## Summary
- Anthropic streams tool arguments as `input_json_delta` events. MaxKB was replaying those raw deltas on the next turn, which Anthropic rejects with `Input tag 'input_json_delta' does not match any of the expected tags` (and empty text blocks with `text content blocks must be non-empty`).
- Assemble completed `tool_use.input` from streamed fragments, drop `input_json_delta` / empty text blocks, and treat `stop_reason=tool_use` as a finish chunk.
- Adds regression tests that do not require an Anthropic account.

Fixes https://github.com/1Panel-dev/MaxKB/issues/6480

## Test plan
- [x] `python -m unittest apps.application.flow.tests.test_anthropic_tool_content -v`
- [ ] SIMPLE agent + custom tool + Anthropic model, `stream: true`: tool call completes and the follow-up turn is not a 400

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/application/flow/anthropic_tool_content.py
+++ b/apps/application/flow/anthropic_tool_content.py
@@ -1,0 +1,135 @@
+# coding=utf-8
+"""Assemble Anthropic streamed tool_use blocks before they are replayed.
+
+Anthropic streams tool arguments as `input_json_delta` events. Those deltas are
+valid on the wire, but they are not legal `messages[].content` block types.
+Replaying them on the next turn yields:
+
+    Input tag 'input_json_delta' found using 'type' does not match any of the expected tags
+
+Empty `text` blocks from the same stream also fail with:
+
+    messages: text content blocks must be non-empty
+"""
+
+from __future__ import annotations
+
+import json
+
+ANTHROPIC_TOOL_STOP_REASONS = ("tool_use", "end_turn")
+_STREAM_DELTA_TYPES = {"input_json_delta"}
+
+
+def is_anthropic_tool_finish(response_metadata, chunk_position=None) -> bool:
+    """True when this chunk closes an Anthropic (or OpenAI-mapped) tool turn."""
+    meta = response_metadata or {}
+    if meta.get("finish_reason") == "tool_calls":
+        return True
+    if meta.get("stop_reason") in ANTHROPIC_TOOL_STOP_REASONS:
+        return True
+    return chunk_position == "last"
+
+
+def collect_input_json_deltas(content) -> dict:
+    """Concatenate `input_json_delta` fragments keyed by content-block index."""
+    collected = {}
+    if not isinstance(content, list):
+        return collected
+    for block in content:
+        if not isinstance(block, dict) or block.get("type") != "input_json_delta":
+            continue
+        index = block.get("index")
+        if index is None:
+            continue
+        piece = block.get("partial_json")
+        if piece is None:
+            piece = block.get("input") or ""
+        if not isinstance(piece, str):
+            piece = json.dumps(piece, ensure_ascii=False) if piece else ""
+        collected[index] = collected.get(index, "") + piece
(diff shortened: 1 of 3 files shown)


liuruibin:
no plan

PR #6743, opened 2026-08-22
fix: filter NUL characters from chat records

XIIRUAN:
#### What this PR does / why we need it?

PostgreSQL rejects native NUL (0x00) characters in text fields. Workflow output can reach `ChatRecord.problem_text` or `ChatRecord.answer_text` and cause the record write to fail.

Related to #6722.

#### Summary of your change

- Remove NUL characters from chat record question and answer text before persistence.
- Keep the fix scoped to `ChatRecord` instead of changing every model.
- Add a regression test that exercises the real `ChatRecord.save()` sanitization path.

#### Validation

- `ChatRecordTestCase.test_save_removes_nul_characters`
- `ruff check apps/application/models/application_chat.py apps/application/tests.py`
- Python compilation and diff checks

- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [x] Considered the docs impact; no documentation change is required for this fix.

### DIFF
--- a/apps/application/models/application_chat.py
+++ b/apps/application/models/application_chat.py
@@ -109,6 +109,11 @@ class ChatRecord(AppModelMixin):
     source = models.JSONField(verbose_name="来源", default=dict)
     ip_address = models.CharField(max_length=128, verbose_name="ip地址", default='')
 
+    def save(self, *args, **kwargs):
+        self.problem_text = self.problem_text.replace("\x00", "")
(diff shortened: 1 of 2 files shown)


shaohuzhang1:
Recommended to use
```
str.replace('\\u0000', '')
```
To filter NUL characters, please do not submit unit test code.

PR #6746, opened 2026-08-23
fix(i18n): correct broken keys and mistranslations in en-US/zh-Hant locales

simpleqt:
## Summary
All verified at HEAD:

- `ui/src/locales/lang/{en-US,zh-Hant}/workflow.ts`: the `showCondition`/`hideCondition` keys contained an invisible **U+200C zero-width non-joiner**, so the lookups in `Constructor.vue` (`workflow.nodes.baseNode.visibilitySetting.showCondition`/`hideCondition`) fell through and the Visibility Setting dialog rendered the raw key instead of "Show Condition"/"Hide Condition". Removed the U+200C from both locales (zh-CN is clean).
- `en-US/common.ts`: task status `STARTED` was translated as **'Padding'**; zh-CN/zh-Hant say 执行中/執行中 → 'Running'. Shown in the execution-record drawers.
- `en-US/workflow.ts`:
  - `initiator: 'Iniiator'` → 'Initiator' (column header in execution records)
  - `restoreVersion: 'Restore Previous Version"'` — stray trailing quote rendered on three workflow pages
  - `'Wireless loop must have a Break node'` → 'Infinite loop...' (无限循环; the same file already uses 'Infinite Loop' for `infiniteLoop`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
   <a href="https://github.com/1Panel-dev/maxkb/releases/latest"><img src="https://img.shields.io/github/v/release/1Panel-dev/maxkb" alt="Latest release"></a>
   <a href="https://github.com/1Panel-dev/maxkb"><img src="https://img.shields.io/github/stars/1Panel-dev/maxkb?color=%231890FF&style=flat-square" alt="Stars"></a>    
   <a href="https://hub.docker.com/r/1panel/maxkb"><img src="https://img.shields.io/docker/pulls/1panel/maxkb?label=downloads" alt="Download"></a><br/>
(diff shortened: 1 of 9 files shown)


shaohuzhang1:
Please do not merge and submit with other businesses to ensure that each PR corresponds to only one business. I will close this GitHub PR

simpleqt:
Understood — the branch had picked up an unrelated docs commit, which is exactly the mixing you're describing. I've resubmitted the locale-only changes (rebased on current main, single commit, no other files) as #6793. Going forward my MaxKB PRs will stay strictly one topic per branch.

PR #6750, opened 2026-08-23
fix(docs): repair root-relative README and tool guide links

simpleqt:
- README.md language switcher used href="/README_CN.md" and
  href="/README.md"; on GitHub absolute paths resolve against the
  domain, not the repo, so both 404. Drop the leading slash so they
  resolve to the sibling files in the repository root.
- ui/public/tool/*/detail.md referenced screenshots via
  /admin/tool/img/*.jpg. The images actually live in ui/public/tool/img/
  and the /admin/ prefix is host-absolute: it only happens to work in a
  default MaxKB deployment (ADMIN_PATH=/admin) and breaks on GitHub or
  any custom ADMIN_PATH. Replace with relative ../img/ paths that point
  at the real files in the repo.


shaohuzhang1:
The use of absolute paths is for the program to use, and using relative paths will cause the program image to 404

simpleqt:
You're right — thank you for catching it. I misread the root-relative image paths as broken links; they're the runtime URL convention the app resolves, and my relative-path rewrite would 404 the images. The README link change falls under the same misreading (both forms work on GitHub). Closing agrees with me — no part of this should land. Sorry for the noise.

PR #6953, opened 2026-09-07
feat(provider): add aimlapi.com model provider

hugoaimlapi:
Hi! I'm Hugo from aimlapi.com — an AI aggregator that gives access to 1000+ models in one API, trusted by 400k+ users.

We'd love to be available as a provider option inside MaxKB — so we went ahead and did all the technical work on our side.

To build our partnership, we offer a 50/50 revenue share on all traffic from this integration. (P.S.: tracking starts as soon as this release goes live, so no earnings will be lost during setup)

My contacts: hugo@aimlapi.com (email / Slack), Telegram: @hug0the

#### What this PR does / why we need it?

MaxKB users can already point the generic OpenAI provider's API URL at AI/ML API, but then it shows up unbranded, with no curated multi-type catalog and no attribution. This PR adds aimlapi.com as a named model provider, wired through the existing `BaseChatOpenAI` / `BaseTextToImage` machinery — the same pattern as the Regolo aggregator provider.

#### Summary of your change

- New provider `apps/models_provider/impl/aimlapi_model_provider/` (OpenAI-compatible, default API URL `https://api.aimlapi.com/v1`, registered as `model_aimlapi_provider`), covering 4 model types: LLM (18 curated ids, default `openai/gpt-4o-mini`), vision (7), embedding (6), text-to-image (5). Any other id from https://api.aimlapi.com/v1/models can be typed in when adding a model.
- `const.py`: attribution headers (`HTTP-Referer`, `X-Title`, `X-AIMLAPI-Partner-ID`, `X-AIMLAPI-Source`) are attached only when the request actually goes to `api.aimlapi.com` — if the user points the API URL at another endpoint or a proxy, nothing is added; user-supplied headers always win over ours.
- Unset optional params are omitted instead of being sent as `null` — several models on AI/ML API return 400 on `null` `temperature`/`top_p`/`seed`/`tools`.
- `model/embedding.py`: sends plain strings via the `openai` SDK — AI/ML API's `/v1/embeddings` rejects the token-id arrays that langchain's `OpenAIEmbeddings` would send.
- `model/tti.py`: `auto`/unset image params are omitted (different image models accept different value sets); `check_auth()` validates the key by requesting a non-existent model — invalid key → 401, valid key → 404 — so credential checks never generate (or bill) an image.
- Tests: `AIMLAPIAttributionTest` in `apps/models_provider/tests.py` (7 tests: header scoping, caller-headers-win, no mutation of the shared constant, null-param filtering, TTI auto-param handling). A malformed partner id is accepted with a 200 and silently unattributed, so the id format is pinned by a test.
- i18n: `en_US` / `zh_CN` / `zh_Hant` `django.po` entries for the two new tooltip strings.
- UI: provider entry + icon in `ui/src/components/dynamics-form/items/model/provider-data.ts`.

#### How to Test

```
cd apps
python manage.py test models_provider.tests.AIMLAPIAttributionTest
```

Manual: Model Settings → Add Model → aimlapi.com, paste an API key from aimlapi.com, then validate an LLM / embedding / vision / text-to-image model.

- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed. (No user-facing docs change required; happy to follow up with a docs PR if you'd like the provider listed.)


liqiang-fit2cloud:
No plan to support this, thanks for your contribution anyway.

hugoaimlapi:
okay, thanks for your reply!

PR #7054, opened 2026-09-15
Register MiniMax H3 text-to-video configuration

octo-patch:
Reason: MiniMax H3 text-to-video generation is missing from the registry and its required generation parameters are unavailable in the settings form.

Register MiniMax H3 as the default text-to-video entry with a dedicated credential form using the v2 API endpoint. Expose required 2K resolution and an integer duration control spanning 4–15 seconds, defaulting to 4 seconds, for the existing v2 payload builder.

Validation:
- Python compilation of both changed files passed.
- Ruff 0.15.12 checks (`--select E9,F63,F7,F82`) passed on both changed files.
- `git diff --check` and the diff secret scan passed.

Full application tests and live API generation were not run.


wxg0103:
Thanks — I’ve applied the suggested changes.

